PR #425, opened 2022-09-23
Make sending descriptors to all substreams optional

james-smith-za:
This was breaking the dsim which initially tried to send descriptors to all its substreams, instead of just the first one. This fix allows the dsim to actually start, which will hopefully give us the ability to run a correlator again.

Closes NGC-809.

### DIFF
--- a/src/katgpucbf/send.py
+++ b/src/katgpucbf/send.py
@@ -57,7 +57,7 @@ class DescriptorSender:
         interval: float,
         first_interval: Optional[float] = None,
         *,
-        all_substreams: Optional[bool] = False,
+        all_substreams: bool = False,
     ) -> None:
         self._stream = stream
         self._heap_reference_list = spead2.send.HeapReferenceList(
--- a/src/katgpucbf/send.py
+++ b/src/katgpucbf/send.py
@@ -45,6 +45,9 @@ class DescriptorSender:
     first_interval
         Delay (in seconds) immediately after starting. If not specified, it
         defaults to `interval`.
+    send_all_descriptors
+        Send descriptors to all substreams if true, otherwise default behaviour
+        is to send only to the first substream.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


james-smith-za:
I'm going to wait until Jenkins finishes the build, so that I have a Docker image to test out in an actual correlator.

src/katgpucbf/send.py
```diff
@@ -45,6 +45,9 @@ class DescriptorSender:
     first_interval
         Delay (in seconds) immediately after starting. If not specified, it
         defaults to `interval`.
+    send_all_descriptors
```
bmerry, line 48: I'm not totally happy with the clarity of this name. How about `all_substreams`?

src/katgpucbf/send.py
```diff
@@ -53,10 +56,13 @@ def __init__(
         descriptors: spead2.send.Heap,
         interval: float,
         first_interval: Optional[float] = None,
+        send_all_descriptors: Optional[bool] = False,
```
bmerry, line 59:
```suggestion
        send_all_descriptors: bool = False,
```
Otherwise you're saying that it can be `None` as well, and I don't see a need for that.

bmerry: I'd suggest making this a keyword-only argument, since you're already calling it as a keyword argument (which I like for boolean flags since it makes the purpose of the argument clear, compared to just seeing an argument that is `True`).

bmerry (changes requested):
Looks good - just bikeshedding.

src/katgpucbf/send.py
james-smith-za: This is a fair point, I concur.

src/katgpucbf/send.py
```diff
@@ -53,10 +56,14 @@ def __init__(
         descriptors: spead2.send.Heap,
         interval: float,
         first_interval: Optional[float] = None,
+        *,
+        all_substreams: Optional[bool] = False,
```
bmerry, line 60:
```suggestion
        all_substreams: bool = False,
```

bmerry (changes requested):

bmerry (approved):

PR #426, opened 2022-09-26
Make dsim generate saturation counts

bmerry:
Emit saturation count and saturation flag in the `digitiser_status` SPEAD item.

A major part of this is some not-really-related refactoring: previously the initial signal was set up by `main`, bypassing the mechanisms used by `request_signal`. That was done to avoid a cyclic dependency, where the DeviceServer needs a Sender which required the start time which could only sanely be computed after the initial signal was sampled (because sampling is slow). I instead broke that dependency by making Sender only need the start time when run instead of when constructed.

Relates to NGC-796.

### DIFF
--- a/src/katgpucbf/dsim/__init__.py
+++ b/src/katgpucbf/dsim/__init__.py
@@ -1,7 +1,7 @@
 # noqa: D104
 
 ################################################################################
-# Copyright (c) 2021, National Research Foundation (SARAO)
+# Copyright (c) 2021-2022, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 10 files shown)


src/katgpucbf/dsim/__init__.py
```diff
@@ -1,7 +1,7 @@
 # noqa: D104
 
 ################################################################################
-# Copyright (c) 2021-2022, National Research Foundation (SARAO)
+# Copyright (c) 2021, National Research Foundation (SARAO)
```
james-smith-za, line 4: This change seems ... counterintuitive... is it just reverted to an older situation by removing lines 22-25?

bmerry: Yes. If you look at the diff against `main`, there are no changes to this file. I could squash this commit with [43dbf4f](https://github.com/ska-sa/katgpucbf/pull/426/commits/43dbf4f63f8b7b877a3bc48d6e1e30661004c0bc) so that there is no spurious history for this file.

james-smith-za: Nah, no need for that I think.

bmerry:
Github says the pr-merge build failed, but it looks fine in Jenkins (seems like it's been flummoxed by the qualification autotesting work), so I'm merging.

PR #430, opened 2022-10-12
Digitiser clip count update

bmerry:
- Rename the sensor to be more consistent with eq and delay sensors
- Add an input_samples Prometheus metric

Closes NGC-796 (when combined with some PRs on other repos).


amishatishpatel (approved):
Excellent. Slightly related, is there a reason we do counter(s) as `cnt` and not `ctr`?

bmerry:
> Slightly related, is there a reason we do counter(s) as cnt and not ctr?

For dig-clip-cnt I just based on the name on what MK uses. You'd probably need to hunt down people who no longer work here to find out the history.

PR #440, opened 2022-10-18
Force pre-commit to run with Python 3.8

bmerry:
This is particularly needed for the pip-compile hook, since the output is different depending on the Python version.

Checklist is N/A.

Relates to NGC-560.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3
+  python: python3.8
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks


bmerry:
Have requested two reviewers just to check that this doesn't break anyone's workflow (e.g. if they don't have a `python3.8` binary) rather than because it needs such careful reading. Please test with `pre-commit run --all-files`.

amishatishpatel (approved):
Assuming you meant test **this branch** when you requested
> Please test with `pre-commit run --all-files`

I've done so in two separate environments (locally and on qgpu02). Both have the `python3.8` binary, both did take some time at those `pip-compile` steps but ultimately passed. Let me know if there's anything else you'd like us to test.

james-smith-za:
[I'm not using Windows anymore.](https://www.youtube.com/watch?v=seKLsHOkN1U)

james-smith-za:
This does break things for me actually:
```
An unexpected error has occurred: CalledProcessError: command: ('/home/jsmith/.pyenv/versions/3.10.6/envs/katgpucbf-3.10/bin/python3.10', '-mvirtualenv', '/home/jsmith/.cache/pre-commit/repo6pyb4m0b/py_env-python3.8', '-p', 'python3.8')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.8'
    
stderr: (none)
```
I did it with `pyenv` to make the local interpreter 3.10.6.

I'm struggling to understand the purpose for the change?

bmerry:
> I'm struggling to understand the purpose for the change?

Try running `pre-commit run --all-files` on `main` when your interpreter is Python 3.10. What I found is that it tells you the requirements.txt files were changed, because some of the dependencies aren't needed for Python 3.10 so it removes them. And if you then committed those changes, you'd break things for people still using Python 3.8.

I also had some weird issues on Jenkins where mypy would complain about some construct in Dask, saying it was 3.9-only. I couldn't reproduce it locally, but forcing the pre-commit environment to be built with a Python 3.8 interpreter seemed to fix it.

james-smith-za:
If I run it on 3.10.6 in main at the moment, I get your dask problem:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

/home/jsmith/.cache/pre-commit/repo_ab0dm0k/py_env-python3/lib/python3.10/site-packages/dask/utils.py:334: error: Parenthesized context managers are only supported in Python 3.9 and greater
Found 1 error in 1 file (errors prevented further checking)
```

bmerry:
> I'm happy with this as an intermediate step, but obviously once the port to 3.10 is complete we'd need to find a fix to the funny mypy/dask issue.

If we ditch 3.8 support and update the mypy config to assume 3.10 that should sort itself out.

PR #446, opened 2022-10-25
Add some LaTeX magic to make ° work in listings

bmerry:
This broke when a test failed and a pytest backtrace was pasted into the report, and one of the functions had a ° in the docstring.


james-smith-za (approved):
I'm curious as to what the `1` on the end is for.

bmerry:
> I'm curious as to what the `1` on the end is for.

It's the "length" in the replacement - I guess for alignment with fixed-width fonts.

PR #447, opened 2022-10-26
Allow 2 degree phase error for delay-rate/phase-rate

bmerry:
These tests are measuring the difference in phase between two things which may each have up to 1° error. The requirement is going to be reviewed, so this is a pragmatic solution to get the test passing on 32768 channels. It will probably need to be revisited once narrowband is implemented.

[32K report](https://github.com/ska-sa/katgpucbf/files/9870699/report.pdf)

Closes NGC-719.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -260,6 +260,7 @@ def check_phases(
     actual: np.ndarray,
     expected: np.ndarray,
     caption: str,
+    tolerance_deg: float = 1,
 ) -> None:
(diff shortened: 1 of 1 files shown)


james-smith-za (commented):
Everything looks OK, but I'm somewhat puzzled by this plot:
image
The stated SFDR doesn't seem to match the plot. I see nothing at -80 dB, unless it's adjacent channels that are squashed too close to the main tone to see?

bmerry:
> Everything looks OK, but I'm somewhat puzzled by this plot:

The same thing seems to be happening for `main` (with 8K channels). Maybe you can file a ticket? I don't recall exactly what @andrewvanderbyl implemented for this test.

PR #458, opened 2022-11-07
Introduce device status sensors for the F- and X-engines

james-smith-za:
This is fairly straightforward. Most of the complexity lives in `aiokatcp` so I'm hoping to keep this PR simple.

Main possible point of contention that I can see is the `update_aggregate()` mechanism - which could potentially be more efficient. I also couldn't really think of a good way to unit-test things.

Closes NGC-442.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -452,7 +452,6 @@ class Engine(aiokatcp.DeviceServer):
     ) -> None:
         super().__init__(katcp_host, katcp_port)
         self._populate_sensors(self.sensors)
-        self.sensors.add(DeviceStatusSensor(self.sensors))
 
         # Attributes copied or initialised from arguments
         self._srcs = list(srcs)
@@ -719,6 +718,7 @@ class Engine(aiokatcp.DeviceServer):
                 initial_status=aiokatcp.Sensor.Status.NOMINAL,
             )
         )
+        sensors.add(DeviceStatusSensor(sensors))
 
     async def _next_in(self) -> InItem | None:
         """Load next InItem for processing.
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -49,6 +49,7 @@ from ..monitor import Monitor
 from ..queue_item import QueueItem
 from ..ringbuffer import ChunkRingbuffer
 from ..send import DescriptorSender
+from ..utils import DeviceStatusSensor
 from . import SAMPLE_BITS, recv, send
 from .compute import Compute, ComputeTemplate
 from .delay import AbstractDelayModel, LinearDelayModel, MultiDelayModel, wrap_angle
@@ -451,6 +452,7 @@ class Engine(aiokatcp.DeviceServer):
     ) -> None:
         super().__init__(katcp_host, katcp_port)
         self._populate_sensors(self.sensors)
+        self.sensors.add(DeviceStatusSensor(self.sensors))
 
         # Attributes copied or initialised from arguments
         self._srcs = list(srcs)
--- a/src/katgpucbf/utils.py
+++ b/src/katgpucbf/utils.py
@@ -20,14 +20,19 @@ import ipaddress
 import logging
 import signal
 from asyncio import get_event_loop
+from enum import Enum
+from typing import Optional, TypeVar
 
-from aiokatcp import DeviceServer
+import aiokatcp
 from katsdptelstate.endpoint import endpoint_list_parser
+
+_T = TypeVar("_T")
+_U = TypeVar("_U")
 
 logger = logging.getLogger(__name__)
 
 
-def add_signal_handlers(server: DeviceServer) -> None:
+def add_signal_handlers(server: aiokatcp.DeviceServer) -> None:
     """Arrange for clean shutdown on SIGINT (Ctrl-C) or SIGTERM."""
     signums = [signal.SIGINT, signal.SIGTERM]
 
@@ -53,3 +58,50 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
         return [(ep.host, ep.port) for ep in endpoints]
     except ValueError:
         return value
+
+
+class DeviceStatus(Enum):
+    """Discrete `device-status` readings.
+
+    Doesn't follow the convention of :class:`aiokatcp.Sensor.Status` because
+    it's useful for `UNKNOWN` to have the highest (or worst) value.
+    """
+
+    OK = 1
+    DEGRADED = 2
+    FAIL = 3
+    UNKNOWN = 4
+
+
+class DeviceStatusSensor(aiokatcp.AggregateSensor):
+    """Summary sensor for quickly ascertaining device status.
+
+    This takes its value from the worst status of its target set of sensors, so
+    it's quick to identify if there's something wrong, or if everything is good.
+    """
+
+    def __init__(self, target: aiokatcp.SensorSet):
+        super().__init__(target=target, sensor_type=DeviceStatus, name="device-status")
+
+    def update_aggregate(
+        self,
+        updated_sensor: Optional[aiokatcp.Sensor[_T]],
+        reading: Optional[aiokatcp.Reading[_T]],
+        old_reading: Optional[aiokatcp.Reading[_T]],
+    ) -> aiokatcp.Reading[DeviceStatus]:  # noqa: D102
+        # For device status it's far simpler just to re-calculate everything
+        # each time. In my mind I thought there might be a more elegant /
+        # efficient way of doing this, but it's the simplest thing that works
+        # for now.
+        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.UNKNOWN
+        for sensor in self.target.values():
+            if self.filter_aggregate(sensor):
+                worst_status = max(worst_status, sensor.status)
+
+        if worst_status == aiokatcp.Sensor.Status.NOMINAL:
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.NOMINAL, DeviceStatus.OK)
+        if worst_status == aiokatcp.Sensor.Status.WARN:
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.WARN, DeviceStatus.DEGRADED)
+        if worst_status == aiokatcp.Sensor.Status.UNKNOWN:  # Do we need to take this situation into account?
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.UNKNOWN, DeviceStatus.UNKNOWN)
+        return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.ERROR, DeviceStatus.FAIL)
--- a/src/katgpucbf/utils.py
+++ b/src/katgpucbf/utils.py
@@ -21,13 +21,12 @@ import logging
 import signal
 from asyncio import get_event_loop
 from enum import Enum
-from typing import Optional, TypeVar
+from typing import TypeVar
 
 import aiokatcp
 from katsdptelstate.endpoint import endpoint_list_parser
 
 _T = TypeVar("_T")
-_U = TypeVar("_U")
 
 logger = logging.getLogger(__name__)
 
@@ -70,7 +69,6 @@ class DeviceStatus(Enum):
     OK = 1
     DEGRADED = 2
     FAIL = 3
-    UNKNOWN = 4
 
 
 class DeviceStatusSensor(aiokatcp.AggregateSensor[DeviceStatus]):
@@ -81,19 +79,21 @@ class DeviceStatusSensor(aiokatcp.AggregateSensor[DeviceStatus]):
     """
 
     def __init__(self, target: aiokatcp.SensorSet) -> None:
-        super().__init__(target=target, sensor_type=DeviceStatus, name="device-status")
+        super().__init__(
+            target=target, sensor_type=DeviceStatus, name="device-status", description="Overall engine health"
+        )
 
     def update_aggregate(
         self,
-        updated_sensor: Optional[aiokatcp.Sensor[_T]],
-        reading: Optional[aiokatcp.Reading[_T]],
-        old_reading: Optional[aiokatcp.Reading[_T]],
+        updated_sensor: aiokatcp.Sensor[_T] | None,
+        reading: aiokatcp.Reading[_T] | None,
+        old_reading: aiokatcp.Reading[_T] | None,
     ) -> aiokatcp.Reading[DeviceStatus] | None:  # noqa: D102
         # For device status it's far simpler just to re-calculate everything
         # each time, than to try and maintain state.
         if reading is not None and old_reading is not None and reading.status == old_reading.status:
             return None  # Sensor didn't change state, so no change in overall device status
-        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.UNKNOWN
+        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.NOMINAL
         for sensor in self.target.values():
             if self.filter_aggregate(sensor):
                 worst_status = max(worst_status, sensor.status)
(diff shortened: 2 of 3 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -451,6 +452,7 @@ def __init__(
     ) -> None:
         super().__init__(katcp_host, katcp_port)
         self._populate_sensors(self.sensors)
+        self.sensors.add(DeviceStatusSensor(self.sensors))
```
amishatishpatel, line 455: Any particular reason you've `add`ed this here instead of within `_populate_sensors`? (Same question for the XBEngine)

src/katgpucbf/utils.py
```diff
@@ -20,14 +20,19 @@
...
 from asyncio import get_event_loop
+from enum import Enum
+from typing import Optional, TypeVar
 
-from aiokatcp import DeviceServer
+import aiokatcp
 from katsdptelstate.endpoint import endpoint_list_parser
 
+_T = TypeVar("_T")
+_U = TypeVar("_U")
```
amishatishpatel, line 30: I see why you've done `_T`, but is `_U` necessary for this PR? Or just future-proofing/insisting on a convention?

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+        for sensor in self.target.values():
+            if self.filter_aggregate(sensor):
+                worst_status = max(worst_status, sensor.status)
+
+        if worst_status == aiokatcp.Sensor.Status.NOMINAL:
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.NOMINAL, DeviceStatus.OK)
+        if worst_status == aiokatcp.Sensor.Status.WARN:
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.WARN, DeviceStatus.DEGRADED)
+        if worst_status == aiokatcp.Sensor.Status.UNKNOWN:  # Do we need to take this situation into account?
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.UNKNOWN, DeviceStatus.UNKNOWN)
```
amishatishpatel, line 104: I would've thought yes, but the question still stands. Maybe Bruce has a better idea.

amishatishpatel: That being said, I see why you've done the mapping this way, but it does seem odd that this doesn't actively account for an `aiokatcp.Sensor.Status.FAILURE`.
Is the 'behaviour' of a `device-status` sensor in an ICD somewhere?

src/katgpucbf/utils.py
```diff
@@ -20,14 +20,19 @@
 import logging
 import signal
 from asyncio import get_event_loop
+from enum import Enum
```
amishatishpatel, line 23: I remember you lobbied for `IntEnum` in `aiokatcp`, any reason you're using the 'normal' one here?

amishatishpatel (commented):
Mainly questions seeking clarification.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+    ) -> aiokatcp.Reading[DeviceStatus]:  # noqa: D102
+        # For device status it's far simpler just to re-calculate everything
+        # each time, than to try and maintain state.
+        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.UNKNOWN
+        for sensor in self.target.values():
+            if self.filter_aggregate(sensor):
+                worst_status = max(worst_status, sensor.status)
+
+        if worst_status == aiokatcp.Sensor.Status.NOMINAL:
+            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.NOMINAL, DeviceStatus.OK)
```
bmerry, line 100: `sensor.timestamp` will be whichever sensor was last in `self.target` (it might even be self), which is probably not your intention. I didn't give as much thought to the timestamp as a I should have in the AggregateSensor design. What do you think of the following:
- If both `old_reading` and `reading` are not None (i.e. a sensor update), use `reading.timestamp`.
- Otherwise, use `time.time()`.

I'd also suggest always taking a max with the current `self.timestamp`, to avoid time warping if some sensors take time from the data stream and others from the host's clock.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+        super().__init__(target=target, sensor_type=DeviceStatus, name="device-status")
+
+    def update_aggregate(
+        self,
+        updated_sensor: Optional[aiokatcp.Sensor[_T]],
+        reading: Optional[aiokatcp.Reading[_T]],
+        old_reading: Optional[aiokatcp.Reading[_T]],
+    ) -> aiokatcp.Reading[DeviceStatus]:  # noqa: D102
+        # For device status it's far simpler just to re-calculate everything
+        # each time, than to try and maintain state.
```
bmerry, line 93: You can probably speed this up a lot and also avoid sending a lot of sensor updates to clients by adding something like this (edit doesn't update the comment).
```suggestion
        # each time, than to try and maintain state.
        if reading is not None and old_reading is not None and reading.status == old_reading.status:
            return None  # Sensor didn't change state, so no change in overall device status
```

src/katgpucbf/utils.py
bmerry: > Is the 'behaviour' of a `device-status` sensor in an ICD somewhere?

Yes, it's in the [katcp spec](https://katcp-python.readthedocs.io/en/latest/_downloads/361189acb383a294be20d6c10c257cb4/NRF-KAT7-6.0-IFCE-002-Rev5-1.pdf), appendix B.2, although it's largely left up to the device to decide what it means. It doesn't allow for the status to be unknown though.

I don't think this case can realistically happen within katgpucbf because it would require all sensors to be UNKNOWN, and some of them are permanently NOMINAL. But more generally if it does happen my feeling is one should assume the device is OK until there is evidence to the contrary.

src/katgpucbf/fgpu/engine.py
bmerry: I'm with @amishatishpatel on this.

src/katgpucbf/utils.py
bmerry: You don't seem to use `_U` anywhere, so I'm guessing it is copy-pasted from aiokatcp.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+    it's useful for `UNKNOWN` to have the highest (or worst) value.
+    """
+
+    OK = 1
+    DEGRADED = 2
+    FAIL = 3
+    UNKNOWN = 4
+
+
+class DeviceStatusSensor(aiokatcp.AggregateSensor):
```
bmerry, line 76:
```suggestion
class DeviceStatusSensor(aiokatcp.AggregateSensor[DeviceStatus]):
```

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+    """Summary sensor for quickly ascertaining device status.
+
+    This takes its value from the worst status of its target set of sensors, so
+    it's quick to identify if there's something wrong, or if everything is good.
+    """
+
+    def __init__(self, target: aiokatcp.SensorSet):
+        super().__init__(target=target, sensor_type=DeviceStatus, name="device-status")
```
bmerry, line 84: Needs a description.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+
+
+class DeviceStatusSensor(aiokatcp.AggregateSensor):
+    """Summary sensor for quickly ascertaining device status.
+
+    This takes its value from the worst status of its target set of sensors, so
+    it's quick to identify if there's something wrong, or if everything is good.
+    """
+
+    def __init__(self, target: aiokatcp.SensorSet):
```
bmerry, line 83:
```suggestion
    def __init__(self, target: aiokatcp.SensorSet) -> None:
```

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,48 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+        super().__init__(target=target, sensor_type=DeviceStatus, name="device-status")
+
+    def update_aggregate(
+        self,
+        updated_sensor: Optional[aiokatcp.Sensor[_T]],
+        reading: Optional[aiokatcp.Reading[_T]],
+        old_reading: Optional[aiokatcp.Reading[_T]],
+    ) -> aiokatcp.Reading[DeviceStatus]:  # noqa: D102
```
bmerry, line 91:
```suggestion
    ) -> aiokatcp.Reading[DeviceStatus] | None:  # noqa: D102
```
For consistency with the base class.

src/katgpucbf/utils.py
```diff
@@ -20,14 +20,19 @@
 import logging
 import signal
 from asyncio import get_event_loop
+from enum import Enum
+from typing import Optional, TypeVar
```
bmerry, line 24: Rather remove Optional and use `T | None` for consistency with the rest of the katgpucbf codebase.

bmerry (changes requested):

src/katgpucbf/utils.py
james-smith-za: I'll have you know I typed it out manually while looking at the aiokatcp source!

I thought initially that I'd need both, but I guess I ended up not using the second one and I forgot to remove it.

src/katgpucbf/utils.py
james-smith-za: I needed the one in aiokatcp to be an IntEnum because I needed to use `max()` with it, which relies on comparison operators.

The one I'm using here takes the place of values, not statuses. So I don't need an IntEnum.

src/katgpucbf/utils.py
james-smith-za: Yeah this is a good point. I hadn't thought through the timestamp very much, I'm implementing your approach.

src/katgpucbf/utils.py
```diff
@@ -70,36 +70,48 @@ class DeviceStatus(Enum):
...
-        if worst_status == aiokatcp.Sensor.Status.WARN:
-            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.WARN, DeviceStatus.DEGRADED)
-        if worst_status == aiokatcp.Sensor.Status.UNKNOWN:  # Do we need to take this situation into account?
-            return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.UNKNOWN, DeviceStatus.UNKNOWN)
-        return aiokatcp.Reading(sensor.timestamp, aiokatcp.Sensor.Status.ERROR, DeviceStatus.FAIL)
+            agg_value = DeviceStatus.OK
+        elif worst_status == aiokatcp.Sensor.Status.WARN:
+            agg_value = DeviceStatus.DEGRADED
+        else:
+            agg_value = DeviceStatus.FAIL
```
bmerry, line 115: Since we're using Python 3.10 this could be a reasonable place to apply the new pattern matching :-)

src/katgpucbf/utils.py
```diff
@@ -70,36 +70,48 @@ class DeviceStatus(Enum):
...
-        old_reading: Optional[aiokatcp.Reading[_T]],
-    ) -> aiokatcp.Reading[DeviceStatus]:  # noqa: D102
-        # For device status it's far simpler just to re-calculate everything
-        # each time, than to try and maintain state.
-        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.UNKNOWN
+        updated_sensor: aiokatcp.Sensor[_T] | None,
+        reading: aiokatcp.Reading[_T] | None,
+        old_reading: aiokatcp.Reading[_T] | None,
+    ) -> aiokatcp.Reading[DeviceStatus] | None:  # noqa: D102
+
```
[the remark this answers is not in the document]
james-smith-za: I wonder why `black` didn't catch that 🤔

src/katgpucbf/utils.py
```diff
@@ -70,36 +70,48 @@ class DeviceStatus(Enum):
...
+        if reading is not None and old_reading is not None and reading.status == old_reading.status:
+            return None  # Sensor didn't change state, so no change in overall device status
+        # Otherwise, we have to recalculate.
+        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.NOMINAL
         for sensor in self.target.values():
             if self.filter_aggregate(sensor):
                 worst_status = max(worst_status, sensor.status)
 
+        if reading is not None and old_reading is not None:  # i.e. an update
+            timestamp = sensor.timestamp
```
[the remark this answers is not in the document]
james-smith-za: Oops, used the wrong variable. My bad.

I'm puzzled as to the motivation for the `assert` statement though. I prefer not to rely on `assert`s for any sort of program logic.

src/katgpucbf/utils.py
james-smith-za: I did think about that. For quite a while. But I ultimately didn't feel like arguing over whether it's a feature that should be used or avoided.

I do think the way I've phrased it now is more clear than what was there previously though.

src/katgpucbf/utils.py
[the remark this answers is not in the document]
james-smith-za: Ah, never mind. Mypy again.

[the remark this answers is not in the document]
bmerry: Sorry, thought I'd explained it but I think I lost the explanation while editing.

src/katgpucbf/utils.py
bmerry: Yes, it's definitely clear now. If there were more branches I'd suggest with pattern-matching or a lookup dictionary (using `.get` to provide a default for everything else), but with the current number of branches it's perfectly readable so I'm happy to leave it.

amishatishpatel (approved):
Looks good to me! I know Bruce's comment on [which timestamp to use](https://github.com/ska-sa/katgpucbf/pull/458#discussion_r1020248219) has been resolved, but what was the reasoning/motivation behind the use of `assert`? (I might have missed a trail of comments.)

bmerry:
> what was the reasoning/motivation behind the use of assert? (I might have missed a trail of comments.)

It tells mypy that `updated_sensor` can't be None in this code path. With it, mypy complains when you try to get attributes from it that don't exist on `None`.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,60 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
         return [(ep.host, ep.port) for ep in endpoints]
     except ValueError:
         return value
+
+
+class DeviceStatus(Enum):
+    """Discrete `device-status` readings.
+
+    Doesn't follow the convention of :class:`aiokatcp.Sensor.Status` because
+    it's useful for `UNKNOWN` to have the highest (or worst) value.
```
bmerry, line 67: The comment no longer matches the code.

src/katgpucbf/utils.py
```diff
@@ -53,3 +58,60 @@ def parse_source(value: str) -> list[tuple[str, int]] | str:
...
+    def update_aggregate(
+        self,
+        updated_sensor: aiokatcp.Sensor[_T] | None,
+        reading: aiokatcp.Reading[_T] | None,
+        old_reading: aiokatcp.Reading[_T] | None,
+    ) -> aiokatcp.Reading[DeviceStatus] | None:  # noqa: D102
+        if reading is not None and old_reading is not None and reading.status == old_reading.status:
+            return None  # Sensor didn't change state, so no change in overall device status
+        # Otherwise, we have to recalculate.
+        worst_status: aiokatcp.Sensor.Status = aiokatcp.Sensor.Status.NOMINAL
```
bmerry, line 96: Is it necessary to explicitly specify the type? I would think mypy would know the type based on the assigned value.

bmerry (changes requested):

src/katgpucbf/utils.py
james-smith-za: I might have started off with this not being obvious. I can remove it.

bmerry (approved):

PR #465, opened 2022-11-14
Rework DeviceStatusSensor with SimpleAggregateSensor

bmerry:
This simplifies the logic somewhat, and makes the timestamp logic the responsibility of aiokatcp. A functional change is that now if the worst status is beyond ERROR (e.g. UNREACHABLE), the aggregate status will be ERROR. That shouldn't have any effect since katgpucbf does not provide any such sensors.

This will fail to run right now because it needs a new release of aiokatcp that implements SimpleAggregateSensor.

Relates to NGC-442.

### DIFF
--- a/src/katgpucbf/utils.py
+++ b/src/katgpucbf/utils.py
@@ -19,8 +19,8 @@
 import ipaddress
 import logging
 import signal
-import time
 from asyncio import get_event_loop
+from collections import Counter
 from enum import Enum
 from typing import TypeVar
 
@@ -68,7 +68,7 @@ class DeviceStatus(Enum):
     FAIL = 3
 
 
-class DeviceStatusSensor(aiokatcp.AggregateSensor[DeviceStatus]):
+class DeviceStatusSensor(aiokatcp.SimpleAggregateSensor[DeviceStatus]):
     """Summary sensor for quickly ascertaining device status.
 
     This takes its value from the worst status of its target set of sensors, so
@@ -76,6 +76,8 @@ class DeviceStatusSensor(aiokatcp.AggregateSensor[DeviceStatus]):
     """
 
     def __init__(self, target: aiokatcp.SensorSet) -> None:
+        # We count the number of sensors with each possible status
+        self._counts: Counter[aiokatcp.Sensor.Status] = Counter()
         super().__init__(
             target=target, sensor_type=DeviceStatus, name="device-status", description="Overall engine health"
         )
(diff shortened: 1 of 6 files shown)


src/katgpucbf/utils.py
```diff
@@ -68,14 +68,16 @@ class DeviceStatus(Enum):
...
+class DeviceStatusSensor(aiokatcp.SimpleAggregateSensor[DeviceStatus]):
     """Summary sensor for quickly ascertaining device status.
 
     This takes its value from the worst status of its target set of sensors, so
     it's quick to identify if there's something wrong, or if everything is good.
     """
 
     def __init__(self, target: aiokatcp.SensorSet) -> None:
+        # We count the number of sensors with each possible status
+        self._counts: Counter[aiokatcp.Sensor.Status] = Counter()
```
james-smith-za, line 80: I did vaguely think of keeping lists / counters / some kind of track of sensors in various states. It seemed more complicated to do it this way though, not less.

james-smith-za (approved):
I don't personally feel as though this simplifies things substantially over what I had. You don't have to handle the timestamp, but you need to add two extra functions, so it ends up being a similar number of lines.

But I don't feel strongly enough about it either way to block it.

bmerry:
> You don't have to handle the timestamp, but you need to add two extra functions, so it ends up being a similar number of lines.

The functions are very simple and easy to follow though - you no longer need to think about what it's going to happen in each of the 4 different cases. So even if LOC stays the same I think the cyclomatic complexity is most likely reduced.

> I did vaguely think of keeping lists / counters / some kind of track of sensors in various states. It seemed more complicated to do it this way though, not less.

It's more scalable though: each sensor update takes O(1) time to process now, rather than O(N).

james-smith-za:
> 
> It's more scalable though: each sensor update takes O(1) time to process now, rather than O(N).

This is a fair point.

src/katgpucbf/utils.py
```diff
@@ -88,26 +90,23 @@ def update_aggregate(
...
+        return True
+
+    def aggregate_remove(self, sensor: aiokatcp.Sensor[_T], reading: aiokatcp.Reading[_T]) -> bool:  # noqa: D102
+        self._counts[reading.status] -= 1
+        return True
+
+    def aggregate_compute(self) -> tuple[aiokatcp.Sensor.Status, DeviceStatus]:  # noqa: D102
+        worst_status = max(
+            (status for status, count in self._counts.items() if count > 0), default=aiokatcp.Sensor.Status.NOMINAL
+        )
```
amishatishpatel, line 106: This is very cool.

amishatishpatel (approved):

bmerry:
Have now bumped the aiokatcp dependency to 1.5.0. I've requested re-reviews from both of you to double-check that I didn't miss anything, but I reckon I'll merge once either of you (rather than both) gets around to it rather than waiting for both.

james-smith-za:
> Have now bumped the aiokatcp dependency to 1.5.0. I've requested re-reviews from both of you to double-check that I didn't miss anything, but I reckon I'll merge once either of you (rather than both) gets around to it rather than waiting for both.

Just waiting for Jenkins to finish before I click approve.

bmerry:
> Just waiting for Jenkins to finish before I click approve.

I'll take that as permission to merge if Jenkins is happy.

james-smith-za (approved):

PR #470, opened 2022-11-16
Remove ability for device-status to return fail

james-smith-za:
I haven't thought of a good way to log things yet, though maybe the things that trigger warnings / errors in the sensors themselves should do the logging?

Checklist:

Closes NGC-843.

### DIFF
--- a/src/katgpucbf/utils.py
+++ b/src/katgpucbf/utils.py
@@ -106,7 +106,6 @@ class DeviceStatusSensor(aiokatcp.SimpleAggregateSensor[DeviceStatus]):
         )
         if worst_status <= aiokatcp.Sensor.Status.NOMINAL:  # NOMINAL or UNKNOWN
             return (aiokatcp.Sensor.Status.NOMINAL, DeviceStatus.OK)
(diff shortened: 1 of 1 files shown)


bmerry (approved):

bmerry:
Approved to get things going. It occurs to me that we don't actually have a test for this sensor though i.e. deliberately force a sensor into warn/error and check the effect on device-status.

james-smith-za:
> Approved to get things going. It occurs to me that we don't actually have a test for this sensor though i.e. deliberately force a sensor into warn/error and check the effect on device-status.

It's a good point, but I can't think of a good way to do that.
Perhaps it's something that should be covered in the qualification tests.

bmerry:
> > Approved to get things going. It occurs to me that we don't actually have a test for this sensor though i.e. deliberately force a sensor into warn/error and check the effect on device-status.
> 
> It's a good point, but I can't think of a good way to do that. Perhaps it's something that should be covered in the qualification tests.

Don't think it'll be too hard to do in a unit test. I'll file a ticket.

PR #474, opened 2022-11-18
Allow dsim to send to unicast addresses

bmerry:
If ibv is not enabled AND none of the endpoints is clearly a multicast address, don't try to pass TTL and interface address.

As a bonus, the descriptor sender now uses the same code for deciding how to construct the stream, and so will now use ibv when specified on the command line.

Closes NGC-846.


james-smith-za (approved):
Looks good to me. Will just need to test it before we turn it over to Wes and crew.

PR #479, opened 2022-11-29
Add sensors for average digitiser power

bmerry:
Add dig-pwr-dbfs sensors which report the average power in the digitiser signal for the last output chunk (with some caveats described in the design doc). These correspond to the `dig.pol?-rms-dbfs` in corr2; I removed "rms" from the name because (as described in NGC 801) it's a straight average of power, not root-mean-square of power.

It doesn't make any attempt to correctly handle missing data. I'll open a ticket about that before merging. I'm not sure how best to handle that case, since trying to take the average over just the good data will add a lot of complexity.

It's currently just a sensor (to match the functionality of the MK v7 ICD) with no Prometheus metric.

- [ ] Open a ticket for missing data handling
- [ ] Update the ICD draft

Relates to NGC-801. A katsdpcontroller is still needed to add the fake sensors and do the sensor renaming.

### DIFF
--- a/doc/fgpu.design.rst
+++ b/doc/fgpu.design.rst
@@ -418,6 +418,26 @@ transmission. It's important to (as far as possible) always run the back-end
 on the same amount of data, because cuFFT bakes the number of FFTs into its
 plan.
 
+Digitiser sample statistics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The PFB kernel also computes the average power of the incoming signal. Ideally
+that would be done by a separate kernel that processed each incoming sample
+exactly once. However, doing so would be expensive in memory bandwidth.
+Instead, we update statistics as samples are loaded for PFB calculations.
+
+Some care is needed to avoid double-counting due to overlapping PFB windows.
+The simplest way to add this to the existing code is that for each output
+spectrum, we include the last 2 × channels samples from the PFB window. In
+steady state operation and in the absence of coarse delay changes, this will
+count each sample exactly once. Coarse delay changes will cause some samples
+to be counted twice or not at all, but these are sufficiently rare that it is
+not likely to affect the statistics.
+
+Average power is updated at the granularity of output chunks. The PFB kernel
+updates a total power accumulator stored in the output item. This is performed
+using (64-bit) integer arithmetic, as this avoids the pitfalls of
+floating-point precision when accumulating a large number of samples.
+
 Network transmit
 ----------------
 The current transmit system is quite simple. A single spead2 stream is created,
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -18,6 +18,7 @@
 
 import asyncio
 import logging
+import math
 import numbers
 from collections import deque
 from collections.abc import Iterable, Sequence
@@ -28,7 +29,7 @@ import aiokatcp
 import katsdpsigproc.accel as accel
 import numpy as np
 import spead2.recv
-from katsdpsigproc.abc import AbstractContext, AbstractEvent
+from katsdpsigproc.abc import AbstractCommandQueue, AbstractContext, AbstractEvent
 from katsdpsigproc.resource import async_wait_for_events
 from katsdptelstate.endpoint import Endpoint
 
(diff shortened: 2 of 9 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -315,6 +336,14 @@ def format_complex(value: numbers.Complex) -> str:
     return f"{value.real}{value.imag:+}j"
 
 
+def dig_pwr_dbfs_status(value: float) -> aiokatcp.Sensor.Status:
+    """Compute status for dig-pwr-dbfs sensor.
+
+    TODO: the thresholds are inherited from MeerKAT. Are they what we want?
+    """
+    return aiokatcp.Sensor.Status.NOMINAL if -32.0 <= value <= -22.0 else aiokatcp.Sensor.Status.WARN
```
james-smith-za, line 344: I have a mild recommendation here to put these as constants in `meerkat.py`.

james-smith-za (approved):
> It doesn't make any attempt to correctly handle missing data. I'll open a ticket about that before merging. I'm not sure how best to handle that case, since trying to take the average over just the good data will add a lot of complexity.

My initial reaction to reading this was just to subtract the number of missing packets before you divide to get average power, but I see in the code that you're using only from the last tap in the PFB (presumably because this does the decimation for free). So taking into account whether *that specific* data is missing will be more challenging.

src/katgpucbf/fgpu/engine.py
bmerry: I've definitely up for making them constants, but I'm not sure if meerkat.py is the right place. The code outside of meerkat.py is supposed to be instrument-independent, and if it starts depending on meerkat.py then it isn't any more.

james-smith-za: Hmmm, fair point.
I'd need to think about it a bit more but I'm happy to leave this as it is for now.

test/fgpu/test_pfb.py
```diff
@@ -56,13 +57,14 @@ def test_pfb_fir(context: AbstractContext, command_queue: AbstractCommandQueue,
     rng = np.random.default_rng(seed=1)
     h_in = rng.integers(0, 256, samples * SAMPLE_BITS // BYTE_BITS, np.uint8)
     weights = rng.uniform(-1.0, 1.0, (2 * channels * taps,)).astype(np.float32)
-    expected = pfb_fir_host(h_in, channels, unzip_factor, weights)
+    expected, expected_total_power = pfb_fir_host(h_in, channels, unzip_factor, weights)
```
amishatishpatel, line 60: Not necessarily worth an update/further commit, but I would've liked to see an equivalent suffix update to the original `expected` - similar to how you've done `h_{out, total_power}`.

amishatishpatel (approved):
Really just the one comment on a line in the documentation. I have similar thoughts to the discussion on moving those constants to `meerkat.py`.

doc/fgpu.design.rst
```diff
@@ -418,6 +418,26 @@ transmission. It's important to (as far as possible) always run the back-end
 on the same amount of data, because cuFFT bakes the number of FFTs into its
 plan.
 
+Digitiser sample statistics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The PFB kernel also computes the average power of the incoming signal. Ideally
+that would be done by a separate kernel that processed each incoming sample
+exactly once. However, doing so would be expensive in memory bandwidth.
+Instead, we update statistics as samples are loaded for PFB calculations.
```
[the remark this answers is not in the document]
bmerry: No, I mean that as each sample is loaded, we use it to update statistics, rather than having a separate pass over the samples.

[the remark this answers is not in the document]
amishatishpatel: Gosh, sorry. Brain clearly isn't Englishing too well this afternoon.

src/katgpucbf/fgpu/engine.py
bmerry: I've added constants to katgpucbf.fgpu.

test/fgpu/test_pfb.py
bmerry: Done - I like how that looks.

bmerry:
Opened NGC-855 for handling missing data.

PR #486, opened 2022-12-05
Add X-engine chan-range sensor

james-smith-za:
I had to re-arrange a few of the initialisation things in the top of XBEngine.__init__(), and make populate_sensors no longer a static function, because it needed to access class members. But it seems to work.

- [X] (N/A) If dependencies are added/removed: update ``setup.cfg`` and appropriate requirements files
- [X] (N/A) If modules are added/removed: use sphinx-apidoc to update files in ``doc/``
- [X] Ensure copyright notices are present and up-to-date
- [X] (N/A) If qualification tests are changed: attach a sample qualification report
- [X] (N/A) If design has changed: ensure documentation is up to date

Addresses NGC-837

### DIFF
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -279,7 +279,6 @@ class XBEngine(DeviceServer):
         context: katsdpsigproc.abc.AbstractContext,
     ):
         super().__init__(katcp_host, katcp_port)
-        self.populate_sensors(self.sensors)
 
         if sample_bits != 8:
             raise ValueError("sample_bits must equal 8 - no other values supported at the moment.")
@@ -298,6 +297,9 @@ class XBEngine(DeviceServer):
         self.n_spectra_per_heap = n_spectra_per_heap
         self.sample_bits = sample_bits
         self.n_samples_between_spectra = n_samples_between_spectra
+        self.channel_offset_value = channel_offset_value
+
+        self.populate_sensors(self.sensors)
 
         self._src = src
         self._src_interface = src_interface
@@ -336,7 +338,6 @@ class XBEngine(DeviceServer):
             math.ceil(rx_reorder_tol / self.rx_heap_timestamp_step / self.heaps_per_fengine_per_chunk) + 1
         )
         n_free_chunks: int = self.max_active_chunks + 8  # TODO: Abstract this 'naked' constant
-        self.channel_offset_value = channel_offset_value
 
         self.monitor = monitor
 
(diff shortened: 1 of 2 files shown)


amishatishpatel (commented):
Hmmmm, I'm not sure if I'm complete on-board with the change away from `@staticmethod`.
* I mean, I understand if you need access to an attribute, just use the most straightforward solution; however,
* This does kind of affect our eventual goal of making a base `Engine` class - not that that should affect our short-term goal of actually doing work required now.

Keen to hear what Bruce thinks.

james-smith-za:
> This does kind of affect our eventual goal of making a base `Engine` class

I sort of gave up on that quite a while ago. The F- and XB-engines are different enough that trying to unify them too much may not be desirable.

In fact there may end up being other sensors which end up being needed which will need to access class members, so I have a feeling it was coming at some point anyway.

test/xbgpu/test_engine.py
```diff
@@ -458,6 +458,11 @@ async def test_xengine_end_to_end(
         timestamp_step = n_samples_between_spectra * n_spectra_per_heap
         missing_antennas = set() if missing_antenna is None else {missing_antenna}
 
+        assert (
+            xbengine.sensors["chan-range"].value
+            == f"({n_channels_per_stream*CHANNEL_OFFSET},{n_channels_per_stream*CHANNEL_OFFSET + n_channels_per_stream - 1})"  # noqa: E501
```
bmerry, line 463: Rather than noqa'ing the extra-long line, I suggest pulling out variables for the first and last channel.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -443,9 +444,19 @@ def __init__(
...
-    @staticmethod
-    def populate_sensors(sensors: aiokatcp.SensorSet) -> None:
+    def populate_sensors(self, sensors: aiokatcp.SensorSet) -> None:
         """Define the sensors for an XBEngine."""
+        # Static sensors
+        sensors.add(
+            aiokatcp.Sensor(
+                str,
+                "chan-range",
+                "The range of channels processed by this XB-engine",
```
bmerry, line 454:
```suggestion
                "The range of channels processed by this XB-engine, inclusive",
```

To be clear that the end channel is included (feel free to phrase it differently - this is what corr2 did)

bmerry (changes requested):
Just a couple of niggles. I'm fine with `populate_sensors` becoming non-static.

amishatishpatel (approved):

bmerry (approved):

PR #487, opened 2022-12-08
Add time synchronization sensors

bmerry:

- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match (ska-sa/katsdpcontroller#634)

Closes NGC-816.


src/katgpucbf/utils.py
```diff
@@ -184,6 +186,76 @@ def cancel(self) -> None:
             sensor.detach(self)
 
 
+def _esterror_status(value: float) -> aiokatcp.Sensor.Status:
```
amishatishpatel, line 189: Part of me wants to request a change to the variable/docstring-esque comment to remind us that `value` is in seconds.

*But*, I know the `units` are specified in the sensor itself. It's not like someone else is going to randomly use this (and `_maxerror_status`) for a difference `status_func`, right?

src/katgpucbf/utils.py
```diff
@@ -184,6 +186,76 @@ def cancel(self) -> None:
...
+    else:
+        return aiokatcp.Sensor.Status.ERROR
+
+
+def _maxerror_status(value: float) -> aiokatcp.Sensor.Status:
+    # maxerror is an over-estimate (it makes very conservative assumptions
+    # about network asymmetry and skew in the clock source). Use conservative
+    # thresholds to avoid warnings when there isn't a stratum 1 time source
+    # in the same data centre. Experimentally, maximum error is < 10ms
+    # when synchronising between Cape Town and the Karoo.
```
amishatishpatel, line 204: Appreciate the explanation, but perhaps prefix this with `NOTE:` ?

src/katgpucbf/utils.py
```diff
@@ -184,6 +186,76 @@ def cancel(self) -> None:
...
+        ),
+        "maxerror": aiokatcp.Sensor(
+            float, "time.maxerror", "Upper bound on time synchronisation error", units="s", status_func=_maxerror_status
+        ),
+        "state": aiokatcp.Sensor(aiokatcp.ClockState, "time.state", "Kernel clock state", status_func=_state_status),
+    }
+    for sensor in mapping.values():
+        sensors.add(sensor)
+
+    agg_sensor = aiokatcp.Sensor(bool, "time.synchronised", "Whether the host clock is synchronised within tolerances")
```
amishatishpatel, line 241: I don't mean to sound like we have a monopoly on what an `AggregateSensor` is, and I know it doesn't really matter what the variable is called (because it only exists here), _but_ ...
* Is there a 'better'/less ambiguous name?

If it doesn't come to you in a minute or so, don't bother changing it.

src/katgpucbf/utils.py
```diff
@@ -184,6 +186,76 @@ def cancel(self) -> None:
...
+    updater = aiokatcp.TimeSyncUpdater(mapping)
+
+    async def run() -> None:
+        while True:
+            updater.update()
+            good = all(sensor.status != aiokatcp.Sensor.Status.ERROR for sensor in mapping.values())
+            agg_sensor.set_value(
+                good,
+                status=aiokatcp.Sensor.Status.NOMINAL if good else aiokatcp.Sensor.Status.ERROR,
+                timestamp=mapping["state"].timestamp,
```
amishatishpatel, line 252: Do we use this timestamp because it's most likely the last to be updated?

src/katgpucbf/utils.py
```diff
@@ -184,6 +186,76 @@ def cancel(self) -> None:
...
+    # when synchronising between Cape Town and the Karoo.
+    if 0.0 <= value < 10e-3:
+        return aiokatcp.Sensor.Status.NOMINAL
+    elif value < 0.1:
+        return aiokatcp.Sensor.Status.WARN
+    else:
+        return aiokatcp.Sensor.Status.ERROR
+
+
+def _state_status(value: aiokatcp.ClockState) -> aiokatcp.Sensor.Status:
```
amishatishpatel, line 213: Promise I'm not being deliberately difficult, but what do you think about prepending all these `status_func`s with `_time_`? Reading `_state_status` last kinda threw me off (I know it's all self-explanatory below regarding what goes where).

amishatishpatel (changes requested):
Aside from the small comments here and there, I might request the abstraction of the naked constants used for thresholds in the `status_func`s - I'm sure they'll live in `utils.py` anyway?

Big fan of the creation/addition of `self._cancel_tasks`.

src/katgpucbf/utils.py
bmerry: Not sure I follow. Isn't every comment a note about how the code works?

src/katgpucbf/utils.py
bmerry: Renaming it to synchronised_sensor.

src/katgpucbf/utils.py
bmerry: The sensors in `mapping` will all be updated with the same timestamp. I just picked one of them to get that common timestamp.

src/katgpucbf/utils.py
bmerry: It's also just generally good practice to use base units internally and convert to things like milliseconds only for display, so in general I don't worry too much about documenting the units of things when they're in base units (which has lead to the code-base containing a mix of `adc_sample_rate` and `adc_sample_rate_hz`). On the other hand, `value` isn't a name that adds a lot of value (ha!), so I'll rename it to `seconds` to make it self-documenting.

src/katgpucbf/utils.py
amishatishpatel: Touché... But I do mean more in terms of these longer comments/reasoning. I see it's almost exclusively used in `xbgpu/engine.py`, if that helps show where my POV originates.

Again, fair point, so I concede the addition is not the most necessary change.

amishatishpatel (commented):
Before I explicitly approve, I see those threshold values are still shown as raw values (and not declared elsewhere/further up). I'll concede the comments alongside do go far enough to clarify why they're there.

I'll head over to the corresponding katsdpcontroller PR while James casts an eye over this.

bmerry:
> Before I explicitly approve, I see those threshold values are still shown as raw values (and not declared elsewhere/further up). I'll concede the comments alongside do go far enough to clarify why they're there.

Oh yes, I missed that point. Will address it now.

src/katgpucbf/utils.py
```diff
@@ -184,6 +186,84 @@ def cancel(self) -> None:
...
+    async def run() -> None:
+        while True:
+            updater.update()
+            good = all(sensor.status != aiokatcp.Sensor.Status.ERROR for sensor in mapping.values())
+            synchronised_sensor.set_value(
+                good,
+                status=aiokatcp.Sensor.Status.NOMINAL if good else aiokatcp.Sensor.Status.ERROR,
+                timestamp=mapping["state"].timestamp,
+            )
+            await asyncio.sleep(1)
```
james-smith-za, line 267: Why 1 second specifically?

amishatishpatel (approved):

src/katgpucbf/utils.py
amishatishpatel: Ah, thank you for catching this. This is why we have two reviewers!

bmerry: I don't recall that conversion. Why 5 seconds?

bmerry: The update process is fairly cheap (just a system call, not a bunch of networking to talk to a daemon), so I figure we might as well update it reasonably fast, and if people want a lower update rate they can use one of the rate-limiting sensor strategies.

james-smith-za: Maybe I'm getting confused with something else.
But is there a reason for selecting 1 second? Maybe just add a comment with the logic. The whole Chesterton's fence thing, so that a future reader of the code can see why it's there.

james-smith-za (approved):

PR #490, opened 2022-12-12
Set an auto strategy on some fast-updating sensors

bmerry:
This should prevent overloading the product controller. See NGC-859.

Closes NGC-859.

### DIFF
--- a/src/katgpucbf/__init__.py
+++ b/src/katgpucbf/__init__.py
@@ -43,9 +43,9 @@ DEFAULT_KATCP_HOST: Final = ""  # All interfaces
 DEFAULT_KATCP_PORT: Final = 7147
 DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
-# Update rate (expressed in seconds) for katcp sensors where the underlying value
-# may update extremely rapidly.
-SENSOR_UPDATE_TIME: Final = 1.0
+# Minimum update period (in seconds) for katcp sensors where the underlying
+# value may update extremely rapidly.
+MIN_SENSOR_UPDATE_PERIOD: Final = 1.0
 
 GPU_PROC_TASK_NAME: Final[str] = "GPU Processing Loop"
 RECV_TASK_NAME: Final[str] = "Receiver Loop"
--- a/src/katgpucbf/__init__.py
+++ b/src/katgpucbf/__init__.py
@@ -26,25 +26,23 @@ except PackageNotFoundError:
     # Package wasn't installed yet?
     __version__ = "unknown"
 
-__all__ = ["__version__"]
-
 BYTE_BITS: Final = 8
 COMPLEX: Final = 2
 N_POLS: Final = 2
 SPEAD_DESCRIPTOR_INTERVAL_S: Final = 5
 SEND_RATE_FACTOR = 1.1
-# Biggest power of 2 that fits in a jumbo MTU. A power of 2 isn't required but
-# it can be convenient to have packet boundaries align with the natural
-# boundaries in the payload (for antenna-channelised-voltage output). Bigger
-# is better to minimise the number of packets/second to process.
+#: Biggest power of 2 that fits in a jumbo MTU. A power of 2 isn't required but
+#: it can be convenient to have packet boundaries align with the natural
+#: boundaries in the payload (for antenna-channelised-voltage output). Bigger
+#: is better to minimise the number of packets/second to process.
 DEFAULT_PACKET_PAYLOAD_BYTES: Final = 8192
 DEFAULT_TTL: Final = 4  #: Default TTL for spead multicast transmission
 DEFAULT_KATCP_HOST: Final = ""  # All interfaces
 DEFAULT_KATCP_PORT: Final = 7147
 DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
-# Minimum update period (in seconds) for katcp sensors where the underlying
-# value may update extremely rapidly.
+#: Minimum update period (in seconds) for katcp sensors where the underlying
+#: value may update extremely rapidly.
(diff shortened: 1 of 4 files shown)


bmerry:
[report.pdf](https://github.com/ska-sa/katgpucbf/files/10209232/report.pdf)

src/katgpucbf/__init__.py
```diff
@@ -39,10 +39,13 @@
...
-DEFAULT_KATCP_HOST = ""  # All interfaces
-DEFAULT_KATCP_PORT = 7147
-DIG_HEAP_SAMPLES = 4096
+DEFAULT_KATCP_HOST: Final = ""  # All interfaces
+DEFAULT_KATCP_PORT: Final = 7147
+DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
+# Update rate (expressed in seconds) for katcp sensors where the underlying value
+# may update extremely rapidly.
+SENSOR_UPDATE_TIME: Final = 1.0
```
amishatishpatel, line 48: I appreciate the comment does explain it clearly enough, but the variable/constant name still doesn't sit quite right with me (rate vs `_TIME`). Bikeshedding, you might rightly say.

src/katgpucbf/fgpu/recv.py
```diff
@@ -230,13 +231,17 @@ def make_sensors(sensor_timeout: float) -> aiokatcp.SensorSet:
                 "The timestamp (in samples) of the last chunk of data received from the digitiser",
                 default=-1,
                 initial_status=aiokatcp.Sensor.Status.ERROR,
+                auto_strategy=aiokatcp.SensorSampler.Strategy.EVENT_RATE,
+                auto_strategy_parameters=(SENSOR_UPDATE_TIME, math.inf),
```
amishatishpatel, line 235: My main question is around what does the `math.inf` get interpreted as? Variable name aside, tracing the usage down to the [SensorSampler._send_update](https://github.com/ska-sa/aiokatcp/blob/master/src/aiokatcp/sensor.py#L368-L382)
* Specifically, `max(self.loop.time(), sched_time + self.longest)`

amishatishpatel (commented):
Just the one comment really before I explicitly Approve.

src/katgpucbf/__init__.py
bmerry: What would you prefer to call it? Or is your beef with the use of "rate" in the comment?

src/katgpucbf/fgpu/recv.py
bmerry: It means that if the sensor value doesn't change, it won't spontaneously send an update to the client.

src/katgpucbf/__init__.py
amishatishpatel: It's not quite an 'update **period**', is it? Because it's used as some kind of 'minimum time'? Or rather, the shortest time before an update of the sensor is sent out.

But it's likely I'm not fully understanding, you must have some reasoning as to why 'rate' is appropriate here? (No beef, just seeking clarity)

bmerry: I guess `period` would be better than `rate`, since it's in units of seconds rather than Hz. For the rx sensors it's a minimum period: if the underlying sensor updates less frequently than this then updates will be sent when the underlying sensor updates. For the time sync sensors it is just the period, since we have no visibility of the updates to the underlying kernel variables.

bmerry: @amishatishpatel See what you think of the update I just pushed.

src/katgpucbf/utils.py
```diff
@@ -264,10 +264,7 @@ async def run() -> None:
                 status=aiokatcp.Sensor.Status.NOMINAL if good else aiokatcp.Sensor.Status.ERROR,
                 timestamp=mapping["state"].timestamp,
             )
-            # 1 second is pretty arbitrary. It's probably faster than
-            # necessary, but subscribers can always use a rate-limiting
-            # sampling strategy to reduce the rate of updates.
```
amishatishpatel, line 269: Ah, comment was no longer necessary.

amishatishpatel (approved):
Excellent, please and thank you.

src/katgpucbf/__init__.py
```diff
@@ -39,10 +39,13 @@
...
-DEFAULT_KATCP_HOST = ""  # All interfaces
-DEFAULT_KATCP_PORT = 7147
-DIG_HEAP_SAMPLES = 4096
+DEFAULT_KATCP_HOST: Final = ""  # All interfaces
+DEFAULT_KATCP_PORT: Final = 7147
+DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
+# Minimum update period (in seconds) for katcp sensors where the underlying
+# value may update extremely rapidly.
+MIN_SENSOR_UPDATE_PERIOD: Final = 1.0
```
james-smith-za, line 48: If I want to be *really* pedantic (and occasionally I do), I'd recommend that the unit be included in the name. I see you have documented it in the comments, so I'm not going to make an issue of it.

I see that this file (and subsequently these constants) don't seem to make their way into whatever Sphinx outputs, at least I can't find them. That may be something worth thinking about.

james-smith-za (approved):

src/katgpucbf/__init__.py
bmerry: > If I want to be really pedantic (and occasionally I do), I'd recommend that the unit be included in the name.

I can see merit in that where the unit isn't obvious e.g. not the standard SI unit. If I see a variable holding time/frequency/length/power, I assume it is in seconds/Hertz/metres/watts.

bmerry: > I see that this file (and subsequently these constants) don't seem to make their way into whatever Sphinx outputs, at least I can't find them. That may be something worth thinking about.

Fixed now.

james-smith-za (approved):

PR #492, opened 2022-12-13
Remove katsdptelstate dependency

james-smith-za:
This I accomplish by plagiarising the `endpoint.py` file (and its associated tests). I do remove a bunch of functionality that we don't use though.

This removes quite a lot from the requirements.

- [x] If modules are added/removed: use sphinx-apidoc to update files in ``doc/``

Closes NGC-862

### DIFF
--- a/doc/katgpucbf.endpoint.rst
+++ b/doc/katgpucbf.endpoint.rst
@@ -1,0 +1,7 @@
+katgpucbf.endpoint module
+======================
+
+.. automodule:: katgpucbf.endpoint
+   :members:
+   :undoc-members:
+   :show-inheritance:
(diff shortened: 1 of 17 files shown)


james-smith-za:
Except for `endpoint.py` and `test_endpoint.py`, it would be about `+13 - 55` by my reckoning and those removed lines are mostly from the various requirements files.

If I'm understanding what I'm seeing correctly, it ends up making almost zero difference to the size of the docker image (which is 5.41 GB!). I had expected it to be at least a measurable few MB smaller.

I'll have to think about whether it's worth the effort to spin `endpoint.py` out into a whole new module.

PR #499, opened 2023-01-12
NGC-860 Stop correlator if still running after qualification tests

ijassiem:
With this change, a check is performed at the end of the Jenkins qualification test to see if there is an associated correlator still running. If this is the case, it is usually due to some fault or error, and the correlator is stopped.

Closes NGC-860

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -89,8 +89,9 @@ pipeline {
   /* This post stage is configured to always run at the end of this pipeline,
    * regardless of the completion status. In this stage an email is sent to
    * the specified address with details of the Jenkins job and a PDF report
-   * containing the qualification test results. The final step removes the workspace when
-   * the build is complete.
+   * containing the qualification test results. Any runnning correlator part
+   * of the Jenkins qualification test that continues to run at this point will be
+   * stopped. The final step removes the workspace when the build is complete.
    */
   post {
     always {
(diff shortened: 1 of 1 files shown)


qualification/Jenkinsfile
```diff
@@ -107,6 +108,14 @@ pipeline {
       subject: 'Qualification Tests - katgpucbf - $BUILD_STATUS!',
       to: '$DEFAULT_RECIPIENTS'
 
+      script {
+        ARRAYS = sh(script: 'katcpcmd lab5.sdp.kat.ac.za:5001 product-list', returnStdout: true).trim()
+        echo "Currently Running Arrays: ${ARRAYS}"
+        if (ARRAYS.contains("jenkins_qualification_correlator")) {
+          sh 'katcpcmd lab5.sdp.kat.ac.za:5001 product-deconfigure jenkins_qualification_correlator 1'
+          echo "Destroying jenkins_qualification_correlator"
```
james-smith-za, line 116: One nit-pick.
I'd do the `echo` before actually issuing the command.
But this is really just a personal preference thing, so feel free to ignore.

james-smith-za (approved):
I like it!

qualification/Jenkinsfile
bmerry: I agree with James: it should be first. By the time you get here, it's not destroying, it's already destroyed.

bmerry (changes requested):
Apart from James' nitpick, the copyright notice should be updated to include 2023.

qualification/Jenkinsfile
ijassiem: Agreed

bmerry (approved):

PR #501, opened 2023-01-16
Amortise some unpack calculations

bmerry:
A lot of the work in figuring out how to unpack in the PFB kernel is fixed for each thread (because the step size is a multiple of 8). So it can be precomputed and stored in an unpack_t struct.

This is a pretty minor optimisation since the kernel is not compute-bound, but it does improve things by about 1%. It may also facilitate some future optimisations.


james-smith-za (commented):
Looks reasonable to me. 1% is 1%.
This will make supporting other bit-widths easier as well, it seems.

james-smith-za (approved):

PR #502, opened 2023-01-16
Optimise the PFB FIR

bmerry:
The "clever" approach to writing samples into a circular buffer using modular arithmetic (and relying on the compiler to collapse it unrolling the loop) seems to perform worse than just shunting all the samples one along every iteration. I suspect in the latter case the compiler eliminates a lot of the copies anyway. It also seems performance is improved by letting the compiler decide on the loop unrolling instead of forcing a specific factor.

Also speed up the dot product fractionally by initialising it with the first product, rather than with zero.

Fully spell out a couple of types just for consistency.

This leads to an 8% speedup in some cases.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -614,7 +614,9 @@ class Engine(aiokatcp.DeviceServer):
         for pol, stream in enumerate(self._src_streams):
             for _ in range(src_chunks_per_stream):
                 if self.use_vkgdr:
-                    device_bytes = self._compute.slots[f"in{pol}"].required_bytes()
+                    slot = self._compute.slots[f"in{pol}"]
+                    assert isinstance(slot, accel.IOSlot)
+                    device_bytes = slot.required_bytes()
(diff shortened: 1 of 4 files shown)


james-smith-za (approved):

james-smith-za:
I'm not on the VPN at the moment but why is Jenkins unhappy?

bmerry:
> I'm not on the VPN at the moment but why is Jenkins unhappy?

Looks like maybe it timed out? I'll investigate.

bmerry:
> I'm not on the VPN at the moment but why is Jenkins unhappy?

It looks like there is an error in the engine tests that causes a worker process to fail and the test to hang. My guess is it's related to the padding but it's one of those where I don't get a sensible backtrace, so I'll need to dig into it a bit.

bmerry:
> I'm not on the VPN at the moment but why is Jenkins unhappy?

Hopefully fixed now. I guess check back in 10 minutes.

james-smith-za (approved):
Looks happy now.

PR #503, opened 2023-01-18
raw_antenna_voltage -> baseband_voltage

amishatishpatel:
As per the (draft) CBF-CAM ICD, the stream type name for raw digitiser data has been updated.

I've attached an example `product_config.json` to show the update - GitHub doesn't support JSON so I renamed to TXT.
- [sim_baseband.txt](https://github.com/ska-sa/katgpucbf/files/10443651/sim_baseband.txt)

Closes NGC-861.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -249,7 +249,7 @@ async def _correlator_config_and_description(
             name = f"{dsim_name}{pol}"
             dig_names.append(name)
             config["outputs"][name] = {
-                "type": "sim.dig.raw_antenna_voltage",
+                "type": "sim.dig.baseband_voltage",
                 "band": band,
                 "adc_sample_rate": adc_sample_rate,
                 "centre_frequency": centre_frequency,
--- a/scratch/populate_telstate_for_sdp.py
+++ b/scratch/populate_telstate_for_sdp.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 ################################################################################
-# Copyright (c) 2021-2022, National Research Foundation (SARAO)
+# Copyright (c) 2021-2023, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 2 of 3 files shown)


james-smith-za (approved):
FWIW you can add JSON just inline in a comment:
```json
{
  "Like this for instance",
}
```

This looks good to me. It'll probably need the updates to katsdpcontroller pushed, but I'd quite like a qualification test just to make sure we haven't missed anything.
Chances are low, grep doesn't find anything, but it never hurts to make sure.

amishatishpatel:
> FWIW you can add JSON just inline in a comment:
> 
> ```json
> {
>   "Like this for instance",
> }
> ```
Indeed indeed, I figured the PR blurb is busy enough (and for good reason). Might as well attach the JSON config as a file.

> This looks good to me. It'll probably need the updates to katsdpcontroller pushed, but I'd quite like a qualification test just to make sure we haven't missed anything. Chances are low, grep doesn't find anything, but it never hurts to make sure.
Indeed once again. I did a qualification run with the `--image-overrides` for both katgpucbf and katsdpcontroller.
- i.e. ` pytest -v --image-override katgpucbf:harbor.sdp.kat.ac.za/cbf/katgpucbf:NGC-861-baseband-voltage-rename --image-override katsdpcontroller:katsdpcontroller:NGC-861-baseband-voltage-rename`

No complaints from any of the qualification tests. Should I attach a report? I didn't initially as none of the tests were actually changed.

james-smith-za:
> No complaints from any of the qualification tests. Should I attach a report? I didn't initially as none of the tests were actually changed.

It's fine, if the run went successfully then we're good.

Just hang on before merging this, until we've gotten input from @bmerry and @ludwigschwardt on https://github.com/ska-sa/katsdpcontroller/pull/654.

amishatishpatel:
As per the discussion during this morning's MK+ check-in, the change here is according to the agreed upon CBF-CAM ICD.
* This change will form part of the continuing development of the 3.1 product-configure schema as per ska-sa/katsdpcontroller#654.

PR #504, opened 2023-01-18
Tweak the number of workitems for the PFB FIR

bmerry:
The new heuristic tries to balance the need for enough work to saturate the GPU with the need to give each workitem enough to do to amortise its startup cost. I've done benchmarking on a 3070 Ti, and the new scheme is improves things by up to 8.9%. The worst regression is 5.1%, but that is for 32-taps, 4-bit input which isn't relevant for MeerKAT. For 16-tap, 10-bit input the worst regression is 1.13%, and the best improvement is 4.4%.

### DIFF
--- a/src/katgpucbf/fgpu/pfb.py
+++ b/src/katgpucbf/fgpu/pfb.py
@@ -182,17 +182,17 @@ class PFBFIR(accel.Operation):
         if self.out_offset + self.spectra > out_buffer.shape[0]:
             raise IndexError("Output buffer does not contain sufficient spectra")
 
-        # We divide up the work so that there are sufficient threads to keep the GPU busy.
-        # Aim for 256K workitems i.e. step * (out_n / stepy) == 256K
-        # See pfb_fir.mako's comments for more info.
-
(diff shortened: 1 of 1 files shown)


james-smith-za (approved):

james-smith-za:
> For 16-tap, 10-bit input the worst regression is 1.13%, and the best improvement is 4.4%.

Over a range of different channel sizes?

The update to the in-code commentary made sense, is there anything in `doc/fgpu.design.rst` that needs to be updated?

bmerry:
Hold off reviewing for now - it seems this is causing some test to fail.

bmerry:
> > For 16-tap, 10-bit input the worst regression is 1.13%, and the best improvement is 4.4%.
> 
> Over a range of different channel sizes?

Yes - 1K, 4K, 8K and 32K.

> The update to the in-code commentary made sense, is there anything in `doc/fgpu.design.rst` that needs to be updated?

I checked, but it doesn't go into sufficient detail to need a change.

I've found the bug I introduced; will push a fix now.

james-smith-za (approved):

PR #505, opened 2023-01-19
Add facility to specify document number on the qualification report

james-smith-za:
In this PR I add the ability to specify the document number for the output report, as a CLI option when running the qualification tests. I figured that an actual release may be something that we want to do manually, or if we do it with Jenkins then it'll be easier to pass along as a CLI option than as a parameter in an ini file.

I used the following command to generate the report (scroll right to see the relevant argument):
```
spead2_net_raw pytest -v qualification --image-override katgpucbf:harbor.sdp.kat.ac.za/cbf/katgpucbf:latest --report-docnum=E1200-foo-bar
```

Attached here: 
[report.pdf](https://github.com/ska-sa/katgpucbf/files/10456552/report.pdf)

**Note**: There is also some LaTeX re-formatting in this PR, latexindent was a bit too eager to help with this, and I was too lazy to make the one-line change that I needed to do in the preamble, in another editor. But the formatting changes are all in a single commit, so you can easily exclude that from the review.

Closes NGC-874.


bmerry:
A high-level question before I look at it line-by-line: is there a particular reason you want to bake in the report number at the time you run the test? I would imagine wanting to assign a number to a report at the time the report is generated, even if it's for a test we ran earlier.

james-smith-za:
> A high-level question before I look at it line-by-line: is there a particular reason you want to bake in the report number at the time you run the test? I would imagine wanting to assign a number to a report at the time the report is generated, even if it's for a test we ran earlier.

This is a fair point. I guess I could make it a cli option for the PDF generator.

james-smith-za:
Moved the functionality to `generate_pdf.py`, now run with the command: 
```
qualification/report/generate_pdf.py --report-docid=E1200-bar-foo report.json report.pdf
```

[report.pdf](https://github.com/ska-sa/katgpucbf/files/10457125/report.pdf)

qualification/report/generate_pdf.py
```diff
@@ -875,11 +876,17 @@ def main():
     parser.add_argument("input", help="Output of pytest --report-log=...")
     parser.add_argument("pdf", help="PDF file to write")
     parser.add_argument("-c", "--commit-id", action="store_true", help="Output commit ID of katgpucbf image")
+    parser.add_argument(
+        "--report-docid",
```
bmerry, line 880:
```suggestion
        "--report-doc-id",
```
For consistency with `--commit-id` (and it being treated as two words in variable names).

bmerry (approved):
Looks good. Do we need something similar for revision? Or will we just churn out a new document ID each time, always at rev 1?

PR #507, opened 2023-01-20
Generate test procedure alongside report

james-smith-za:
I do this by just making two documents in parallel, and only putting the "steps" and not the "detail" into the procedure document. Unfortunately it is not possible to generate the procedure without actually running the test, but we can perhaps look at that as a feature and not a bug, because you won't be able to generate a "procedure" for tests that don't actually work. Though you could fudge things by just commenting stuff out I guess.

It could probably use a little bit of polish, but I'm fairly pleased with how this is structured so far. 

[test_report.pdf](https://github.com/ska-sa/katgpucbf/files/10466443/test_report.pdf)
[test_procedure.pdf](https://github.com/ska-sa/katgpucbf/files/10466444/test_procedure.pdf)

This may be easier to review on a per-commit basis, as I had to do a fair amount of refactoring first.

Closes NGC-878.

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -77,8 +77,8 @@ pipeline {
              sh 'spead2_net_raw pytest -v -c qualification/pytest-jenkins.ini qualification --image-override katgpucbf:harbor.sdp.kat.ac.za/cbf/katgpucbf:latest'
            }
            finally {
-             COMMIT_ID = sh(script: 'qualification/report/generate_pdf.py report.json report.pdf -c', returnStdout: true)
-             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report', reportDir: '', reportFiles: 'report.pdf, report.json', reportTitles: 'report.pdf, report.json'])
+             COMMIT_ID = sh(script: 'qualification/report/generate_pdf.py report.json test -c', returnStdout: true)
+             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report / Procedure', reportDir: '', reportFiles: 'test_report.pdf, report.json', reportTitles: 'test_report.pdf, report.json'])
              currentBuild.displayName = "#${BUILD_NUMBER} (${COMMIT_ID})"
            }
          }
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -78,7 +78,7 @@ pipeline {
            }
            finally {
              COMMIT_ID = sh(script: 'qualification/report/generate_pdf.py report.json test -c', returnStdout: true)
-             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report / Procedure', reportDir: '', reportFiles: 'test_report.pdf, report.json', reportTitles: 'test_report.pdf, report.json'])
+             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report / Procedure', reportDir: '', reportFiles: 'test_procedure.pdf test_report.pdf, report.json', reportTitles: 'test_report.pdf, report.json'])
              currentBuild.displayName = "#${BUILD_NUMBER} (${COMMIT_ID})"
            }
          }
--- a/qualification/report/generate_pdf.py
+++ b/qualification/report/generate_pdf.py
@@ -37,6 +37,7 @@ from docutils.core import publish_parts
 from pylatex import (
     Command,
     Document,
+    Enumerate,
     LongTable,
     MiniPage,
     MultiColumn,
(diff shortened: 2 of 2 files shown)


bmerry:
I'll start poring over the details next, but here's a devil's advocate thought: rather than having generate_pdf generate both reports at once, what about having it take a command-line argument to control whether it generates the procedure or the report, and call it twice in Jenkins? The advantages are:
- The code probably changes less, since you're still generating one document, just making some code conditional.
- Users can choose to only generate the document that they need.
- Can have fewer command-line arguments e.g. instead of `--report-doc-id` and `--procedure-doc-id`, you just need a `--doc-id`.

qualification/Jenkinsfile
```diff
@@ -77,8 +77,8 @@ pipeline {
              sh 'spead2_net_raw pytest -v -c qualification/pytest-jenkins.ini qualification --image-override katgpucbf:harbor.sdp.kat.ac.za/cbf/katgpucbf:latest'
            }
            finally {
-             COMMIT_ID = sh(script: 'qualification/report/generate_pdf.py report.json report.pdf -c', returnStdout: true)
-             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report', reportDir: '', reportFiles: 'report.pdf, report.json', reportTitles: 'report.pdf, report.json'])
+             COMMIT_ID = sh(script: 'qualification/report/generate_pdf.py report.json test -c', returnStdout: true)
+             publishHTML(target: [keepAll: true, reportName: 'Qualification Test Report / Procedure', reportDir: '', reportFiles: 'test_procedure.pdf test_report.pdf, report.json', reportTitles: 'test_report.pdf, report.json'])
```
bmerry, line 81: Put the test report first in reportFiles, since that's the one that shows up when you click the link in Jenkins.

reportTitles needs the test procedure added (it should have the same length as reportFiles).

qualification/report/generate_pdf.py
```diff
@@ -886,9 +913,11 @@ def main():
     result_list = list_from_json(args.input)
     if args.commit_id:
         print(test_image_commit(result_list))
-    doc = document_from_list(result_list, args.report_doc_id)
+    procedure_doc, report_doc = document_from_list(result_list, args.report_doc_id, "E1200-0000-004")
     if args.pdf.endswith(".pdf"):
         args.pdf = args.pdf[:-4]  # Strip .pdf suffix, because generate_pdf appends it
+    report_name = args.pdf + "_report"
+    procedure_name = args.pdf + "_procedure"
```
bmerry, line 920: I'm not really happy with this. If you stick with having this script generate both reports in one go, have two command-line arguments so that the user gets to pick explicitly what to call both files.

bmerry (changes requested):

james-smith-za:
> a devil's advocate thought: rather than having generate_pdf generate both reports at once, what about having it take a command-line argument to control whether it generates the procedure or the report, and call it twice in Jenkins? 

That's an interesting idea.
In my mind, the advantage to such a thing would be to generate a procedure doc without actually having run the tests.

But you need the entire JSON file in order to generate the procedure as well, so there seems to be little advantage to being able to do one without the other.

If we had started with that kind of approach in mind, and structured the framework in order to be able to generate a procedure without actually running the tests, there might be that kind of advantage.

bmerry:
> structured the framework in order to be able to generate a procedure without actually running the tests

I did think about that at the time you were setting things up, but I couldn't see a viable way of doing that while still allowing the documentation of the procedure to be interleaved into the code, particularly when it is more than just straight-line code e.g. parts of procedures defined in helper functions, loops etc.

One other thought that occurs to me is that if any test failed, we should maybe refuse to generate a test procedure, since it is potentially incomplete (unless we can identify that the failure was in a `check` rather than an `assert` or some other exception).

james-smith-za:
It may have been possible at the expense of more complexity, so this after-the-fact kludge to get the procedure is probably a fair tradeoff for simpler test code.

I'm not terribly worried about suppressing the procedure doc output in the case of a failed test at this stage. Perhaps I can make the `--procedure-doc-id` argument optional, and only output the procedure doc if it's actually given?

tyronevb:
@james-smith-za , if I understood Bruce's suggestion correctly, the tests are still required to be run to generate the test procedure, but we're just spitting out one document? And one of the main advantages is less code that needs to be maintained (only a single piece of logic for document generation)?

Based on this understanding:
- I'm all for the solution that requires the least amount of maintenance and doesn't depend developers knowing they need to make updates in multiple places
- The idea of being able to identify broken/failing tests is interesting - but instead of just omitting the test from the procedure, can it be flagged with a note?
- Following the above - how do we distinguish between broken tests (i.e. incomplete or buggy implementation) and failing tests (correlator not meeting spec) for the purposes of report generation? I'm not opposed to having the test procedure indicate which tests are failing, since it'll be useful to know this upfront when preparing for formal qualification.

qualification/report/generate_pdf.py
```diff
@@ -886,9 +913,11 @@ def main():
     result_list = list_from_json(args.input)
     if args.commit_id:
         print(test_image_commit(result_list))
-    doc = document_from_list(result_list, args.report_doc_id)
+    procedure_doc, report_doc = document_from_list(result_list, args.report_doc_id, "E1200-0000-004")
```
amishatishpatel, line 916: Minor compared to other observations, but perhaps declare these document number constants elsewhere. It looks particularly odd here where you've got the cmd-line arg with a default, then this 'hardcoded' arg here.

amishatishpatel (changes requested):
Beaten to the punch by Tyrone's suggestions/questions above - his third point was my question exactly.

james-smith-za:
Closing this one, will try with a different branch. I think it'll be easier for me to implement from main rather than trying to fix what I've got here.

PR #508, opened 2023-01-24
Support multiple network interfaces

bmerry:
Allow `--src-interface` to take up to two interfaces (one per pol), and `--dst-interface` to take any number of interfaces that divides into the number of endpoints. This allows bandwidths to scale beyond the amount provided by a single network port.

There are also a couple of scratch scripts I'm using to test high bandwidth operation.

### DIFF
--- a/doc/fgpu.design.rst
+++ b/doc/fgpu.design.rst
@@ -450,12 +450,22 @@ floating-point precision when accumulating a large number of samples.
 
 Network transmit
 ----------------
-The current transmit system is quite simple. A single spead2 stream is created,
-with one substream per multicast destination. For each output chunk, memory
-together with a set of heaps is created in advance. The heaps are carefully
-constructed so that they reference numpy arrays (including for the timestamps),
-rather than copying data into spead2. This allows heaps to be recycled for new
-data without having to create new heap objects.
+The current transmit system is quite simple. By default a single spead2 stream
+is created, with one substream per multicast destination. For each output
+chunk, memory together with a set of heaps is created in advance. The heaps are
+carefully constructed so that they reference numpy arrays (including for the
+timestamps), rather than copying data into spead2. This allows heaps to be
+recycled for new data without having to create new heap objects.
+
+If the traffic for a single engine exceeds the bandwidth of the network
+interface, it is necessary to distribute it over multiple interfaces. In this
+case, several spead2 streams are created (one per interface). Each of them has
+a substream for every multicast destination, but they are not all used (the
+duplication simplifies indexing). When heaps are transmitted, a stream is
+selected for each heap to balance the load. Descriptors and stop heaps are
+just sent through the first stream for simplicity. This scheme assumes that
+all the interfaces are connected to the same network and hence it does not
+matter which interface is used other than for load balancing.
 
 PeerDirect
 ^^^^^^^^^^
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -364,7 +364,7 @@ class Engine(aiokatcp.DeviceServer):
     srcs
         A list of source endpoints for the incoming data.
     src_interface
-        IP address of the network device to use for input.
+        IP addresses of the network devices to use for input.
     src_ibv
         Use ibverbs for input.
     src_affinity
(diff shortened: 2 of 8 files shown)


src/katgpucbf/fgpu/send.py
```diff
@@ -150,17 +172,24 @@ def _inc_counters(frame: Frame, future: asyncio.Future) -> None:
...
+        sensors: SensorSet,
     ) -> None:
-        """Transmit heaps on a SPEAD stream.
+        """Transmit heaps over SPEAD streams.
 
-        Frames from 0 to `frames` - 1 are sent asynchronously.
+        Frames from 0 to `frames` - 1 are sent asynchronously. The contents of
+        each frames are distributed over the streams. If the number of streams
+        does not divide into the number of destination endpoints, there will be
+        imbalances, because the partitioning is the same for every frame.
```
james-smith-za, line 186: i.e. so it'll work, but you may not get the performance you expect?

james-smith-za (approved):
I see what you mean about an increase in complexity, but I don't see it as a bad tradeoff in this case.

While it's unlikely that we'll end up seriously using this, given that we've specced a single 200G NIC, I feel it's useful to have the ability. Especially given your research.

src/katgpucbf/fgpu/send.py
bmerry: If the imbalance is more than a few percent I think it won't work, because each of the N streams is rate-limited to 1/N of the total rate (plus a safety margin). So if there is an imbalance then the most-loaded stream will get heaps in faster than it's allowed to send them out. main.py has a safety check to prevent this happening.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1264,8 +1264,8 @@ async def _run_transmit(self, stream: "spead2.send.asyncio.AsyncStream") -> None
                 pass  # It's already logged by the chunk_finished callback
         stop_heap = spead2.send.Heap(send.FLAVOUR)
         stop_heap.add_end()
-        for substream_index in range(stream.num_substreams):
-            await stream.async_send_heap(stop_heap, substream_index=substream_index)
+        for substream_index in range(streams[0].num_substreams):
+            await streams[0].async_send_heap(stop_heap, substream_index=substream_index)
```
amishatishpatel, line 1268: It's likely that my spead2 fog is hindering my view, but why are we/you only sending the stop heap on one of the streams (instead of `each stream in streams`)?

amishatishpatel (commented):
Before I explicitly approve, just the one comment. My `sh fu` isn't strong, so can't really comment on your scratch scripts (apologies).

src/katgpucbf/fgpu/engine.py
bmerry: substream_index is still iterating over all the multicast groups. I *could* do something like `_multi_send` and partition the groups over the streams, but since it's just one heap I don't really care which network interface is used.

amishatishpatel: I see, I see.

amishatishpatel (approved):

bmerry:
Thanks for the reviews. I noticed that the design doc needed updating as well, so I've requested a re-review to add a paragraph.

doc/fgpu.design.rst
```diff
@@ -450,12 +450,22 @@ floating-point precision when accumulating a large number of samples.
...
+carefully constructed so that they reference numpy arrays (including for the
+timestamps), rather than copying data into spead2. This allows heaps to be
+recycled for new data without having to create new heap objects.
+
+If the traffic for a single engine exceeds the bandwidth of the network
+interface, it is necessary to distribute it over multiple interfaces. In this
+case, several spead2 streams are created (one per interface). Each of them has
+a substream for every multicast destination, but they are not all used (the
+duplication simplifies indexing). When heaps are transmitted, a stream is
+selected for each heap to balance the load. Descriptors and stop heaps are
```
amishatishpatel, line 465: > When heaps are transmitted, a stream is selected for each heap to balance the load.

While this is technically true, I've rather been reading `_multi_send` logic as 'alternate/round-robin across all streams available for sending'. Not saying you should change it; perhaps your wording is simpler (principle of least astronishment and all that).

amishatishpatel (approved):
My apologies, I remember seeing the documentation checkbox empty when seeing the PR. Thank you for the reminder to review.

james-smith-za (approved):

doc/fgpu.design.rst
bmerry: I didn't want the design doc to get into the minutiae of how the selection is done.

PR #510, opened 2023-01-27
Generate either a QTP or a QTR using the json outcome of qualification

james-smith-za:

Closes NGC-878.


james-smith-za:
The reviewing goes much more easily if you hide the whitespace.

james-smith-za:
[test.pdf](https://github.com/ska-sa/katgpucbf/files/10518327/test.pdf)
This is the QTP that gets produced.

The hyperlinks don't work, which puzzles me. But other than that, it seems to work.

qualification/report/generate_pdf.py
```diff
@@ -876,17 +894,21 @@ def main():
     parser.add_argument("input", help="Output of pytest --report-log=...")
     parser.add_argument("pdf", help="PDF file to write")
     parser.add_argument("-c", "--commit-id", action="store_true", help="Output commit ID of katgpucbf image")
+    parser.add_argument("--generate-procedure-doc", action="store_false", dest="make_report", default=True)
     parser.add_argument(
-        "--report-doc-id",
+        "--doc-id",
         help="Document number to write to the qualification test report",
```
bmerry, line 900: Need to update the help

qualification/report/generate_pdf.py
```diff
@@ -876,17 +894,21 @@ def main():
...
         type=str,
-        default="E1200-0000-005",
+        default=DEFAULT_DOC_ID,
```
bmerry, line 902: I would rather leave out the default here, and then `if args.doc_id is None` determine a default.

qualification/report/generate_pdf.py
```diff
@@ -876,17 +894,21 @@ def main():
     parser.add_argument("input", help="Output of pytest --report-log=...")
     parser.add_argument("pdf", help="PDF file to write")
     parser.add_argument("-c", "--commit-id", action="store_true", help="Output commit ID of katgpucbf image")
+    parser.add_argument("--generate-procedure-doc", action="store_false", dest="make_report", default=True)
```
bmerry, line 897: What do you think about using `--doc=report` / `--doc=procedure` rather than a boolean flag (you could still have a default)? Sooner or later someone might want to generate something else. Similarly the argument to `document_from_list` could be a string ("report" or "procedure") rather than a boolean.

bmerry (changes requested):
This is certainly more a more compact and easy-to-follow change than the previous approach - glad I suggested it. The lack of hyperlinks is a concern; let me know if you'd like help debugging it.

I think the previous PR also updated the Jenkinsfile to generate the QTP. Are you still aiming to bring that back in this PR, or leaving it for a separate one?

qualification/report/generate_pdf.py
```diff
@@ -876,17 +894,21 @@ def main():
...
     parser.add_argument(
-        "--report-doc-id",
+        "--doc-id",
         help="Document number to write to the qualification test report",
         type=str,
-        default="E1200-0000-005",
+        default=DEFAULT_DOC_ID,
     )
     args = parser.parse_args()
+    if args.make_report is False and args.doc_id == DEFAULT_DOC_ID:
```
bmerry, line 905: `not foo` is the usual way of testing whether a boolean is false, except in cases where you want to be sure it is exactly the value False and not some falsey non-boolean.

qualification/report/generate_pdf.py
james-smith-za: That's a fair point. Done.

qualification/report/generate_pdf.py
james-smith-za: Done.

qualification/report/generate_pdf.py
james-smith-za: I've given this some thought, and I think on balance I'd let the YAGNI principle be the guiding one here.

I don't forsee (right now) another kind of document that we might want to generate with this specific code. Handling sets of valid strings is also more tricky than simple booleans.

If the need arises, it's straightforward enough to add the functionality, but I don't see a need right now.

qualification/report/generate_pdf.py
james-smith-za: I remember thinking about it, and I think my logic was, I wasn't expecting it to be anything other than `False`.

But I've handled the argument validation in a different way now so this line goes away.

james-smith-za:
> QTP itself looks good. We probably need to have further discussions around the detail included for the test procedures themselves.
 
Indeed. But that's beyond the scope of this PR I think.

> I also picked up that we're still listing Thomas Abbott as the MK PrgM. I propose the following revisions to the document signatories:
> 
>     * DSP Team as-is (although, you may want to correct your designation to DSP Engineer)
> 
>     * Approved by: Marcel Gouws, Senior System Analyst
> 
>     * Accepted by: Scarre Ceilers, MK+ System Engineer

So I noticed a bit of an inconsistency in that the author is listed as DSP team but it's my name on the list of signatories. Which kind of makes sense I suppose, as a "team" can't sign something, unless we want to individually list every person's name.
[test_230130.pdf](https://github.com/ska-sa/katgpucbf/files/10535762/test_230130.pdf)
Is this OK?

bmerry (changes requested):

qualification/report/generate_pdf.py
bmerry: This should probably have a help string too.

bmerry (changes requested):
I think the lack of hyperlinks is because for procedure doesn't explicitly use the Hyperref Python class. You can fix it by just adding `\usepackage{hyperref}` to preamble.tex.

tyronevb:
> > QTP itself looks good. We probably need to have further discussions around the detail included for the test procedures themselves.
> 
> Indeed. But that's beyond the scope of this PR I think.
> 
> > I also picked up that we're still listing Thomas Abbott as the MK PrgM. I propose the following revisions to the document signatories:
> > ```
> > * DSP Team as-is (although, you may want to correct your designation to DSP Engineer)
> > 
> > * Approved by: Marcel Gouws, Senior System Analyst
> > 
> > * Accepted by: Scarre Ceilers, MK+ System Engineer
> > ```
> 
> So I noticed a bit of an inconsistency in that the author is listed as DSP team but it's my name on the list of signatories. Which kind of makes sense I suppose, as a "team" can't sign something, unless we want to individually list every person's name. [test_230130.pdf](https://github.com/ska-sa/katgpucbf/files/10535762/test_230130.pdf) Is this OK?

Seems like there was some misunderstanding. I meant to leave the DSP Team member signatories as it was before:

- Submitted by: J. Smith, DSP Engineer
- Approved by: T. van Balla, Functional Manager: DSP
- Approved by: M. Gouws, ....
- Accepted by: S. Ceiliers, ...

qualification/report/preamble.tex
```diff
@@ -31,10 +31,9 @@
 
 %% Format: \addcontributor{Role}{Name}{Designation}{Affiliation}{Date}
 \newcommand{\docApproval}{
-  \addcontributor{Submitted by}{J.N. Smith}{Digital Engineer}{SARAO}{}
-  \addcontributor{Approved by}{T. van Balla}{Functional Manager: DSP}{SARAO}{}
+  \addcontributor{Submitted by}{J.N. Smith}{DSP Engineer}{SARAO}{}
+  \addcontributor{Approved by}{M. Gouws}{Senior System Analyst}{SARAO}{}
```
tyronevb, line 35:
```suggestion
  \addcontributor{Approved by}{T. van Balla}{Functional Manager: DSP}{SARAO}{}
  \addcontributor{Approved by}{M. Gouws}{Senior System Analyst}{SARAO}{}
```

tyronevb (changes requested):

amishatishpatel (commented):
Comments are mostly covered by others, just the small one from me specifically.

qualification/report/generate_pdf.py
```diff
@@ -764,65 +772,75 @@ def document_from_list(result_list: list, doc_id: str) -> Document:
...
+    if make_report:
+        _doc_test_configuration(doc, test_configuration)
 
-    with doc.create(Section("Result Summary")) as summary_section:
-        _doc_result_summary(summary_section, results)
+        with doc.create(Section("Result Summary")) as summary_section:
+            _doc_result_summary(summary_section, results)
 
-    with doc.create(Section("Detailed Test Results")) as section:
+    with doc.create(Section("Detailed Test Results" if make_report else "Test Procedures")) as section:
```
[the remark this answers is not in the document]
james-smith-za: Yeah, the "Detailed test results" have more "detail" in them than the plain old procedures.

Generally "step" tells what you're going to do, and "detail" reports on some kind of information that's happened as a result of taking that step.

james-smith-za:
> I think the lack of hyperlinks is because for procedure doesn't explicitly use the Hyperref Python class. You can fix it by just adding `\usepackage{hyperref}` to preamble.tex.

Good catch! Thanks!

james-smith-za:
> Seems like there was some misunderstanding. I meant to leave the DSP Team member signatories as it was before:
> 
>     * Submitted by: J. Smith, DSP Engineer
> 
>     * Approved by: T. van Balla, Functional Manager: DSP
> 
>     * Approved by: M. Gouws, ....
> 
>     * Accepted by: S. Ceiliers, ...

You're correct, I did misunderstand. Your correction included (now with fixed hyperlinks):
[test.pdf](https://github.com/ska-sa/katgpucbf/files/10569189/test.pdf)

qualification/report/generate_pdf.py
```diff
@@ -56,7 +56,7 @@
 from pylatex.package import Package
 from pylatex.utils import bold, escape_latex
 
-DEFAULT_DOC_ID = "E1200-0000-005-dev"  # We default to the report, not the procedure
+DEFAULT_REPORT_DOC_ID = "E1200-0000-005-dev"  # We default to the report, not the procedure
```
bmerry, line 59: The comment can be removed now I think.

bmerry (changes requested):
Just one niggle. If that's addressed I don't need another round of review.

PR #512, opened 2023-01-28
Bump versions in pre-commit configuration

bmerry:
This was needed because isort 5.10 no longer installs (see
https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co), but I just ran `pre-commit autoupdate` to update everything.

The .flake8 needed to be fixed because flake8 doesn't support inline comments, and the newer version was complaining more loudly about that.


amishatishpatel (approved):
> The .flake8 needed to be fixed because flake8 doesn't support inline comments [...]

Just the kind of low stakes, facepalming laugh I needed this morning.

PR #516, opened 2023-02-09
Use blue text instead of red borders for QTR hyperlinks

bmerry:
I find it a bit more comfortable to read.

### DIFF
--- a/qualification/report/generate_pdf.py
+++ b/qualification/report/generate_pdf.py
@@ -767,6 +767,7 @@ def document_from_list(result_list: list, doc_id: str, *, make_report=True) -> D
         document_options=["11pt", "english", "twoside"],
         inputenc=None,  # katdoc inputs inputenc with specific options, so prevent a clash
     )
+    doc.packages.add(Package("hyperref", "colorlinks,linkcolor=blue"))
     today = date.today()  # TODO: should store inside the JSON
     doc.set_variable("theAuthor", "DSP Team")
     doc.set_variable("docDate", today.strftime("%d %B %Y"))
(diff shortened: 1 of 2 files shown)


bmerry:
An example report: [test.pdf](https://github.com/ska-sa/katgpucbf/files/10697167/test.pdf)

bmerry:
@tyronevb we need a casting vote: I think this looks better, but James prefers the original of a coloured border around links.

bmerry:
@tyronevb ping

tyronevb (changes requested):
I'm a bit conflicted:

- the blue text is certainly easier on the eyes than the red borders (especially in cases where there are links on successive lines in a table)
- it isn't _as obvious_ that it's a hyperlink in some cases

Is there a way to make is a bit more obvious? One idea is to underline them as well (but not the links in the TOC), but I'm not how that'll look in the larger tables which contain lots of links.

james-smith-za:
> Is there a way to make is a bit more obvious? One idea is to underline them as well (but not the links in the TOC), but I'm not how that'll look in the larger tables which contain lots of links.

I'm not sure that there's a way to do it selectively unless you do it manually.
I think you can do links without formatting then manually format each link, but that would be too much of a pain.

bmerry:
It's easy enough to underline the links we add to the body. I've attached a sample showing what that would look like.
[report.pdf](https://github.com/ska-sa/katgpucbf/files/10883267/report.pdf)

james-smith-za:
I'd rather go with plain blue text than underlined. Since a lot of the links are in tables, the blue underline and the table cell boundaries don't look nice so close together.

If we want to get rid of the red boxes, it doesn't make sense to replace them with something that's almost equally unpleasant.

I'll withdraw my objection if you feel strongly about it.

tyronevb:
Thanks, @bmerry.

Yeah, the underlined text looks messy in the tables. Let's rather go with the plain blue text.

PR #517, opened 2023-02-09
Make qualification test flexible about what configs to test

bmerry:
- Allow a maximum number of antennas to be set (and test a range of interesting antenna counts up to the maximum)
- Allow the channelisations and bands to be chosen
- Add definitions for S-band (5 variants for the different sub-bands)

Checklist (if not applicable, edit to add `(N/A)` and mark as done):

- [x] (N/A) If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`
- [x] (N/A) If modules are added/removed: use sphinx-apidoc to update files in `doc/`
- [x] Ensure copyright notices are present and up-to-date
- [ ] If qualification tests are changed: attach a sample qualification report
- [x] (N/A) If design has changed: ensure documentation is up to date
- [x] (N/A) If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match

Closes NGC-695.

PR #519, opened 2023-02-13
Increase timeout for Jenkins qualification test

bmerry:
PR #517 made the qualification report include a lot more tests, so it's now exceeding the 1 hour time limit. Bump it up to 5.

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -59,7 +59,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 1, unit: 'HOURS')
+    timeout(time: 5, unit: 'HOURS')
   }
 
   stages {


james-smith-za (approved):
I wonder whether it's worthwhile trying to make the number a little bit more dynamic so that it adjusts the timeout based on the size of the set of work that it wants to do.

Though perhaps this won't change frequently enough that it warrants having that mechanism in place.

bmerry:
> I wonder whether it's worthwhile trying to make the number a little bit more dynamic so that it adjusts the timeout based on the size of the set of work that it wants to do.

I suspect you're overthinking it :-) I picked 5 hours since that means it should finish by the time people are generally starting work. If it exceeds that, we're probably want to do more than just increase it further e.g. we might need to move it to weekends only.

PR #522, opened 2023-02-15
Use cross-correlations to better measure channel shape

bmerry:
The channel shape was being limited by the dither/quantisation noise in the dsim. Because auto-correlations were being used, this noise was self-correlated. Using cross-correlations, with the same tone but independent noise, means the noise is uncorrelated and so tends to wash out.

The implementation uses the two pols of each dsim as the two sources to correlate. This simplifies the code somewhat, but halves the efficiency as it now requires two inputs to do what previous was done with one. To compensate, I've bumped DEFAULT_ANTENNAS up to 8.


bmerry:
Sample report (8K channels only): [report-cross3.pdf](https://github.com/ska-sa/katgpucbf/files/10744428/report-cross3.pdf)

james-smith-za (approved):

PR #523, opened 2023-02-17
Bump ipython to the latest version

bmerry:
Dependabot was complaining about the old version. It's not used programmatically (just a convenience for requirements-dev.txt to install it), so I think this is a zero-risk change.


amishatishpatel (approved):
> I think this is a zero-risk change.

Tempting fate with statements like these 👀

PR #527, opened 2023-02-27
Fix -c option to generate_pdf

bmerry:
It causes `parse` to be called twice. pytest.Report._from_json actually modifies the dictionary it was given, and on the second time round that causes the parsing to fail because it's already modified. Fix it by giving it a deep copy of the original.

Closes NGC-875.


james-smith-za (approved):
Looks fine.

I've been thinking of getting this functionality at an earlier, more robust stage in the pipeline though, because if something fails then we don't actually see the commit information.

PR #528, opened 2023-02-27
Fix test_delay_application_time

bmerry:
The test sets a delay model for 200ms in the future, then tries to grab the dump that brackets that change. The *start* of that dump may be in the past, so (implicitly) using steady_state_timestamp means we might just throw away the dump we actually want.

I'm not sure why we've only run into this with UHF and not with other bands.

Partly addresses NGC-897.

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -266,7 +266,7 @@ class BaselineCorrelationProductsReceiver:
         Parameters
         ----------
         min_timestamp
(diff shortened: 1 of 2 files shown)


bmerry:
Sample report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/10839025/report.pdf)

amishatishpatel (approved):

PR #530, opened 2023-02-28
Mark phase rate test as expected to fail

bmerry:
The test is flawed, but rewriting it will require CBF-REQ-0128 to be clarified. For now mark it as xfail, and update the report generator to better handle xfail.

Closes NGC-897 (for now).

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -532,6 +532,7 @@ async def test_delay_phase(
 
 
 @pytest.mark.requirements("CBF-REQ-0128,CBF-REQ-0112")
+@pytest.mark.xfail(reason="requirements unclear")
 async def test_phase_rate(
     correlator: CorrelatorRemoteControl,
     receive_baseline_correlation_products: BaselineCorrelationProductsReceiver,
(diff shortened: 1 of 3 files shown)


bmerry:
Report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/10850165/report.pdf)

qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -532,6 +532,7 @@ async def test_delay_phase(
 
 
 @pytest.mark.requirements("CBF-REQ-0128,CBF-REQ-0112")
+@pytest.mark.xfail(reason="requirements unclear")
```
james-smith-za, line 535: My only concern is that, as far as I understand things, this marks all the `phase_rate` tests as `xfail`.

My reading of the [pytest documentation ](https://docs.pytest.org/en/6.2.x/skipping.html) suggests that something which unexpectedly passes should be marked as `xpass`, but that doesn't seem to be the case with the L-band phase-rate test in your report.

All this looks fine to me, I don't have any objections, but I'm struggling to understand why this is happening.

bmerry: It's reported as xpass in the summary on stdout, but the `outcome` field in the JSON structure is still `passed`. There is probably some way to distinguish an xpass from a normal pass, but I didn't see much need to distinguish them.

james-smith-za: I see. So there's a bit of a disconnect somewhere.

This is fine for now I think.

james-smith-za (approved):

PR #531, opened 2023-03-02
Make report generation more robust to tests that fail early

bmerry:
If a test fails early enough, we don't get the `start_time` information. Instead of failing the entire report generation, just don't try to report it.

It's not the prettiest thing, because we also don't get the test name or parameters, but at least it lets us see results for any tests that did run properly, while also being aware that there was a failure and seeing the backtrace/exception.


james-smith-za (approved):

bmerry:
The Jenkins failure is because Harbor is down, so I'm merging anyway.

PR #535, opened 2023-03-17
fgpu: add command-line arguments for narrowband

bmerry:
This adds a `--narrowband` command-line argument, which takes properties to set via comma-separated key=value pairs. I chose this approach rather than having lots of separate narrowband arguments (`--narrowband-channels`, `--narrowband-taps` etc) because this notation allows for multiple narrowband products to be specified (the original MK requirements called for 5 independently steerable narrowband streams, for example).

I've also put in some abstraction in Engine to take a list of outputs to produce, which should reduce the churn in the Engine constructor interface in future, although at the moment it only supports a single wideband output.

The list of parameters that are output-specific may grow in future; this is mostly about establishing the design.

- [x] If modules are added/removed: use sphinx-apidoc to update files in `doc/`

Closes NGC-567.

### DIFF
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -80,6 +80,8 @@ def comma_split(
 
     def func(value: str) -> list[_T]:  # noqa: D102
         parts = value.split(",")
+        if parts == [""]:
+            parts = []
         n = len(parts)
         if count is not None and n == 1 and allow_single:
             parts = parts * count
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -336,8 +336,8 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         if "taps" not in output:
             output["taps"] = args.taps
         if "w_cutoff" not in output:
-            output["taps"] = args.w_cutoff
-    args.narrowband = [NarrowbandOutput(*output) for output in args.narrowband]
+            output["w_cutoff"] = args.w_cutoff
+    args.narrowband = [NarrowbandOutput(**output) for output in args.narrowband]
 
     return args
 
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -25,7 +25,7 @@ import argparse
 import asyncio
 import logging
 from collections.abc import Callable, Sequence
-from typing import TypeVar
+from typing import TypedDict, TypeVar
 
 import katsdpservices
 import katsdpsigproc.accel as accel
@@ -33,7 +33,7 @@ import prometheus_async
 from katsdpservices import get_interface_address
 from katsdpservices.aiomonitor import add_aiomonitor_arguments, start_aiomonitor
 from katsdpsigproc.abc import AbstractContext
-from katsdptelstate.endpoint import endpoint_list_parser
+from katsdptelstate.endpoint import Endpoint, endpoint_list_parser
 
 from .. import (
     DEFAULT_KATCP_HOST,
@@ -45,9 +45,11 @@ from .. import (
     __version__,
 )
 from ..monitor import FileMonitor, Monitor, NullMonitor
+from ..spead import DEFAULT_PORT
 from ..utils import add_signal_handlers, parse_source
 from . import DIG_SAMPLE_BITS_VALID
 from .engine import Engine
+from .output import NarrowbandOutput, WidebandOutput
 
 _T = TypeVar("_T")
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


src/katgpucbf/fgpu/main.py
```diff
@@ -266,6 +330,15 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
             parser.error("Live source requires --src-interface")
     if len(args.dst) % len(args.dst_interface) != 0:
         parser.error("number of destinations must be divisible by number of destination interfaces")
+
+    # Convert narrowband from NarrowbandOutputDict to NarrowbandOutput
+    for output in args.narrowband:
+        if "taps" not in output:
+            output["taps"] = args.taps
+        if "w_cutoff" not in output:
+            output["taps"] = args.w_cutoff
```
amishatishpatel, line 339: Methinks this should be `output["w_cutoff"]` - and not `output["taps"]` again.

src/katgpucbf/fgpu/main.py
```diff
@@ -266,6 +330,15 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
...
+    args.narrowband = [NarrowbandOutput(*output) for output in args.narrowband]
```
amishatishpatel, line 340: After originally typing this comment up, perhaps it's best if we just chat for a few minutes for you to explain the logic to me (please).

It's likely
* my ipython hackery to understand this, 
* coupled with my ~incomplete understanding of what's meant to happen,
* and/or what it means to iterate over a `TypedDict`,

are all contributing to my general confusion about this line.

In a nutshell, 
* I'm not sure if this is meant to yield a single `NarrowbandOutput` instance, and
* I'm thrown by it seeming like it creates a separate instance for each `output` it expands/flattens.

amishatishpatel (changes requested):
Firstly, [TIL what churn is](https://linearb.io/blog/what-is-code-churn/) - always appreciate the side-quest learnings. Certainly appreciate the reasoning for `output.py` and general infrastructure around it.

I've requested one explicit change, just a small one in `main.py`. The other is a more general request for clarification.

I'm pretty sure I'm on-board with `test_main.py`, but just want to briefly go over the final conversion from `NarrowbandOutputDict` to `NarrowbandOutput` before approving.

test/fgpu/test_main.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+    def test_missing_key(self, missing: str, value: str) -> None:
+        """Test without one of the required keys."""
+        with pytest.raises(ValueError, match=f"--narrowband: {missing} is missing"):
+            parse_narrowband(value)
+
+    def test_duplicate_key(self) -> None:
+        """Test with a key specified twice."""
+        with pytest.raises(ValueError, match="--narrowband: channels specified twice"):
+            parse_narrowband("channels=8,channels=9,decimation=8,dst=239.1.2.3+1:7148")
```
james-smith-za, line 95: Only suggestion I'd make, not that it's particularly important, would be to not have the repeated parameter right next to itself, because that would be more likely what a user would do (IMO).

test/fgpu/test_main.py
```diff
@@ -0,0 +1,127 @@
...
+    def test_narrowband_defaults(self) -> None:
+        """Test that missing narrowband config is taken from the global config."""
+        raw_args = [
+            "--src-interface=lo",
+            "--dst-interface=lo",
+            "--adc-sample-rate=1712000000.0",
+            "--channels=1024",
+            "--sync-epoch=0",
+            "--taps=64",
+            "--w-cutoff=0.9",
```
james-smith-za, line 110: The Engine __init__() method refactor was pretty good. I almost want to suggest we add a 
```
--wideband=...
```
command-line option as well, because it looks a bit more consistent.

BUT

that would break backwards compatibility so I'm not sure whether it would be worth the effort.

james-smith-za (approved):

src/katgpucbf/fgpu/main.py
```diff
@@ -88,6 +92,55 @@ def func(value: str) -> list[_T]:  # noqa: D102
...
+                    case "dst":
+                        kws[key] = endpoint_list_parser(DEFAULT_PORT)(data)
+                    case _:
+                        raise ValueError(f"--narrowband: unknown key {key}")
+            case _:
+                raise ValueError(f"--narrowband: missing '=' in {part}")
+    for key in ["channels", "decimation", "dst"]:
+        if key not in kws:
+            raise ValueError(f"--narrowband: {key} is missing")
+    return kws
```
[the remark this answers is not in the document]
bmerry: I thought you'd appreciate it :smile:

test/fgpu/test_main.py
james-smith-za: This is a fair point. Please ignore.

test/fgpu/test_main.py
bmerry: It's definitely something I'm considering adding (in a future PR), particularly since I said I wanted to try adding support for multiple widebands (to flush out all the issues with having multiple streams) before trying to support narrowband.

Backwards compatibility can be finessed e.g. fall back to the current behaviour if zero `--wideband` options are given.

PR #537, opened 2023-03-24
Refactor fgpu to support multiple output streams

bmerry:
The result of this refactoring is that fgpu should be pretty close to working with multiple output streams, with all the major refactoring done, but it is not actually supported or tested. This is a stepping stone that should still be backwards compatible, before changing the interface with katsdpcontroller. Getting it onto main now should reduce future merge conflicts.

The main change is that the bulk of `run_processing` and `run_transmit` and their related data structures are moved into a separate class called `Pipeline`. There is a single `Engine` which potentially manages multiple pipelines. Each pipeline has its own compute and download command queues. There is still only a single spead2 transmit stream: this makes the packet pacing global for the engine, and avoids the need for extra transmission threads.  This uber-stream has a set of substreams per output, and each `run_transmit` sends data to the appropriate subsets (this bit may still be a bit buggy and will need testing).

Copying the prefix of each chunk to the tail of the previous chunk used to happen as part of `run_processing`, but is now moved to the receive code, since it only needs to be done once for all Pipelines. I think this actually simplifies the processing loop a lot and I wish I'd done it to start with. I introduced a new command queue for this (`_copy_queue`) since there isn't an Engine-wide compute queue any more.

InItems have a `refcount` field to keep track of how many Pipelines are still using them, so that they can be returned to the free queue only when none of the Pipelines need them any more.

One slight worry I have with the design is that there is no mechanism enforcing fairness between the pipelines. If one pipeline ends up getting its GPU work prioritised, it could end up several chunks ahead of another, and that would lock up a bunch of InItems and stall the receiver. But I think that's a bridge we can burn when we get to it.

Closes NGC-573.

### DIFF
--- a/doc/engines.rst
+++ b/doc/engines.rst
@@ -81,10 +81,11 @@ The general operation of the DSP engines is illustrated in the diagram below:
 
 .. tikz:: Data Flow. Double-headed arrows represent data passed through a
    queue and returned via a free queue.
-   :libs: chains
+   :libs: chains, fit
 
    \tikzset{proc/.style={draw, rounded corners, minimum width=4.5cm, minimum height=1cm},
-            pproc/.style={proc, minimum width=2cm},
+            pproc-base/.style={minimum width=2cm, minimum height=1cm},
+            pproc/.style={proc, pproc-base},
             flow/.style={->, >=latex, thick},
             queue/.style={flow, <->},
             fqueue/.style={queue, color=blue}}
@@ -109,7 +110,7 @@ The general operation of the DSP engines is illustrated in the diagram below:
      \node[pproc, draw=none, anchor=east,
            start chain=tx\s-1 going below, on chain=tx\s-1] (transmit\s-1) at (transmit\s.east) {};
      \foreach \i in {0, 1} {
-       \node[pproc, on chain=tx\s-\i] (outstream\s-\i) {Stream};
+       \node[pproc-base, on chain=tx\s-\i] (outstream\s-\i) {};
        \draw[flow] (transmit\s-\i) -- (outstream\s-\i);
      }
      \draw[queue] (align) -- (process\s);
@@ -117,6 +118,8 @@ The general operation of the DSP engines is illustrated in the diagram below:
      \draw[queue] (download\s) -- (transmit\s);
      \end{scope}
    }
+   \node[proc, fit=(outstream0-0) (outstream1-1), inner sep=0pt, outer sep=0pt] (outstream) {};
+   \node at (outstream.center) {Stream};
    \foreach \i in {0, 1} {
      \node[pproc, on chain=rx\i] (receive\i) {Receive};
      \node[pproc, on chain=rx\i] (stream\i) {Stream};
@@ -127,6 +130,13 @@ The general operation of the DSP engines is illustrated in the diagram below:
 
 The F-engine uses two input streams and aligns two incoming polarisations, but
 in the XB-engine there is only one.
+
+There might not always be multiple processing pipelines. When they exist, they
+are to support multiple outputs generated from the same input, such as wide-
+and narrow-band F-engines, or multiple beams. A single stream is used so that
+all the outputs go through a single thread (and hence only one core is needed
+for sending) and a single rate-limiter (preventing micro-bursts if each
+pipeline sends data at the same time).
 
 Chunking
 ^^^^^^^^
--- a/doc/engines.rst
+++ b/doc/engines.rst
@@ -88,31 +88,42 @@ The general operation of the DSP engines is illustrated in the diagram below:
             flow/.style={->, >=latex, thick},
             queue/.style={flow, <->},
             fqueue/.style={queue, color=blue}}
-   \node[proc, start chain=going below, on chain] (align) {Align, copy to GPU};
+   \begin{scope}[start chain=chain going below]
+   \node[proc, on chain] (align) {Align, copy to GPU};
    \node[pproc, draw=none, anchor=west,
          start chain=rx0 going above, on chain=rx0] (align0) at (align.west) {};
    \node[pproc, draw=none, anchor=east,
          start chain=rx1 going above, on chain=rx1] (align1) at (align.east) {};
-   \node[proc, on chain] (process) {GPU processing};
-   \node[proc, on chain] (download) {Copy from GPU};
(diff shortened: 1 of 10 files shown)


src/katgpucbf/send.py
```diff
@@ -44,9 +45,9 @@ class DescriptorSender:
     first_interval
         Delay (in seconds) immediately after starting. If not specified, it
         defaults to `interval`.
-    all_substreams
-        Send descriptors to all substreams if true, otherwise default behaviour
-        is to send only to the first substream.
+    substreams
+        Substream indexes to which descriptors are sent. If not specified,
```
james-smith-za, line 49:
```suggestion
        Substream indices to which descriptors are sent. If not specified,
```
Just for bike-shedding purposes, ignore at your preference.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1312,126 +875,607 @@ def set_gains(self, input: int, gains: np.ndarray) -> None:
...
+        dst_packet_payload: int,
+        dst_affinity: int,
+        dst_comp_vector: int,
+        outputs: list[Output],
+        adc_sample_rate: float,
+        send_rate_factor: float,
+        feng_id: int,
+        num_ants: int,
+        chunk_samples: int,
+        chunk_jones: int,
```
james-smith-za, line 1008: This param is missing from the docstring.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1312,126 +875,607 @@ def set_gains(self, input: int, gains: np.ndarray) -> None:
...
+        in the output heaps so that the X-engine can determine where the data
+        fits in.
+    num_ants
+        The number of antennas in the array. Used for numbering heaps so as
+        not to collide with other antennas transmitting to the same X-engine.
+    chunk_samples
+        Number of samples in each input chunk, excluding padding samples.
+    spectra
+        Number of spectra that will be produced from a chunk of incoming
+        digitiser data.
```
james-smith-za, line 957: This param isn't actually  in the argument list.

james-smith-za (commented):
On the whole it looks good to me.

A few nitpicks.

I haven't had time to interrogate the logic to check whether there would be any race conditions or anything, I suspect that would be easier to verify while actually running with a bunch of different parameters, and once you actually have NB output streams to test (am I correct in understanding that there's no mechanism currently to actually fire up multiple wideband outputs?)

doc/engines.rst
```diff
@@ -88,31 +88,42 @@ The general operation of the DSP engines is illustrated in the diagram below:
...
+   \foreach \s in {0, 1} {
+     \begin{scope}[continue chain=chain/stream\s]
+     \node[proc, on chain] (download\s) {Copy from GPU};
+     \node[proc, on chain] (transmit\s) {Transmit};
+     \node[pproc, draw=none, anchor=west,
+           start chain=tx\s-0 going below, on chain=tx\s-0] (transmit\s-0) at (transmit\s.west) {};
+     \node[pproc, draw=none, anchor=east,
+           start chain=tx\s-1 going below, on chain=tx\s-1] (transmit\s-1) at (transmit\s.east) {};
+     \foreach \i in {0, 1} {
+       \node[pproc, on chain=tx\s-\i] (outstream\s-\i) {Stream};
```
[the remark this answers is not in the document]
bmerry: I'll take a look - if not the diagram, it should probably be in the text somewhere.

bmerry:
@james-smith-za I think I've addressed all your comments.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1312,41 +873,516 @@ def set_gains(self, input: int, gains: np.ndarray) -> None:
...
+        ID of the F-engine indicating which one in the array this is. Included
+        in the output heaps so that the X-engine can determine where the data
+        fits in.
+    num_ants
+        The number of antennas in the array. Used for numbering heaps so as
+        not to collide with other antennas transmitting to the same X-engine.
+    chunk_samples
+        Number of samples in each input chunk, excluding padding samples.
+    chunk_jones
+        Number of Jones vectorsin each output chunk.
```
james-smith-za, line 954:
```suggestion
        Number of Jones vectors in each output chunk.
```

james-smith-za (approved):
One suggestion to fix a typo, other than that I am happy to merge.

amishatishpatel (approved):
I believe the saying goes
> Lots of moving parts.

Still, I think I'm with you most of the way. There is a faint air of circular referencing about the way the `{In, Out}Items` are trafficked between the processing loops, but I suppose that's necessary (and obviously cleanly done, by the looks of it).

Not really too many comments from my side (that haven't been addressed by James already). One overall comment Re: your blurb for this PR - not necessarily important either.
> Each pipeline has its own compute and upload command queues. 

I'm not sure if you're referring to queues by name, by convention (direction of operation), or interchanging `command_queues` and `asyncio.Queues` (i.e. Item queues). 

In terms of `command_queues`:
* `Engine` has an `_upload_queue` (alongside its new `_copy_queue`)
* `Pipeline` has a `compute_queue` and a `_download_queue`.

In terms of `asyncio.Queues`:
* `Engine` just has the one `_in_free_queue`,
* `Pipeline` has five across `in/in_free`, `out/out_free` and `send_free`.

src/katgpucbf/fgpu/main.py
```diff
@@ -125,6 +127,8 @@ def parse_narrowband(value: str) -> NarrowbandOutputDict:
                 match key:
                     case _ if key in kws:
                         raise ValueError(f"--narrowband: {key} specified twice")
+                    case "name":
+                        kws[key] = data
```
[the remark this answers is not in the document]
james-smith-za: No, there won't be spaces here, so I don't think it's necessary to do that kind of validation.

If something goes terribly wrong rather just let it crash IMO.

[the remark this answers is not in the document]
bmerry: There is nothing stopping you including almost any character (including spaces) when invoking fgpu from the command line - about the only thing you couldn't include would be a comma, since that splits up the keys-value pairs. At the moment I think it won't really matter. Once the name starts getting included in katcp sensors and the like, I like the idea of adding a check on the name to ensure that the sensor names are valid.

The product controller restricts stream names though, so this check would just be to protect people starting fgpu manually.

PR #542, opened 2023-03-29
fgpu: make some katcp sensors pipeline-specific

bmerry:
Prefix the names of eq, delay, feng-clip-cnt and dig-pwr-dbfs sensors with the output name. The last of these is done for expediency: it ought to be roughly the same for all outputs, but it's calculated per-pipeline, and probably not on the identical set of inputs and at the identical cadence due to the interaction with PFB windows.

See NGC-569.

- [ ] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match (PR coming soon)

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -419,6 +419,7 @@ class Pipeline:
         # Initialize delays and gains
         self.delay_models: list[MultiDelayModel] = []
         self.gains = np.zeros((output.channels, self.pols), np.complex64)
+        self._populate_sensors()
         self._init_delay_gain()
 
         self.descriptor_heap = send.make_descriptor_heap(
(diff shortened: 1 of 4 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -419,19 +419,68 @@ def __init__(self, output: Output, engine: "Engine", context: AbstractContext, s
...
+            )
+
     def _init_delay_gain(self) -> None:
         """Initialise the delays and gains."""
-        # TODO[nb]: create per-pipeline sensors
         for pol in range(N_POLS):
             delay_model = MultiDelayModel(
-                callback_func=partial(self.update_delay_sensor, delay_sensor=self.engine.sensors[f"input{pol}.delay"])
+                callback_func=partial(
+                    self.update_delay_sensor, delay_sensor=self.engine.sensors[f"{self.output.name}.input{pol}.delay"]
```
james-smith-za, line 482: I don't really like this in-between thing that `black` does, to me it makes more sense to either have the entire function call on one line, or each argument on its own line.

No change needed, just felt the need to complain about this a bit.

james-smith-za (approved):
Looks fine to me.

I didn't check *too* carefully that all usage of the sensors was updated, but that will come out quickly enough in the wash if qualification starts failing.

src/katgpucbf/fgpu/engine.py
bmerry: All the line-wrapping in this function is a symptom that it's trying to stuff too much into a single statement. See if you like the new version better.

james-smith-za (approved):
My sense of feng-shui approves of this change.

amishatishpatel (approved):

PR #543, opened 2023-03-30
Rework fgpu/Engine's unhandled async tasks

amishatishpatel:
Two points to consider when reviewing:
1. As you can see, I went for the **literal** option in renaming `Pipeline::_chunk_finished` - happy to change as requested.
2. I decided not to roll the _send and cleanup_ functionality/chain of events into `Chunk.send` because it needed the `logger` to log the unexpected `Exception` - and the params list was busy enough anyway.

- [x] If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml` (N/A)
- [x] If modules are added/removed: use sphinx-apidoc to update files in `doc/` (N/A)
- [x] If qualification tests are changed: attach a sample qualification report (N/A)
- [x] If design has changed: ensure documentation is up to date (N/A)
- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match (N/A)

Closes NGC-899.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -724,10 +724,22 @@ class Pipeline:
     async def _chunk_send_and_cleanup(
         self, streams: list["spead2.send.asyncio.AsyncStream"], n_frames: int, chunk: send.Chunk
     ) -> None:
-        """Return a chunk to the free queue after it has completed transmission."""
+        """Transmit a chunk's data and return it to the free queue.
+
+        The returning of the chunk to the free queue happens whether the data
+        transmission was successful or not.
+
+        Parameters
+        ----------
+        streams
+            The streams transmitting data.
+        n_frames
+            Number of frames of data to be transmitted.
+        chunk
+            :class:`~send.Chunk` used to facilitate data transmission.
+        """
         try:
-            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
-            await task
+            await chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors)
         except asyncio.CancelledError:
             pass
         except Exception:
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -721,20 +721,21 @@ class Pipeline:
         logger.debug("run_processing completed")
         self._out_queue.put_nowait(None)
 
-    def _chunk_finished(self, chunk: send.Chunk, future: asyncio.Future) -> None:
-        """Return a chunk to the free queue after it has completed transmission.
-
-        This is intended to be used as a callback on an :class:`asyncio.Future`.
-        """
-        if chunk.cleanup is not None:
-            chunk.cleanup()
-            chunk.cleanup = None  # Potentially helps break reference cycles
+    async def _chunk_send_and_cleanup(
+        self, streams: list["spead2.send.asyncio.AsyncStream"], n_frames: int, chunk: send.Chunk
+    ) -> None:
+        """Return a chunk to the free queue after it has completed transmission."""
         try:
-            future.result()  # No result, but want the exception
+            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
+            await task
         except asyncio.CancelledError:
             pass
         except Exception:
             logger.exception("Error sending chunk")
+        finally:
+            if chunk.cleanup is not None:
+                chunk.cleanup()
+                chunk.cleanup = None  # Potentially helps break reference cycles
 
     async def run_transmit(self, streams: list["spead2.send.asyncio.AsyncStream"]) -> None:
         """Get the processed data from the GPU to the Network.
@@ -818,8 +819,8 @@ class Pipeline:
                 # We're not in PeerDirect mode
                 # (when we are the cleanup callback returns the item)
                 self._out_free_queue.put_nowait(out_item)
-            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
-            task.add_done_callback(partial(self._chunk_finished, chunk))
+            task = asyncio.create_task(self._chunk_send_and_cleanup(streams, n_frames, chunk))
+            self.engine.add_service_task(task)
 
         if task:
             try:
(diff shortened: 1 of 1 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -721,20 +721,21 @@ async def run_processing(self) -> None:
...
-
-        This is intended to be used as a callback on an :class:`asyncio.Future`.
-        """
-        if chunk.cleanup is not None:
-            chunk.cleanup()
-            chunk.cleanup = None  # Potentially helps break reference cycles
+    async def _chunk_send_and_cleanup(
+        self, streams: list["spead2.send.asyncio.AsyncStream"], n_frames: int, chunk: send.Chunk
+    ) -> None:
+        """Return a chunk to the free queue after it has completed transmission."""
```
bmerry, line 727: Update the docstring.

src/katgpucbf/fgpu/engine.py
```diff
@@ -721,20 +721,21 @@ async def run_processing(self) -> None:
...
         try:
-            future.result()  # No result, but want the exception
+            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
+            await task
```
bmerry, line 730:
```suggestion
            await chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors)
```
No need to create an asynchronous task which you immediately synchronise with.

amishatishpatel:
> Unrelated to the content of the review, when you put spaces around the x's in the checklist it renders like it does now, but if you remove the spaces it'll render as an actual checkbox (and you'll also get the bit that says 6/6 tasks done). You can also create the review without filling in the x's and then just click on the checkboxes.

Colour me doubly embarrassed. Please and thank you.

src/katgpucbf/fgpu/engine.py
```diff
@@ -818,8 +819,8 @@ async def run_transmit(self, streams: list["spead2.send.asyncio.AsyncStream"]) -
                 # We're not in PeerDirect mode
                 # (when we are the cleanup callback returns the item)
                 self._out_free_queue.put_nowait(out_item)
-            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
-            task.add_done_callback(partial(self._chunk_finished, chunk))
+            task = asyncio.create_task(self._chunk_send_and_cleanup(streams, n_frames, chunk))
+            self.engine.add_service_task(task)
 
         if task:
             try:
```
bmerry, line 826: On line 829 there is a reference in a comment to chunk_finished, which needs to be updated.

src/katgpucbf/fgpu/engine.py
```diff
@@ -818,8 +819,8 @@ async def run_transmit(self, streams: list["spead2.send.asyncio.AsyncStream"]) -
                 # We're not in PeerDirect mode
                 # (when we are the cleanup callback returns the item)
                 self._out_free_queue.put_nowait(out_item)
-            task = asyncio.create_task(chunk.send(streams, n_frames, self.engine.time_converter, self.engine.sensors))
-            task.add_done_callback(partial(self._chunk_finished, chunk))
+            task = asyncio.create_task(self._chunk_send_and_cleanup(streams, n_frames, chunk))
```
bmerry, line 822: Not directly related to this PR, but it might be nice to give the task a name.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1225,7 +1226,8 @@ def free_in_item(self, item: InItem) -> None:
                     assert pol_data.chunk is not None
                     chunks.append(pol_data.chunk)
                     pol_data.chunk = None
-                asyncio.create_task(self._push_recv_chunks(chunks, item.events))
+                task = asyncio.create_task(self._push_recv_chunks(chunks, item.events))
```
bmerry, line 1229: Not really related to this PR, but it might be nice to give this task a name.

bmerry (changes requested):
Looks good overall - just some nitpicks.

src/katgpucbf/fgpu/engine.py
james-smith-za: If you do 
```python
await coroutine()
```
then it won't get garbage-collected, because the interpreter is waiting at this point for the coroutine to finish before it continues execution of this function.

If you want to start a coroutine and carry on with some other stuff while you wait for it to finish, then something like this:
```python
my_task = create_task(coroutine())
foo = 1 + 1
bar = 2 + 2
# Now all the stuff we wanted to do is finished, we just need to wait for the coroutine
await my_task # If the task is already finished, this will just be nothing, but if we need to wait for it, execution will pause here
return foo + bar
```
this is fine, because the task has a handle and won't get garbage collected.

Now the problem comes in, say you have a task that needs to get done in the background, you don't really care when, but it just must not get forgotten.


```python
create_task(coroutine())
foo = 1 + 1
bar = 2 + 2
return foo + bar
```
now the task doesn't have a handle, the function returns, etc. so the interpreter doesn't have a way to know whether or not it still needs to be there, so is in danger of being garbage-collected. This is the kind of thing that we want to avoid. There seems to be (based on our experience) a low probability of this happening, but when you do it enough times, it's certainly going to come up now and again and the origin of the ticket is that it's going to be difficult to debug this, so rather just program defensively. So we avoid this by using the `add_service_task(my_task)` mechanism which maintains a reference to the task and prevents it being garbage collected.

If that makes sense.

bmerry: Thanks @james-smith-za, that's a good explanation.

A short rule of thumb is that any task or coroutine that you `await` is safe from this issue, because the fact that you're able to refer to it to `await` it means you still have a reference to it.

bmerry (approved):
Looks good. One bit of bikeshedding; I don't feel like I need to review it again if you make the change.

src/katgpucbf/fgpu/engine.py
```diff
@@ -818,14 +831,17 @@ async def run_transmit(self, streams: list["spead2.send.asyncio.AsyncStream"]) -
...
+                name="Chunk Send and Cleanup Task",
+            )
+            self.engine.add_service_task(task)
 
         if task:
             try:
                 await task
             except Exception:
-                pass  # It's already logged by the chunk_finished callback
+                pass  # It's already logged by chunk_send_and_cleanup
```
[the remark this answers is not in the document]
amishatishpatel: Damn and blast, I went on the convention used previously (private method commented without leading underscore). Thought about using it in a more factually correct way, will do so now.

PR #545, opened 2023-04-03
Bump redis to 4.5.4

bmerry:
To keep dependabot happy - we're not actually using it.


bmerry:
The build failures are due to the build machine running out of disk space.

james-smith-za (approved):

PR #549, opened 2023-04-04
Rename "decimation" to "subsampling" in DDC kernel

bmerry:
The term "decimation" is used elsewhere in CBF to mean the factor by which bandwidth is reduced, rather than the factor by which the sample rate is reduced. The mixing+filtering removes the symmetry properties of the original real-valued signal, so they differ by a factor of 2.

Closes NGC-925.

### DIFF
--- a/src/katgpucbf/fgpu/ddc.py
+++ b/src/katgpucbf/fgpu/ddc.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2022, National Research Foundation (SARAO)
+# Copyright (c) 2022-2023, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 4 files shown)


src/katgpucbf/fgpu/ddc.py
```diff
@@ -44,42 +44,44 @@ class DDCTemplate:
         The GPU context that we'll operate in
     taps
         Number of taps in the FIR filter
-    decimation
+    subsampling
         Fraction of samples to retain after filtering
```
amishatishpatel, line 48: I know you've updated fgpu.design, but worth asking if the docstring needs an update? As per your description on the ticket, subsampling is
> the factor by which the time between samples is increased by discarding samples [...]

Is this still implied by the existing docstring/parameter explainer?

amishatishpatel (commented):
One comment within before explicit approval. Adding to that, are there any logistics/minor plumbing required to ensure `subsampling == 2*decimation`?

src/katgpucbf/fgpu/ddc.py
bmerry: It's the same thing. If there was a sample every 1ns before, and we retain 1/8 of the samples, then (assuming that means we retain every 8th sample) there will be a sample every 8ns afterwards.

amishatishpatel (approved):

james-smith-za (approved):

PR #550, opened 2023-04-04
Add complex version of PFB filter

bmerry:
With narrowband, the PFB will take its input from the DDC filter output i.e. complex64 rather than packed real values. Add support for that mode of operation.

The complex version also does not emit digitiser total power, since it is not the first step in the pipeline.

There are probably going to need to be some updates to the design documentation, but I'll rather do a separate update to bring that up to date once I've got everything implemented.

Closes NGC-925.

### DIFF
--- a/src/katgpucbf/fgpu/compute.py
+++ b/src/katgpucbf/fgpu/compute.py
@@ -116,7 +116,7 @@ class Compute(accel.OperationSequence):
         spectra: int,
         spectra_per_heap: int,
     ) -> None:
-        self.dig_sample_bits = template.pfb_fir.dig_sample_bits
+        self.dig_sample_bits = template.pfb_fir.input_sample_bits
         self.template = template
         self.samples = samples
         self.spectra = spectra
--- a/src/katgpucbf/fgpu/pfb.py
+++ b/src/katgpucbf/fgpu/pfb.py
@@ -27,6 +27,7 @@ from katsdpsigproc import accel
 from katsdpsigproc.abc import AbstractCommandQueue, AbstractContext
 
 from .. import BYTE_BITS
+from . import DIG_SAMPLE_BITS_VALID
 
 
 class PFBFIRTemplate:
@@ -40,25 +41,41 @@ class PFBFIRTemplate:
         The number of taps that you want the resulting PFB-FIRs to have.
     channels
         Number of channels into which the input data will be decomposed.
-    dig_sample_bits
-        Bits per digitiser sample.
+    input_sample_bits
+        Bits per input sample. If `complex_input` is true, the input values
+        are floating-point complex numbers and this is the number of bits in
+        the complex value (and must equal 64). Otherwise, the inputs are
+        packed integers, and the value must be in
+        :data:`DIG_SAMPLE_BITS_VALID`.
     unzip_factor
         The output is reordered so that every unzip_factor'ith pair of
         outputs is placed contiguously.
     """
 
     def __init__(
-        self, context: AbstractContext, taps: int, channels: int, dig_sample_bits: int, unzip_factor: int = 1
+        self,
+        context: AbstractContext,
+        taps: int,
+        channels: int,
+        input_sample_bits: int,
+        unzip_factor: int = 1,
+        *,
+        complex_input: bool = False,
     ) -> None:
         if taps <= 0:
             raise ValueError("taps must be at least 1")
         self.wgs = 128
         self.taps = taps
         self.channels = channels
-        self.dig_sample_bits = dig_sample_bits
+        self.input_sample_bits = input_sample_bits
         self.unzip_factor = unzip_factor
-        if dig_sample_bits < 2 or (dig_sample_bits > 10 and dig_sample_bits not in {12, 16}):
-            raise ValueError("dig_sample_bits must be 2-10, 12 or 16")
+        self.complex_input = complex_input
+        if complex_input:
+            if input_sample_bits != 64:
+                raise ValueError("input_sample_bits must be 64 when complex_input is true")
+        else:
+            if input_sample_bits not in DIG_SAMPLE_BITS_VALID:
+                raise ValueError("input_sample_bits must be 2-10, 12 or 16 when complex_input is false")
         if (2 * channels) % self.wgs != 0:
             raise ValueError(f"2*channels must be a multiple of {self.wgs}")
         if channels <= 1 or channels & (channels - 1):
@@ -73,8 +90,9 @@ class PFBFIRTemplate:
                     "wgs": self.wgs,
                     "taps": self.taps,
                     "channels": channels,
-                    "dig_sample_bits": dig_sample_bits,
+                    "input_sample_bits": input_sample_bits,
                     "unzip_factor": unzip_factor,
+                    "complex_input": complex_input,
                 },
                 extra_dirs=[str(resource_dir)],
             )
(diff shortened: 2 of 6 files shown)


src/katgpucbf/fgpu/pfb.py
```diff
@@ -40,25 +41,41 @@ class PFBFIRTemplate:
...
     def __init__(
-        self, context: AbstractContext, taps: int, channels: int, dig_sample_bits: int, unzip_factor: int = 1
+        self,
+        context: AbstractContext,
+        taps: int,
+        channels: int,
+        input_sample_bits: int,
+        unzip_factor: int = 1,
+        *,
+        complex_input: bool = False,
```
amishatishpatel, line 63: I know it's about as obvious as a parameter can get, but do you think it's worth having a oneliner under Parameters in the docstring?

src/katgpucbf/fgpu/pfb.py
```diff
@@ -150,34 +183,44 @@ def __init__(
...
             return
-        step = 2 * self.template.channels
+        rps = 2 if self.template.complex_input else 1  # real values per sample
+        real_step = 2 * self.template.channels  # step size in real values
+        sample_step = real_step // rps  # step size in samples
         in_buffer = self.buffer("in")
         out_buffer = self.buffer("out")
-        in_buffer_samples = in_buffer.shape[0] * BYTE_BITS / self.template.dig_sample_bits
-        if self.in_offset + step * (self.spectra + self.template.taps - 1) > in_buffer_samples:
+        in_buffer_bytes = in_buffer.shape[0] * in_buffer.dtype.itemsize
```
amishatishpatel, line 221: I see I see

src/katgpucbf/fgpu/pfb.py
```diff
@@ -189,25 +232,28 @@ def _run(self) -> None:
...
-        groupsy = max(groupsy, accel.divup(128 * 1024, step))
+        groupsy = max(groupsy, accel.divup(128 * 1024, real_step))
         # Re-compute work_spectra to balance the load
         work_spectra = accel.divup(self.spectra, groupsy)
-        stepy = work_spectra * step
+        stepy = work_spectra * real_step
         # Rounding up may have left some workgroups with nothing to do, so recalculate
         # groupsy again.
         groupsy = accel.divup(self.spectra, work_spectra)
+        args = [
```
amishatishpatel, line 242: Not at all priority, but would you consider renaming this to `kernel_args`?

test/fgpu/test_pfb.py
```diff
@@ -82,5 +99,35 @@ def test_pfb_fir(
...
-    np.testing.assert_allclose(h_out, expected_out, rtol=1e-5, atol=1e-6 * 2**dig_sample_bits)
+    np.testing.assert_allclose(h_out, expected_out, rtol=1e-5, atol=1e-6 * 2**input_sample_bits)
     assert h_total_power == expected_total_power
+
+
+@pytest.mark.parametrize("unzip_factor", [1, 2, 4])
+def test_pfb_fir_complex(context: AbstractContext, command_queue: AbstractCommandQueue, unzip_factor: int) -> None:
+    taps = 16
+    spectra = 3123
+    channels = 4096
```
amishatishpatel, line 110: Once again, not at all priority, but should we earmark a general 'consolidation of variable declarations'? Is it worth declaring these as fixtures?

amishatishpatel (commented):
A few small comments, honestly nothing serious. Before I explicitly approve, I see the documentation update checkbox is still unchecked (and no file to review here) - should we wait here for another round of reviewing?

src/katgpucbf/fgpu/pfb.py
```diff
@@ -122,11 +142,24 @@ class PFBFIR(accel.Operation):
...
+        See above
+    **weights** : channels*taps, float32
+        See above
+
     Raises
     ------
     ValueError
-        If ``samples`` is not a multiple of 8, or if ``2*channels`` is not a
-        multiple of the workgroup size (currently 128).
+        If ``samples`` is not a multiple of 8 and ``complex_input`` is false
```
james-smith-za, line 166: Or if complex_input is true and input_sample_bits is not 64

bmerry:
Thanks @amishatishpatel, they're all excellent points and I've push a new commit to address them.

test/fgpu/test_pfb.py
amishatishpatel, line 1: Man, these changes make it look so slick. Thank you.

james-smith-za (commented):
Some minor comments, but nothing that I feel strongly enough about to prevent you from merging.

src/katgpucbf/fgpu/pfb.py
```diff
@@ -33,6 +33,13 @@
 class PFBFIRTemplate:
     """Template for the PFB-FIR operation.
 
+    The operation can operate in two different modes. In the first mode
```
amishatishpatel, line 36: I feel funny (in a good way) about reading the words "This operation can operate".

amishatishpatel (approved):
Please and thank yous all around.

src/katgpucbf/fgpu/pfb.py
```diff
@@ -40,25 +41,41 @@ class PFBFIRTemplate:
         The number of taps that you want the resulting PFB-FIRs to have.
     channels
         Number of channels into which the input data will be decomposed.
-    dig_sample_bits
-        Bits per digitiser sample.
+    input_sample_bits
+        Bits per input sample. If `complex_input` is true, the input values
+        are floating-point complex numbers and this is the number of bits in
+        the complex value (and must equal 64). Otherwise, the inputs are
```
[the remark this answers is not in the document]
amishatishpatel: Yeah I was on the verge of making a similar point, just put it down to "it's so obvious/out-there, why bother". Still, good point.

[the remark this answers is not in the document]
bmerry: I was going to do that at first, but I wanted to use this value to convert between number of samples and number of bytes. Let me check if I'm actually depending on that though.

src/katgpucbf/fgpu/compute.py
```diff
@@ -116,7 +116,7 @@ def __init__(
         spectra: int,
         spectra_per_heap: int,
     ) -> None:
-        self.dig_sample_bits = template.pfb_fir.dig_sample_bits
+        self.dig_sample_bits = template.pfb_fir.input_sample_bits
```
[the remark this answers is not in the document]
bmerry: > If you're going to rename the parameter of the pfb_fir, it makes some sense to rename the above in the "compute", earlier in the file.

I renamed the pfb_fir parameter because it does not always refer to the number of bits in digitiser samples; in `complex_input` mode it is instead the number of bits that come out of the DDC. On the other hand, I wanted this attribute to refer to the bits in digitiser samples regardless.

> If the data is needed, it can just be accessed as self.template.pfb_fir.input_sample_bits anyway.

Once the DDC is wired in, it becomes more complicated. For wideband it will be `self.template.pfb_fir.input_sample_bits` but for narrowband it will be `self.template.ddc.dig_sample_bits`.

> It doesn't look as though this member variable actually ever gets used. Perhaps it can be removed?

That's a reasonable point - I'm not sure why I have it. Will try deleting it and see what happens.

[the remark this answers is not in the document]
bmerry: It looks like it can be removed, but doing so on this PR will cause merge conflicts with later work I've done. If you don't object, I'll rather remove it in a future PR.

src/katgpucbf/fgpu/pfb.py
bmerry: The ValueError documentation was a bit of a mess (some conditions actually checked on the template were documented here, and a lot of conditions checked on the template were not documented at all). Hopefully my latest push improves things.

james-smith-za (approved):
Happy.

PR #554, opened 2023-04-12
Update latex preamble for qualification report

james-smith-za:
to match some of the revisions that I've made recently in dsp_group_documents.

It's a little bit more dynamic.

Closes NGC-880

Report:
[report.pdf](https://github.com/ska-sa/katgpucbf/files/11211978/report.pdf)
(all the tests are failing, it was from an old problematic json that I had from a while ago, but the preamble is what you need to see here so it should be fine.)


amishatishpatel:
Should I/we wait until you test with an updated JSON file before reviewing?

james-smith-za:
Updated report:
[report.pdf](https://github.com/ska-sa/katgpucbf/files/11220899/report.pdf)
I notice that there are some failing tests, and for some reason their names don't seem to appear properly in the result summary table.
I don't *think* this is due to something that I've introduced. I'm going to generate a PDF with `main` branch quickly to check. I'll file a bug if I see it there as well.

qualification/report/katdoc.sty
```diff
@@ -103,147 +123,121 @@
...
 
-      % Approval Table
-      \LARGE
-      \textbf{Document Approval}
+    \normalsize
+    \begin{tabularx}{\textwidth}{|l|l|l|l|X|X|}
+      \hline
+      \textbf{Function} & \textbf{Name} & \textbf{Designation} & \textbf{Affiliation} & \textbf{Date} & \textbf{Signature} \\
+      \hline
+      \docApproval%
```
amishatishpatel, line 145: Some lines end with `%` characters, any particular reason(s) why that is? I thought earlier on it meant "it's a multiline command coming", perhaps e.g. `\docApproval` is a multiline cmd declared elsewhere?

EDIT: Same with `docHistory` and `docSoftware` further down, previously they didn't have a trailing `%`.

qualification/report/katdoc.sty
```diff
@@ -255,68 +249,3 @@
...
-
-  \ifdefined\referenceDocs
-    \begin{enumerate}
-      \referenceDocs
-    \end{enumerate}
-  \else
-    \begin{enumerate}
-      \item \emph{No reference documents}
-    \end{enumerate}
-  \fi
```
amishatishpatel, line 320: Ah, good call.

qualification/report/katdoc.sty
```diff
@@ -103,147 +123,121 @@
           \hline
         \end{tabularx}
       \end{table}
+    \end{center}
+  \end{titlepage}
+  %\stepcounter{page}
```
amishatishpatel, line 128: I see previously there were commented-out lines of code/LaTeX still in the file (mainly `\vspace`), is this left in mostly as a reminder?

amishatishpatel (commented):
A couple minor comments before explicit approval.

qualification/report/katdoc.sty
james-smith-za: There's a tool called checklatex (I think) which moans about there being spaces right after the command.

There are two ways to silence that which are technically more correct, either put a % directly after, or an empty {} I think. I usually default to the former.

qualification/report/katdoc.sty
james-smith-za: Yeah, in the old katdoc.sty there were hacky ways to get the bibliography in. We've changed that because we didn't like it. This was just the clutter that has been cut.

amishatishpatel (approved):
Good to go from my side, although my LaTeX Fu is Blue Belt at best.

qualification/report/katdoc.sty
```diff
@@ -1,64 +1,84 @@
...
 \usepackage{float}
 \usepackage{enumitem}
+\usepackage{mathtools}
+\usepackage{tablefootnote}
+\usepackage{color}
+\usepackage{colortbl}
 
 % For plotting to work nicely.
-\usepackage[utf8]{inputenc}
+\usepackage{pgf}
```
bmerry, line 26: Is this necessary, or will pgfplots suck it in anyway?

qualification/report/katdoc.sty
```diff
@@ -1,64 +1,84 @@
...
 \usepgfplotslibrary{groupplots,dateplot}
 \usetikzlibrary{patterns,shapes.arrows}
 \pgfplotsset{compat=newest}
 
-\geometry{verbose,a4paper,tmargin=32mm,bmargin=27mm,lmargin=21mm,rmargin=19mm,headheight=10mm,headsep=12mm,footskip=12mm}
-\pagestyle{empty}
+\geometry{nomarginpar,verbose,a4paper,tmargin=32mm,bmargin=27mm,lmargin=21mm,rmargin=19mm,headheight=10mm,headsep=12mm,footskip=12mm}
 \setlength{\parskip}{\medskipamount}
 \setlength{\parindent}{0pt}
-\pagenumbering{arabic}
```
bmerry, line 36: I'm guessing this is removed because it's the default?

qualification/report/katdoc.sty
```diff
@@ -1,64 +1,84 @@
...
+
+  \begin{tabular}{|p{0.1\textwidth}|p{0.84\textwidth}|}
+    \hline
+    \abbreviations%
+  \end{tabular}
+}
+
+% Extract version information from pdftex
+\def\extractpdftexversion#1-#2-#3 #4\@nil{#3}%
+\def\myversion{\expandafter\extractpdftexversion\pdftexbanner\@nil}%
```
bmerry, line 74: Is it worth trying to protect this so that it produces something sane when compiled with a different engine?

qualification/report/katdoc.sty
bmerry: It seems a bit odd that it would complain about spaces after a command, because a "feature" of commands is that they eat the following whitespace. Normally when I've seen end-of-line `%` in LaTeX it's to prevent the end-of-line being interpreted as whitespace that ends up as space in the document e.g. `\newcommand{\makeInfoPage}{%`

qualification/report/katdoc.sty
```diff
@@ -103,147 +123,120 @@
           \hline
         \end{tabularx}
       \end{table}
+    \end{center}
+  \end{titlepage}
+}
 
-      \newpage{}
+\newcommand{\makeInfoPage}{%
```
bmerry, line 130: Suggest making this `\makeinfopage` for consistency with `\maketitle`.

qualification/report/katdoc.sty
```diff
@@ -103,147 +123,120 @@
...
+
+  \renewcommand{\headrulewidth}{0.4pt}
+  \renewcommand{\footrulewidth}{0.4pt}
 }
 
 % The entire beginning part of the document
 
-\newcommand{\makekatdocbeginning}{
+\newcommand{\makekatdocbeginning}{%
+  \pagenumbering{roman}
```
bmerry, line 231: Do I understand correctly that you're changing the preamble from Arabic to Roman numbering? SE/config management might not like that if it's not consistent with their templates (I recall Simon once getting all but one signature on a document, with someone insisting it had to be redone because the page numbering wasn't done right).

It also means that the "Page N or M" footer will only be counting the number of Roman-numbered pages, not the number of pages in the document.

bmerry (changes requested):

qualification/report/katdoc.sty
james-smith-za: It was somewhere previously, I just rearranged it.
Turns out it isn't, so I'll remove it.

qualification/report/katdoc.sty
james-smith-za: No, because the preamble pages are numbered with roman numerals. It changes back to arabic later on (line 238 as I look at it at the moment)

qualification/report/katdoc.sty
james-smith-za: I don't see myself ever bothering to use a different engine.

It's probably better practice to do so, but I can forsee myself spending some hours figuring out how to do that, which is not really an exercise that I'm interested in right now.

qualification/report/katdoc.sty
james-smith-za: This matches what is currently done in the design doc, which was an issue that @tyronevb flagged. So I'm presuming that this is what SE wants.

qualification/report/katdoc.sty
james-smith-za: Without it, `chktex` (that's what it's called) gives me this:
image

With the `%`, the message goes away.

I can remove them if you want, and just disable the warning. They don't actually affect the output. I just got the impression that it might be considered best practice, since chktex complains.

qualification/report/katdoc.sty
bmerry: Ok, if it's what they want I'm happy with it.

james-smith-za: I see what you mean. @tyronevb, can you comment on what the SE style guide says?

qualification/report/katdoc.sty
bmerry: Just tested with lualatex and it reports that `\pdftexbanner` is an undefined control sequence, which I think is an acceptable outcome (it doesn't produce a document with incorrect information, and the error is a fairly good hint). You can detect whether the engine is pdftex with `\ifpdftex`, but I found it wasn't all that helpful because while you can emit an error message, latexmk keeps rolling on and it gets lost in amongst all the other errors.

bmerry (changes requested):
There still seem to be more libraries imported than necessary, although I appreciate that some of them were imported (but possibly also some of them were used in the bits of code you deleted).

qualification/report/katdoc.sty
bmerry: The Internet showed me a simpler solution: add `\RequirePDFTeX` (e.g. after importing all the other packages - it's provided by the `iftex` package so for completeness you could import that). If using the wrong engine, it prints the following then aborts:
```
 ********************************************
 * pdfTeX is required to compile this document.
 * Sorry!
 ********************************************
```

qualification/report/katdoc.sty
```diff
@@ -17,13 +17,8 @@
...
-\usepackage{mathtools}
-\usepackage{tablefootnote}
-\usepackage{color}
-\usepackage{colortbl}
 
 % For plotting to work nicely.
-\usepackage{pgf}
 \usepackage{pgfplots}
 \DeclareUnicodeCharacter{2212}{−}
 \usepgfplotslibrary{groupplots,dateplot}
```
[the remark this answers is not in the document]
james-smith-za: It's needed for generating the report. I guess a more natural home for it though would be the preamble, to have this style file more generically applicable?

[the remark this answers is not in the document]
bmerry: That's what I was getting at.

james-smith-za:
Updated report
[test_report.pdf](https://github.com/ska-sa/katgpucbf/files/11259265/test_report.pdf)

PR #555, opened 2023-04-14
Misc narrowband fixes

bmerry:
Assorted bug fixes and helper properties. Many of these issues were found through unit testing, but I'm keeping the unit test updates for a future PR because they're quite substantial.

Relates to NGC-949.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -547,6 +547,7 @@ class Pipeline:
                     saturated=item.saturated.empty_like(),
                     substreams=substreams,
                     feng_id=self.engine.feng_id,
+                    spectra_samples=self.output.spectra_samples,
                 )
                 item.chunk = chunk
                 send_chunks.append(chunk)
@@ -564,6 +565,7 @@ class Pipeline:
                         accel.HostArray((heaps, N_POLS), np.uint32, context=self._compute.template.context),
                         substreams=substreams,
                         feng_id=self.engine.feng_id,
+                        spectra_samples=self.output.spectra_samples,
                     )
                 )
                 self._send_free_queue.put_nowait(send_chunks[-1])
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -782,7 +782,10 @@ class Pipeline:
                 if batch_spectra > 0:
                     logging.debug("Processing %d spectra", batch_spectra)
                     out_slice = np.s_[self._out_item.n_spectra : self._out_item.n_spectra + batch_spectra]
-                    self._out_item.fine_delay[out_slice] = fine_delays.T / self.output.subsampling
+                    # Convert units of fine delay from digitiser samples to phase slope
+                    # across the band. Narrowband has `decimation` times less bandwidth,
+                    # so the phase change across the band is that much less.
+                    self._out_item.fine_delay[out_slice] = fine_delays.T / self.output.decimation
                     # The phase is referenced to the centre frequency, but
                     # coarse delay in wideband affects the phase of the centre
                     # frequency (each sample shifts it by pi/2) and this
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -55,7 +55,7 @@ from ..utils import DeviceStatusSensor, TimeConverter, add_time_sync_sensors
 from . import DIG_POWER_DBFS_HIGH, DIG_POWER_DBFS_LOW, INPUT_CHUNK_PADDING, recv, send
 from .compute import Compute, ComputeTemplate, NarrowbandConfig
 from .delay import AbstractDelayModel, AlignedDelayModel, LinearDelayModel, MultiDelayModel, wrap_angle
-from .output import NarrowbandOutput, Output
+from .output import NarrowbandOutput, Output, WidebandOutput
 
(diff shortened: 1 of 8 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -1581,21 +1590,13 @@ def comma_string_to_float(comma_string: str) -> tuple[float, float]:
...
 
             delay_samples = delay * self.adc_sample_rate
-            # For compatibility with MeerKAT, the phase given is the net change in
-            # phase for the centre frequency, including delay, and we need to
-            # compensate for the effect of the delay at that frequency. The centre
-            # frequency is 4 samples per cycle, so each sample of delay reduces
-            # phase by pi/2 radians.
-            delay_phase_correction = 0.5 * np.pi * delay_samples
-            phase += delay_phase_correction
-            phase_rate_correction = 0.5 * np.pi * delay_rate
```
james-smith-za, line 1591: Was this bit not correct? Was it replaced by what is now on lines 785 - 798?

test/fgpu/test_postproc.py
```diff
@@ -47,9 +47,7 @@ def postproc_host_pol(
         # Throw out last channel (Nyquist frequency)
         data = data_rfft.astype(np.complex64)[:, :channels]
     # Compute delay phases
-    channel_idx = np.arange(channels, dtype=np.float32)[np.newaxis, :]
-    if complex_pfb:
-        channel_idx -= channels / 2
+    channel_idx = np.arange(channels, dtype=np.float32)[np.newaxis, :] - channels / 2
```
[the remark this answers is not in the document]
bmerry: Yes. See the commit message on 2637739.

src/katgpucbf/fgpu/engine.py
bmerry: Replaced by line 797 for the coarse delay and line 261 in postproc.mako for the fine delay.

james-smith-za (approved):

src/katgpucbf/fgpu/engine.py
```diff
@@ -1581,21 +1590,13 @@ def comma_string_to_float(comma_string: str) -> tuple[float, float]:
             phase, phase_rate = comma_string_to_float(phase_str)
 
             delay_samples = delay * self.adc_sample_rate
-            # For compatibility with MeerKAT, the phase given is the net change in
-            # phase for the centre frequency, including delay, and we need to
-            # compensate for the effect of the delay at that frequency. The centre
-            # frequency is 4 samples per cycle, so each sample of delay reduces
-            # phase by pi/2 radians.
```
amishatishpatel, line 1588: Do we need to maintain the "For compatibility with MeerKAT" component of this comment?
For ease of reference, the comment added on Line 789+ reads
> The phase is referenced to the centre frequency, but coarse delay in wideband affects the 
phase of the centre frequency (each sample shifts it by pi/2) and this needs to be corrected for.

Perhaps also "The centre frequency is 4 samples per cycle" to explain the `np.array(start_coarse_delays) % 4` (?)

src/katgpucbf/fgpu/engine.py
```diff
@@ -1581,21 +1590,13 @@ def comma_string_to_float(comma_string: str) -> tuple[float, float]:
...
-            delay_phase_correction = 0.5 * np.pi * delay_samples
-            phase += delay_phase_correction
-            phase_rate_correction = 0.5 * np.pi * delay_rate
             new_linear_models.append(
                 LinearDelayModel(
                     start_sample_count,
                     delay_samples,
                     delay_rate,
                     phase,
-                    phase_rate / self.adc_sample_rate + phase_rate_correction,
```
amishatishpatel, line 1598: Not that I'm emotionally invested in it, but what happened to `phase_rate_correction`?
* Is it being applied/incorporated somehow elsewhere?
* Or is the updated approach of delay handling/calculation handling this correction factor?

src/katgpucbf/fgpu/main.py
```diff
@@ -185,16 +185,17 @@ def field_callback(kws: WidebandOutputDict, key: str, data: str) -> None:
     try:
         kws: WidebandOutputDict = {}
         _parse_stream(value, kws, field_callback)
+        kws = {"taps": DEFAULT_TAPS, "w_cutoff": DEFAULT_W_CUTOFF, **kws}  # type: ignore
```
amishatishpatel, line 188: Very interesting, the presence of `taps` or `w_cutoff` in `kws` overwrites (?) the default accommodation here. I did not know that was a thing! i.e. Specifying a key's value twice in dict instantiation and it takes the last mention.

amishatishpatel (commented):
As always, a couple comments for clarification before explicit approval. Please and thank you for your patience.

src/katgpucbf/fgpu/engine.py
bmerry: We're now handling phase and phase_rate with the same interpretation as they're given i.e. the total phase adjustment at the centre of the band, instead of at channel 0. So this fiddling is no longer needed.

src/katgpucbf/fgpu/engine.py
bmerry: I don't think we particularly need to state that it's for MeerKAT compatibility; it is defined by the ICD, and a history lesson doesn't really matter. But I'll clarify the 4 samples per cycle and the `% 4`.

amishatishpatel (approved):

PR #558, opened 2023-04-19
Add --narrowband option to sim_correlator.py

bmerry:
There are also some `--narrowband-foo options` to customise the narrowband, but they have defaults that make it possible to just launch.

Closes NGC-574.


bmerry:
This probably just needs one reviewer - whoever gets to it first.

james-smith-za (approved):
I don't really like `narrow0-antenna-channelised-voltage` that much as a stream name, but it works I guess. I don't have a better suggestion. I guess we need to come up with *something* if they're going to share a subarray.

bmerry:
> I don't really like `narrow0-antenna-channelised-voltage` that much as a stream name, but it works I guess. I don't have a better suggestion. I guess we need to come up with _something_ if they're going to share a subarray.

Yeah, [naming is hard](https://martinfowler.com/bliki/TwoHardThings.html)

PR #559, opened 2023-04-19
Add a qualification test for CBF-REQ-0103

bmerry:
There is still some discussion about exactly how to interpret the requirement, but the good news is that we meet it with a massive safety margin.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/11271553/report.pdf)

Closes NGC-941.

### DIFF
--- a/qualification/baseline_correlation_products/test_saturation.py
+++ b/qualification/baseline_correlation_products/test_saturation.py
@@ -1,0 +1,55 @@
+################################################################################
+# Copyright (c) 2022-2023, National Research Foundation (SARAO)
+#
+# Licensed under the BSD 3-Clause License (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+"""Test probability of saturation."""
+
+import math
+
+import pytest
+import scipy.stats
+from pytest_check import check
+
+from .. import BaselineCorrelationProductsReceiver
+from ..reporter import Reporter
+
+
+@pytest.mark.requirements("CBF-REQ-0103")
+async def test_saturation(
+    receive_baseline_correlation_products: BaselineCorrelationProductsReceiver,
+    pdf_report: Reporter,
+) -> None:
+    r"""Check probability of saturation.
+
+    Verification method
+    -------------------
+    Verification by analysis. Saturation is most likely for autocorrelations.
+    Consider input (channelised) noise :math:`X` where :math:`E[|X|^2] =
+    \sigma^2`. Let :math:`k` be the number of spectra accumulated, and the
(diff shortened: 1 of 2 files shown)


qualification/baseline_correlation_products/test_saturation.py
```diff
@@ -0,0 +1,55 @@
...
+    Verification by analysis. Saturation is most likely for autocorrelations.
+    Consider input (channelised) noise :math:`X` where :math:`E[|X|^2] =
+    \sigma^2`. Let :math:`k` be the number of spectra accumulated, and the
+    autocorrelation of :math:`X` be :math:`Y`. Then :math:`\frac{2Y}{\sigma^2}`
+    has a :math:`\chi^2` distribution with :math:`2k` degrees of freedom (the
+    factors of 2 are because :math:`X` is a complex variable whose real and
+    imaginary components are independent Gaussian distributions).  We can use
+    this to analytically determine the value of :math:`\sigma` for which
+    :math:`P[Y \ge 2^{31}] = 10^{-4}`.
+    """
```
james-smith-za, line 47: My memory of these kinds of things is getting rather hazy, so the only real comment that I can make is to suggest that we include some kind of definition for E and P.  And maybe sigma for completeness.

bmerry: I disagree. It's a test report, not a statistics textbook, and these are very standard notations and I'm not aware of any ambiguities in conventions. Someone reviewing a requirement expressed with a statistical definition should expect to encounter some basic statistics.

I guess I haven't explicitly said that the noise is assumed to be complex zero-mean Gaussian, and uncorrelated in time, which might be worthwhile.

Incidentally, sigma *is* defined (`E[|X|^2] = \sigma^2`); I avoided referring to it as "standard deviation" because for a complex Gaussian some people seem to use that term to refer to the standard deviation of the real and imaginary components, and that's different to the definition I used here.

james-smith-za: If I see E the first thing I think of is [energy in a signal](https://en.wikipedia.org/wiki/Energy_(signal_processing)). So I (as [a moron in a hurry](https://en.wikipedia.org/wiki/A_moron_in_a_hurry)) got confused. The definitions of energy in a signal looks similar enough (with an x^2 term) to what you've written to make me think there was something that I have forgotten from undergraduate days about standard deviations being involved in the energy of a noise signal? But noise is typically classified as a power signal, not an energy signal?

Now that you mention statistics, it's clear that I should have read "expected value", in which case the confusion goes away. 

P can also be power in signal-processing parlance (considering my above misunderstanding, it's the first thing that jumped to my mind) but I guess that's my bad for not making the link with the wording of the requirement properly.

qualification/report/requirements_text/CBF-REQ-0103.tex
```diff
@@ -0,0 +1,4 @@
+The CBF shall have a probability of saturation of the Baseline Correlation
+Products for the maximum permissible accumulation interval of
+$\le 0.01\%$, with a Gaussian noise input signal with
```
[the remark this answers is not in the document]
bmerry: 0.01% = 0.0001 = 10^-4.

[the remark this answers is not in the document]
james-smith-za: Ah yes, of course. Forgot about the percent.

james-smith-za (approved):

qualification/baseline_correlation_products/test_saturation.py
ludwigschwardt: The stats look good.

PR #565, opened 2023-05-15
Use vkgdr for small host->device copies

bmerry:
This prevents them being blocked behind large transfers of digitiser
samples (see NGC-973).

The gains are now only copied when necessary.

Closes NGC-973.

### DIFF
--- a/doc/fgpu.design.rst
+++ b/doc/fgpu.design.rst
@@ -391,6 +391,14 @@ Fine delays and the twiddle factor for the Cooley-Tukey transformation are
 computed using the ``sincospi`` function, which saves both a multiplication by
 :math:`\pi` and a range reduction.
 
+The gains, fine delays and phases need to be made available to the kernel. We
+found that transferring them through the usual CUDA copy mechanism leads to
+sub-optimal scheduling, because these (small) transfers could end up queued
+behind the much larger transfers of digitiser samples. Instead, we use `vkgdr`_
+to allow the CPU to write directly to the GPU buffers. The buffers are
+replicated per output item, so that it is possible for the CPU to be updating
+the values for one output item while the GPU is computing on another.
+
 Coarse delays
 ^^^^^^^^^^^^^
 One of the more challenging aspects of the processing design was the handling
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,6 +35,7 @@ vkgdr = vkgdr
 
 gpu =
     katsdpsigproc[CUDA]
+    vkgdr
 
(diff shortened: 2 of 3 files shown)


james-smith-za (approved):
This seems to make vkgdr compulsory rather than optional.

Do the `--use-vkgdr` options make sense if you now have to have it anyway?
(Perhaps this is less relevant for xbgpu, but it looks like for fgpu we may as well remove this.)

bmerry:
> This seems to make vkgdr compulsory rather than optional.
> 
> Do the `--use-vkgdr` options make sense if you now have to have it anyway? (Perhaps this is less relevant for xbgpu, but it looks like for fgpu we may as well remove this.)

It probably needs an update to the help description (I'd meant to do that but forgot - I blame still being a bit sick). It still controls whether to use vkgdr for the bulk data (digitiser samples); vkgdr is only mandatory for the small bits of data (fine-delay, phase, gain).

james-smith-za:
>  It still controls whether to use vkgdr for the bulk data (digitiser samples); vkgdr is only mandatory for the small bits of data (fine-delay, phase, gain).

Since at this point vkgdr is installed anyway, is there a condition under which the user may opt *not* to use vkgdr for the bulk digitiser samples?

src/katgpucbf/fgpu/main.py
```diff
@@ -422,7 +422,9 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         help="Mask off bottom bits of timestamp (workaround for broken digitiser)",
     )
     parser.add_argument(
-        "--use-vkgdr", action="store_true", help="Assemble chunks directly in GPU memory (requires Vulkan)"
+        "--use-vkgdr",
+        action="store_true",
+        help="Assemble chunks directly in GPU memory (requires sufficient GPU BAR space)",
```
amishatishpatel, line 427: I think we can all agree the argument `--use-vkgdr` is pretty unambiguous as it is. _But_, still worth asking
* Can you adjust GPU BAR space without Vulkan?
* Or rather, you can adjust it without Vulkan, it just makes more/most sense to do it with a package that takes advantage of it (?)

amishatishpatel (approved):
One comment/question, otherwise good to go.

src/katgpucbf/fgpu/main.py
bmerry: The amount of BAR space cannot be adjusted at all. Vulkan just lets you use the available space.

PR #566, opened 2023-05-16
fgpu: reduce InItems from 4 to 3

bmerry:
This reduces the memory footprint and doesn't seem to cost any performance.


amishatishpatel (approved):
Of course, this change is so deliberate you must have actively looked into it. Should I/we request proof this is the case? Slash, how you came to this conclusion?

bmerry:
> Of course, this change is so deliberate you must have actively looked into it. Should I/we request proof this is the case? Slash, how you came to this conclusion?

I ran some benchmarks using the tools I developed for the F-engine paper, using either 1 or 4 F-engines and with 1/8  narrowband.

PR #567, opened 2023-05-16
Make output Prom counters exist from the start

bmerry:
Because narrowband support changes made them labelled by the output stream, they didn't exist at all until the first time they were referenced for a specific stream. Ensure that they get referenced (and hence made to exist) before any data is transmitted.

Closes NGC-956.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -47,7 +47,7 @@ class PromDiff:
         self,
         *,
         registry: prometheus_client.CollectorRegistry = prometheus_client.REGISTRY,
-        namespace: str | None = None
+        namespace: str | None = None,
     ) -> None:
         self._registry = registry
         self._before: list[prometheus_client.samples.Sample] = []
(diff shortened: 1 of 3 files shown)


test/__init__.py
```diff
@@ -82,14 +82,16 @@ def get_sample_diff(self, name: str, labels: dict[str, str] | None = None) -> fl
...
+        raises an exception.
         """
         if labels is None:
             labels = {}
         before = self._get_value(self._before, name, labels)
         after = self._get_value(self._after, name, labels)
         if before is not None and after is not None:
             return after - before
+        elif after is not None:
+            raise ValueError(f"Metric {name}{labels} did not exist at start")
```
amishatishpatel, line 94: You can put this here/declare it so because you're confident `before is None` at this point ?

amishatishpatel (approved):

test/__init__.py
bmerry: Exactly.

PR #568, opened 2023-05-17
xbgpu: add command-line arguments for beamformer

amishatishpatel:
Much like #535, this adds a `--beamformer` command-line argument, taking properties to set via comma-separate key=value pairs.

I may have overstepped the boundaries of the PR in some way as I have an eye on NGC-929 (a la PR #537 )
* I purposefully left a `self.n_ants` attribute in `XBEngine` (used in the `_sender_loop`), and
* I updated `populate_sensors` to take the `chan-range` sensor default values as an arg.

It's a conundrum I regularly dance around (do the straightforward thing or be less abrupt with the change), please share your grievances accordingly.

- [x] If modules are added/removed: use sphinx-apidoc to update files in `doc/`

Closes NGC-905.

### DIFF
--- a/scratch/xbgpu/run-xbgpu.sh
+++ b/scratch/xbgpu/run-xbgpu.sh
@@ -46,6 +46,7 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --src-interface $iface \
     --dst-interface $iface \
     --src-ibv --dst-ibv \
+    --corrprod=name=bcp1,dst=$dst_mcast \
     --adc-sample-rate ${adc_sample_rate:-1712000000} \
     --array-size ${array_size:-64} \
     --spectra-per-heap ${spectra_per_heap:-256} \
@@ -57,4 +58,4 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --sync-epoch 0 \
     --katcp-port $port \
     --tx-enabled \
-    $src_mcast $dst_mcast
+    $src_mcast
--- a/scratch/xbgpu/run-xbgpu.sh
+++ b/scratch/xbgpu/run-xbgpu.sh
@@ -46,7 +46,7 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --src-interface $iface \
     --dst-interface $iface \
     --src-ibv --dst-ibv \
-    --corrprod=name=bcp1,dst=$dst_mcast \
+    --corrprod=name=bcp1,heap_accumulation_threshold=${heap_accumulation_threshold:-52},samples_between_spectra=${samples_between_spectra:-$((channels*2))},spectra_per_heap=${spectra_per_heap:-256},dst=$dst_mcast \
     --adc-sample-rate ${adc_sample_rate:-1712000000} \
     --array-size ${array_size:-64} \
     --spectra-per-heap ${spectra_per_heap:-256} \
--- a/scratch/xbgpu/run-xbgpu.sh
+++ b/scratch/xbgpu/run-xbgpu.sh
@@ -54,7 +54,6 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --channels-per-substream $channels_per_substream \
     --samples-between-spectra ${samples_between_spectra:-$((channels*2))} \
     --channel-offset-value $channel_offset \
-    --heap-accumulation-threshold ${heap_accumulation_threshold:-52} \
     --sync-epoch 0 \
     --katcp-port $port \
     --tx-enabled \
--- a/scratch/xbgpu/run-xbgpu.sh
+++ b/scratch/xbgpu/run-xbgpu.sh
@@ -46,7 +46,7 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --src-interface $iface \
     --dst-interface $iface \
     --src-ibv --dst-ibv \
-    --corrprod=name=bcp1,heap_accumulation_threshold=${heap_accumulation_threshold:-52},samples_between_spectra=${samples_between_spectra:-$((channels*2))},spectra_per_heap=${spectra_per_heap:-256},dst=$dst_mcast \
+    --corrprod=name=bcp1,heap_accumulation_threshold=${heap_accumulation_threshold:-52},dst=$dst_mcast \
     --adc-sample-rate ${adc_sample_rate:-1712000000} \
     --array-size ${array_size:-64} \
     --spectra-per-heap ${spectra_per_heap:-256} \
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -42,7 +42,6 @@ import numpy as np
 import spead2.recv
 from aiokatcp import DeviceServer
 from katsdpsigproc import accel
-from katsdptelstate.endpoint import Endpoint
 
 from .. import (
     DESCRIPTOR_TASK_NAME,
@@ -61,7 +60,7 @@ from ..send import DescriptorSender
 from ..utils import DeviceStatusSensor, TimeConverter, add_time_sync_sensors
 from . import recv
 from .correlation import Correlation, CorrelationTemplate
-from .output import Output, XengineOutput
+from .output import Output, XOutput
 from .xsend import XSend, incomplete_accum_counter, make_stream
 
 logger = logging.getLogger(__name__)
@@ -197,7 +196,7 @@ class XBEngine(DeviceServer):
         spectrum reassembly by the downstream receiver.
     outputs
         Output streams to generate. Currently this must be a single
-        XengineOutput.
+        XOutput.
     src
         Endpoint for the incoming data.
     src_interface
@@ -267,7 +266,6 @@ class XBEngine(DeviceServer):
         src_affinity: int,
         src_comp_vector: int,
         src_buffer: int,
-        dst: Endpoint,
         dst_interface: str,
         dst_ttl: int,
         dst_ibv: bool,
@@ -285,7 +283,7 @@ class XBEngine(DeviceServer):
 
         # B-engine doesn't work yet
         assert len(outputs) == 1
-        assert isinstance(outputs[0], XengineOutput)
+        assert isinstance(outputs[0], XOutput)
 
         if sample_bits != 8:
             raise ValueError("sample_bits must equal 8 - no other values supported at the moment.")
@@ -443,8 +441,8 @@ class XBEngine(DeviceServer):
             context=self.context,
             packet_payload=dst_packet_payload,
             stream_factory=lambda stream_config, buffers: make_stream(
-                dest_ip=dst.host,
-                dest_port=dst.port,
+                dest_ip=outputs[0].dst[0].host,
+                dest_port=outputs[0].dst[0].port,
                 interface_ip=dst_interface,
                 ttl=dst_ttl,
                 use_ibv=dst_ibv,
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -61,6 +61,7 @@ from ..send import DescriptorSender
 from ..utils import DeviceStatusSensor, TimeConverter, add_time_sync_sensors
 from . import recv
 from .correlation import Correlation, CorrelationTemplate
+from .output import Output, XengineOutput
 from .xsend import XSend, incomplete_accum_counter, make_stream
 
 logger = logging.getLogger(__name__)
@@ -194,6 +195,9 @@ class XBEngine(DeviceServer):
         The index of the first channel in the subset of channels processed by
         this XB-Engine. Used to set the value in the XB-Engine output heaps for
         spectrum reassembly by the downstream receiver.
+    outputs
+        Output streams to generate. Currently this must be a single
+        XengineOutput.
     src
         Endpoint for the incoming data.
     src_interface
@@ -250,16 +254,13 @@ class XBEngine(DeviceServer):
         katcp_host: str,
         katcp_port: int,
         adc_sample_rate_hz: float,
-        send_rate_factor: float,
-        n_ants: int,
-        n_channels_total: int,
-        n_channels_per_stream: int,
         n_samples_between_spectra: int,
         n_spectra_per_heap: int,
         sample_bits: int,
         heap_accumulation_threshold: int,
         sync_epoch: float,
         channel_offset_value: int,
+        outputs: list[Output],
         src: list[tuple[str, int]],  # It's a list but it should be length 1 in xbgpu case.
         src_interface: str,
         src_ibv: bool,
@@ -282,23 +283,22 @@ class XBEngine(DeviceServer):
         super().__init__(katcp_host, katcp_port)
         self._cancel_tasks: list[asyncio.Task] = []  # Tasks that need to be cancelled on shutdown
 
+        # B-engine doesn't work yet
+        assert len(outputs) == 1
+        assert isinstance(outputs[0], XengineOutput)
+
         if sample_bits != 8:
             raise ValueError("sample_bits must equal 8 - no other values supported at the moment.")
 
-        if channel_offset_value % n_channels_per_stream != 0:
-            raise ValueError("channel_offset must be an integer multiple of n_channels_per_stream")
+        if channel_offset_value % outputs[0].channels_per_substream != 0:
+            raise ValueError("channel_offset must be an integer multiple of channels_per_substream")
 
         # Array configuration parameters
         self.adc_sample_rate_hz = adc_sample_rate_hz
-        self.send_rate_factor = send_rate_factor
         self.heap_accumulation_threshold = heap_accumulation_threshold
         self.time_converter = TimeConverter(sync_epoch, adc_sample_rate_hz)
-        self.n_ants = n_ants
-        self.n_channels_total = n_channels_total
-        self.n_channels_per_stream = n_channels_per_stream
-        self.n_spectra_per_heap = n_spectra_per_heap
+        self.n_ants = outputs[0].antennas  # Still needed in sender_loop
         self.sample_bits = sample_bits
-        self.n_samples_between_spectra = n_samples_between_spectra
         self.channel_offset_value = channel_offset_value
 
         self._src = src
@@ -328,7 +328,7 @@ class XBEngine(DeviceServer):
         # configure the receiver, we pass it as a seperate argument to the
         # reciever for cases where the n_channels_per_stream changes across
         # streams (likely for non-power-of-two array sizes).
-        self.rx_heap_timestamp_step = self.n_samples_between_spectra * self.n_spectra_per_heap
+        self.rx_heap_timestamp_step = n_samples_between_spectra * n_spectra_per_heap
 
         self.populate_sensors(
             self.sensors,
@@ -339,6 +339,7 @@ class XBEngine(DeviceServer):
                 * self.rx_heap_timestamp_step
                 / adc_sample_rate_hz,
             ),
+            (channel_offset_value, channel_offset_value + outputs[0].channels_per_substream),
         )
 
         self.timestamp_increment_per_accumulation = self.heap_accumulation_threshold * self.rx_heap_timestamp_step
@@ -357,9 +358,9 @@ class XBEngine(DeviceServer):
         )
         free_ringbuffer = spead2.recv.ChunkRingbuffer(n_free_chunks)
         self._src_layout = recv.Layout(
-            n_ants=self.n_ants,
-            n_channels_per_stream=self.n_channels_per_stream,
-            n_spectra_per_heap=self.n_spectra_per_heap,
+            n_ants=outputs[0].antennas,
+            n_channels_per_stream=outputs[0].channels_per_substream,
+            n_spectra_per_heap=n_spectra_per_heap,
             sample_bits=self.sample_bits,
             timestamp_step=self.rx_heap_timestamp_step,
             heaps_per_fengine_per_chunk=self.heaps_per_fengine_per_chunk,
@@ -383,9 +384,9 @@ class XBEngine(DeviceServer):
 
         correlation_template = CorrelationTemplate(
             self.context,
-            n_ants=self.n_ants,
-            n_channels=self.n_channels_per_stream,
-            n_spectra_per_heap=self.n_spectra_per_heap,
+            n_ants=outputs[0].antennas,
+            n_channels=outputs[0].channels_per_substream,
+            n_spectra_per_heap=n_spectra_per_heap,
         )
         self.correlation = correlation_template.instantiate(
             self._proc_command_queue, n_batches=heaps_per_fengine_per_chunk
@@ -408,20 +409,20 @@ class XBEngine(DeviceServer):
         allocator = accel.DeviceAllocator(self.context)
         for _ in range(n_rx_items):
             buffer_device = self.correlation.slots["in_samples"].allocate(allocator, bind=False)
-            present = np.zeros(shape=(self.heaps_per_fengine_per_chunk, n_ants), dtype=np.uint8)
+            present = np.zeros(shape=(self.heaps_per_fengine_per_chunk, outputs[0].antennas), dtype=np.uint8)
             rx_item = RxQueueItem(buffer_device, present)
             self._rx_free_item_queue.put_nowait(rx_item)
 
         for _ in range(n_tx_items):
             buffer_device = self.correlation.slots["out_visibilities"].allocate(allocator, bind=False)
             saturated = self.correlation.slots["out_saturated"].allocate(allocator, bind=False)
-            present_ants = np.zeros(shape=(n_ants,), dtype=bool)
+            present_ants = np.zeros(shape=(outputs[0].antennas,), dtype=bool)
             tx_item = TxQueueItem(buffer_device, saturated, present_ants)
             self._tx_free_item_queue.put_nowait(tx_item)
 
         for _ in range(n_free_chunks):
             buf = self.correlation.slots["in_samples"].allocate_host(self.context)
-            present = np.zeros(n_ants * self.heaps_per_fengine_per_chunk, np.uint8)
+            present = np.zeros(outputs[0].antennas * self.heaps_per_fengine_per_chunk, np.uint8)
             chunk = recv.Chunk(data=buf, present=present, stream=self.receiver_stream)
             chunk.recycle()  # Make available to the stream
 
@@ -433,11 +434,11 @@ class XBEngine(DeviceServer):
         self.dump_interval_s: float = self.timestamp_increment_per_accumulation / adc_sample_rate_hz
 
         self.send_stream = XSend(
-            n_ants=self.n_ants,
-            n_channels=self.n_channels_total,
-            n_channels_per_stream=self.n_channels_per_stream,
+            n_ants=outputs[0].antennas,
+            n_channels=outputs[0].channels,
+            n_channels_per_stream=outputs[0].channels_per_substream,
             dump_interval_s=self.dump_interval_s,
-            send_rate_factor=self.send_rate_factor,
+            send_rate_factor=outputs[0].send_rate_factor,
             channel_offset=self.channel_offset_value,  # Arbitrary for now - depends on F-Engine stream
             context=self.context,
             packet_payload=dst_packet_payload,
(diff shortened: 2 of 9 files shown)


src/katgpucbf/xbgpu/main.py
```diff
@@ -48,13 +49,83 @@
...
+
+def parse_bengine(value: str) -> BengineOutputDict:
+    """Parse a string with B-engine configuration data.
+
+    The string has a comma-separated list of key=value pairs. See
+    :class:`BengineOutput` for the valid keys and types. The following
+    keys are required:
+
+    - beams
+    - channels_per_substream
```
bmerry, line 80: I'm not seeing how channels_per_substream can be configured for an output. We receive some number of channel as input, and the same set of channels are in the output.

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+from abc import ABC
+from dataclasses import dataclass
+
+from katsdptelstate.endpoint import Endpoint
+
+
+@dataclass
+class Output(ABC):
+    """Static configuration for an output stream."""
+
```
bmerry, line 28: You're also going to want a `name` field - might as well add it now (and parse it from the command line). Otherwise you're going to have to make a non-backwards-compatible change to the CLI later.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -282,23 +283,22 @@ def __init__(
...
         # Array configuration parameters
         self.adc_sample_rate_hz = adc_sample_rate_hz
-        self.send_rate_factor = send_rate_factor
         self.heap_accumulation_threshold = heap_accumulation_threshold
         self.time_converter = TimeConverter(sync_epoch, adc_sample_rate_hz)
-        self.n_ants = n_ants
-        self.n_channels_total = n_channels_total
-        self.n_channels_per_stream = n_channels_per_stream
-        self.n_spectra_per_heap = n_spectra_per_heap
+        self.n_ants = outputs[0].antennas  # Still needed in sender_loop
```
bmerry, line 300: There is a bit of a smell in the number of antennas being obtained from an Output, given that it's determined by the input. I'm not sure it belongs in the Output class at all, although I can see it may be convenient to have there for computing things like the `baselines` property. I'd suggest removing it from there for now and making everything else work (to ensure that the input side of things doesn't depend on having a X output) then think about adding it back.

bmerry (changes requested):

src/katgpucbf/xbgpu/main.py
```diff
@@ -48,13 +49,83 @@
...
+        type=parse_bengine,
+        default=[],
+        action="append",
+        metavar="KEY=VALUE[,KEY=VALUE...]",
+        help="Add a beamformer output (may be repeated). The required keys are: beams, dst.",
+    )
+    parser.add_argument(
+        "--ipython",
+        action="store_true",
+    )
```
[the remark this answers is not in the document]
amishatishpatel: Argh, my apologies, thought I got all of it (I was interrogating `args.beamformer` manually).

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+    channels_per_substream: int
+    dst: list[Endpoint]
+    send_rate_factor: float
+
+
+@dataclass
+class XengineOutput(Output):
```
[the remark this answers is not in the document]
amishatishpatel: Fair points all around, it did feel a bit odd when naming/using it.

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+    channels: int
+
+    @property
+    def baselines(self) -> int:
+        """Number of baselines for this array configuration."""
+        return (self.antennas + 1) * (self.antennas) * 2
+
+
+@dataclass
+class BengineOutput(Output):
```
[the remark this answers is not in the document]
amishatishpatel: Yeah this sums up my thoughts more accurately. I couldn't think of any neat shorthand/nickname for the XOutput, perhaps I'll go with BOutput for consistency with XOutput.

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+    """Static configuration for a Beamformer stream."""
+
+    beams: int
```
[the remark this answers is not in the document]
amishatishpatel: Related to a few other comments, a brief discussion had during a check-in a while back, and a general consult with Andrew Martens:
* I wasn't sure where to mention my general thought process behind having a single BengineOutput (BOutput) holding details of all beams.
* Perhaps I was inception'd by the concept of the 'uber stream' you created for Narrowband, and I was convinced I need that hammer for this nail.

I promise I didn't just entirely ignore the points you raised previously. I should've have paid closer attention to a similar point discussed with AMartens around (beam config for corr2)
> the two destination multicast groups for a beam won't be contiguous and hence can't be specified as a.b.c.d+1:7148

Appreciate the way you've (re-)worded it, as it reminds me of [James's comment on NGC-929](https://skaafrica.atlassian.net/browse/NGC-929?focusedCommentId=188765)
> [...] each beam will be its own output product.

Back to the drawing board!

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+from katsdptelstate.endpoint import Endpoint
+
+
+@dataclass
+class Output(ABC):
+    """Static configuration for an output stream."""
+
+    channels_per_substream: int
+    dst: list[Endpoint]
+    send_rate_factor: float
```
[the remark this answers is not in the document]
amishatishpatel: Ah, I see I see. I thought I gauged the difference as you explained and figured it was still useful/necessary here. Anyway, specifically
> Is there a need for this to be configured per Output?

I think not.

src/katgpucbf/xbgpu/main.py
amishatishpatel: Again, I think this points to my internally-broken telephone where, as per [this comment on the Jira ticket](https://skaafrica.atlassian.net/browse/NGC-905?focusedCommentId=191493)
* I ended up looking at the literal output data dimensions/attributes needed to describe the output.
* In hindsight, that does seem strange, those attributes would already be part of the Engine (or even eventual Pipeline [by proxy]).

src/katgpucbf/xbgpu/main.py
```diff
@@ -237,28 +318,32 @@ def make_engine(context: AbstractContext, args: argparse.Namespace) -> tuple[XBE
...
+        outputs=[xengine] + args.beamformer,
         src=args.src,
         src_interface=args.src_interface,
         src_ibv=args.src_ibv,
         src_affinity=args.src_affinity,
         src_comp_vector=args.src_comp_vector,
         src_buffer=args.src_buffer,
-        dst=args.dst,
         dst_interface=args.dst_interface,
+        dst=args.dst,
```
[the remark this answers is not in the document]
amishatishpatel: Definitely done by accident. As you mention, I had removed `args.dst` and relied solely what was specified in the `outputs`.
* Some (unqualified) internal strife led me to adding this back in.

Will remove now.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: I'm with you. As I mentioned in [an earlier reply](https://github.com/ska-sa/katgpucbf/pull/568#discussion_r1197425815), you're entirely correct Re:
> [...] I can see it may be convenient to have there for computing things like the `baselines` property.

Shall move forward with your suggestion.

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@dataclass
+class XengineOutput(Output):
+    """Static configuration for an X-engine output stream."""
+
+    antennas: int
+    channels: int
```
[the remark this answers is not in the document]
amishatishpatel: Like I mentioned in an earlier comment, I (eventually) took inspiration from looking at the literal attributes available/used on the output - in this case, `xsend.XSend`. Again, like you mentioned in [your weather report a bit earlier](https://github.com/ska-sa/katgpucbf/pull/568#discussion_r1196419470), it does seem to be a ~convenience thing.
* Even worse, looks like `channels` are only used to check if `channels_per_stream` are an integer multiple 🤦

src/katgpucbf/xbgpu/engine.py
```diff
@@ -455,15 +454,29 @@ def __init__(
...
+            - (channel_offset, channel_offset + channels_per_substream)
+        """
         # Static sensors
         sensors.add(
             aiokatcp.Sensor(
                 str,
                 "chan-range",
                 "The range of channels processed by this XB-engine, inclusive",
-                default=f"({self.channel_offset_value},{self.channel_offset_value + self.n_channels_per_stream - 1})",
+                default=f"({chan_range[0]},{chan_range[1] - 1})",
```
james-smith-za, line 479: I don't really see the reason for this change?

james-smith-za (commented):
I'm in general in support of Bruce's feedback.

I'm thinking I'd like to see something like
```
--beam=name=stephen0,dst=239.1.1.1,etc
--beam=name=stephen1,dst=239.1.1.7,etc
```
would be the incoming argument, and the number of beams being simply the number of `--beam` arguments passed.

test/xbgpu/test_main.py
```diff
@@ -19,46 +19,41 @@
...
-            parse_bengine(value)
+        with pytest.raises(ValueError, match=f"--beam: {missing} is missing"):
+            parse_beam(value)
 
     def test_duplicate_key(self) -> None:
         """Test with a key specified twice."""
-        with pytest.raises(ValueError, match="--beamformer: channels_per_substream specified twice"):
-            parse_bengine("beams=8,channels_per_substream=8,channels_per_substream=9,dst=239.1.2.3+7:7148")
+        with pytest.raises(ValueError, match="--beam: channels_per_substream specified twice"):
+            parse_beam("name=food,channels_per_substream=8,channels_per_substream=9,dst=239.1.2.3:7148")
```
james-smith-za, line 59: For completeness you might want to test with an invalid key as well.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: In my original move to make `channels_per_substream` part of an `{X, B}Output`
* I realised that the attribute would no longer be part of the `XBEngine`, and 
* therefore wouldn't be accessible as `self.<something>` in `populate_sensors`.

I figured making `populate_sensors` accept the channel range as an arg would be a straightforward solution.
* Even setting aside the `{X, B}Output` conundrum, is it ~good practice to make an arg an attribute even if it's only used once elsewhere outside the scope of `__init__`?

james-smith-za: Ok, I see the logic now.

It would break from the convention of the other `populate_sensors` functions though.

amishatishpatel: Yeah that is the concern, maybe I'll roll it back for now and earmark it for addition/update during NGC-929.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -455,8 +463,21 @@ def __init__(
...
+    ) -> None:
+        """Define the sensors for an XBEngine.
+
+        Parameters
+        ----------
+        rx_sensor_timeout
+            See :meth:`.recv.make_sensors` for more information.
+        chan_range
+            Tuple of integers showing the two values of the chan-range sensor
+            - (channel_offset, channel_offset + channels_per_substream)
```
james-smith-za, line 479: Going to need to update the docstring as well.

amishatishpatel:
I think I've addressed all the initial comments (and rolled back/refined to the simplest version of this update). Will request another round of reviewing now.

src/katgpucbf/xbgpu/main.py
```diff
@@ -48,13 +49,71 @@
...
+    """Configuration options for a beam output.
+
+    Unlike :class:`BOutput`, all the fields are optional, so that it can
+    be built up incrementally. They must all be filled in before using it to
+    construct a :class:`BOutput`.
+    """
+
+    name: str
+    dst: Endpoint
+    channels_per_substream: int
```
james-smith-za, line 67: I know @bmerry does the F-engine OutputDicts in `main.py`, but I'm wondering whether they shouldn't be in `output.py` instead, just to keep this file a bit less cluttered.

src/katgpucbf/xbgpu/main.py
```diff
@@ -217,6 +276,17 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
...
 
+    # Convert from *BOutputDict to *BOutput
+    used_names = {"xengine"}
+    for output in args.beam:
+        name = output["name"]
+        if name in used_names:
+            parser.error(f"output name {name} already used.")
+        used_names.add(name)
+        if "channels_per_substream" not in output:
+            output["channels_per_substream"] = args.channels_per_substream
```
james-smith-za, line 287: I am currently leaning towards suggesting removing this.

It shouldn't be possible for the beamformer to have a different number of channels per substream than the X-engine that it's co-located with.

src/katgpucbf/xbgpu/main.py
```diff
@@ -237,6 +307,11 @@ def make_engine(context: AbstractContext, args: argparse.Namespace) -> tuple[XBE
         monitor = NullMonitor()
 
     logger.info("Initialising XB-Engine on %s", context.device.name)
+    xengine = XOutput(
+        name="xengine",
+        dst=args.dst,
+        channels_per_substream=args.channels_per_substream,
+    )
```
james-smith-za, line 314: My sense of aesthetics almost wants the user to have to input
```
--corrprod=name=steven,dst=239.0.0.1
```
corresponding to the `--beam` argument(s). That way you bind the `dst` to the output quite nicely.

(Or some other word, I can't currently think of a better way to refer to the correlation products.)

src/katgpucbf/xbgpu/engine.py
```diff
@@ -282,23 +283,30 @@ def __init__(
...
-        if channel_offset_value % n_channels_per_stream != 0:
-            raise ValueError("channel_offset must be an integer multiple of n_channels_per_stream")
+        for output in outputs:
+            if channel_offset_value % output.channels_per_substream != 0:
+                raise ValueError(f"{output.name}: channel_offset must be an integer multiple of channels_per_substream")
+            if output.channels_per_substream != n_channels_per_stream:
+                raise ValueError(
+                    f"{output.name} has {output.channels_per_substream} channels-per-substream, \
+                                 which is not compatible with the global value ({n_channels_per_stream})"
+                )
```
james-smith-za, line 300: My other comment refers. I kind of suggest not including it in the `--beam=""` thing, because there should just be one value for the entire engine.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -433,17 +441,17 @@ def __init__(
...
-            send_rate_factor=self.send_rate_factor,
+            send_rate_factor=send_rate_factor,
             channel_offset=self.channel_offset_value,  # Arbitrary for now - depends on F-Engine stream
             context=self.context,
             packet_payload=dst_packet_payload,
             stream_factory=lambda stream_config, buffers: make_stream(
-                dest_ip=dst.host,
-                dest_port=dst.port,
+                dest_ip=outputs[0].dst.host,
+                dest_port=outputs[0].dst.port,
```
james-smith-za, line 454: @bmerry do you think it is worth protecting this somehow?

Currently `outputs` is created with the correlator output first, but that could change and this could be easy to miss. Something like
```
assert isinstance(outputs[0], XOutput)
``` 
just before the `XSend()` call.

james-smith-za (changes requested):
It's getting there. Just a couple of niggles.

src/katgpucbf/xbgpu/main.py
bmerry: The reason for keeping it in main.py is that it's internal to the implementation of argument parsing, rather than a public API. In fact I think it may be worth using an underscore prefix to emphasise that it's not part of the public API of main.py (in which case, do the same in fgpu.main).

src/katgpucbf/xbgpu/main.py
```diff
@@ -48,13 +49,71 @@
...
+    name: str
+    dst: Endpoint
+    channels_per_substream: int
+
+
+def parse_beam(value: str) -> BOutputDict:
+    """Parse a string with beam configuration data.
+
+    The string has a comma-separated list of key=value pairs. See
+    :class:`BengineOutput` for the valid keys and types. The following
```
bmerry, line 74:
```suggestion
    :class:`BOutput` for the valid keys and types. The following
```

src/katgpucbf/xbgpu/main.py
bmerry: I agree with this. In fgpu I started out by keeping backwards compatibility in the first PR and later updating it to take the wideband config in `--wideband`, but doing that involved a lot of incremental refactoring and I think it's going to be cleaner to bite the bullet now.

src/katgpucbf/xbgpu/output.py
```diff
@@ -0,0 +1,45 @@
...
+    name: str
+    dst: Endpoint
+    channels_per_substream: int
+
+
+@dataclass
+class XOutput(Output):
+    """Static configuration for an output X-engine stream."""
+
+    pass
```
bmerry, line 38: I'd suggest making the integration time (or whatever the internal representation of it is) a property of XOutput. In theory (although I can't really see a use case), you could have an engine produce two baseline-correlation-product streams with different integration times.

src/katgpucbf/xbgpu/main.py
bmerry: In theory the output could be subdivided more finely than the input, and maybe someone will want that (e.g. to parallelise a beam across more receivers). But I'm not expecting that to happen, and we can add CLI support for that later if we decide to implement it. For now I'd recommend that channels-per-substream is considered an input property, not an output property.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -282,23 +283,30 @@ def __init__(
...
-            raise ValueError("channel_offset must be an integer multiple of n_channels_per_stream")
+        for output in outputs:
+            if channel_offset_value % output.channels_per_substream != 0:
+                raise ValueError(f"{output.name}: channel_offset must be an integer multiple of channels_per_substream")
+            if output.channels_per_substream != n_channels_per_stream:
+                raise ValueError(
+                    f"{output.name} has {output.channels_per_substream} channels-per-substream, \
+                                 which is not compatible with the global value ({n_channels_per_stream})"
+                )
+            # TODO: Do we need to ensure all output.dst addresses do not overlap?
```
bmerry, line 301: I wouldn't bother. If the user wants it, they can have it. In production the addresses are generated by the computer, so you don't need to protect the user from finger trouble.

src/katgpucbf/xbgpu/engine.py
bmerry: Agreed. If channels_per_substream is removed from the output, this can be reverted.

src/katgpucbf/xbgpu/engine.py
bmerry: I wouldn't worry at this point. This code will all go through major refactoring anyway.

test/xbgpu/test_main.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_duplicate_key(self) -> None:
+        """Test with a key specified twice."""
+        with pytest.raises(ValueError, match="--beam: channels_per_substream specified twice"):
+            parse_beam("name=food,channels_per_substream=8,channels_per_substream=9,dst=239.1.2.3:7148")
+
+    def test_invalid_key(self) -> None:
+        """Test with an unknown key/value pair."""
+        with pytest.raises(ValueError, match="--beam: unknown key fizz"):
+            parse_beam("fizz=buzz,name=food,channels_per_substream=8,channels_per_substream=9,dst=239.1.2.3:7148")
```
bmerry, line 63: I'd remove the duplicate channels_per_substream, so that there is only one error with the argument.

bmerry (changes requested):
Generally happy except for the inclusion of channels_per_substream in the Output class.

It occurs to me that some of the back-and-forth is driven by tension about what the Output classes are for.  In fgpu they serve several functions: they hold configuration that comes directly from the command line, but they also function as calculators for derived output properties. That works well in fgpu because the information needed for the derived properties is already in place. For xbgpu, a lot of derived properties depend on the input configuration (e.g. number of antennas or channels-per-substream), so it doesn't work so well. Once you get to implementing the beamformer you might need/want helper classes to track all the properties of an output stream, but I think it will work better to have those be distinct from the classes that hold CLI configuration (and we can bikeshed how to name things at that point).

src/katgpucbf/xbgpu/main.py
james-smith-za: Ok, that makes sense.

In which case I'd suggest making that a separate, technical-debt ticket. @amishatishpatel, ignore this comment for the purposes of this PR.

src/katgpucbf/xbgpu/output.py
amishatishpatel: As mentioned earlier, I'm absolutely for this. Fair warning, this comment did involve some ~overthinking on my part.

In going through implementing it, and with the context of [your review summation](https://github.com/ska-sa/katgpucbf/pull/568#pullrequestreview-1452387027), I guess I should conclude
* An Output class shall cart around input config items as long as they are of use and we can agree it's useful (?)

e.g. I did away with the `antennas` attribute on `XOutput` because the calculation of the `n_baselines` property really was for convenience. 

For the record, it depends on three input config items used for `timestamp_increment_per_accumulation`
* `heap_accumulation_threshold`
* `samples_between_spectra`, and
* `spectra_per_heap`.

That seems like a lot to me, but if it's needed then who am I to judge 🤷 I'll go ahead with 
```
class XOutput(Output):
    heap_accumulation_threshold: int
    samples_between_spectra: int
    spectra_per_heap: int

    @property
    def timestamp_increment_per_accumulation(self) -> int:
        .
        .
        .
```

bmerry: Not quite what I had in mind. After reviewing the current CLI, what I meant was to add just `heap_accumulation_threshold` to the class (that is the "integration time", just specified in a units of heaps).

amishatishpatel: Oh dear, if only I'd seen this 30 seconds ago. Apologies, rolling back now.

scratch/xbgpu/run-xbgpu.sh
```diff
@@ -46,6 +46,7 @@ exec spead2_net_raw numactl -C $other_affinity xbgpu \
     --src-interface $iface \
     --dst-interface $iface \
     --src-ibv --dst-ibv \
+    --corrprod=name=bcp1,heap_accumulation_threshold=${heap_accumulation_threshold:-52},dst=$dst_mcast \
```
bmerry, line 49: Don't forget to remove the original `--heap-accumulation-threshold`.

src/katgpucbf/xbgpu/main.py
```diff
@@ -51,56 +51,119 @@
 from .correlation import device_filter
 from .output import BOutput, XOutput
 
+_OD = TypeVar("_OD", bound="OutputDict")
 logger = logging.getLogger(__name__)
+DEFAULT_HEAP_ACCUMULATION_THRESHOLD = 52
```
bmerry, line 56: I'm tempted to say let's just take the opportunity to make it required for the user to pass it, rather than making an arbitrary default that's only applicable for one use case. But I also don't want to hold up this PR for longer, so maybe it should be a ticket.

src/katgpucbf/xbgpu/main.py
```diff
@@ -51,56 +51,119 @@
...
+            case "heap_accumulation_threshold":
+                kws[key] = int(data)
+            case _:
+                raise ValueError(f"unknown key {key}")
+
+    try:
+        kws: XOutputDict = {}
+        _parse_stream(value, kws, _field_callback)
+        # TODO: Perhaps do something clever/neat and only pass XOutput the complete
+        #       timestamp_increment_per_accumulation (and not each component)
```
bmerry, line 162: Perhaps this comment is no longer relevant?

src/katgpucbf/xbgpu/main.py
```diff
@@ -51,56 +51,119 @@
...
+def parse_corrprod(value: str) -> XOutput:
+    """Parse a string with correlation product configuration data.
+
+    The string has comma-separated list of key=value pairs. See
+    :class:`XOutput` for the valid keys and types. The following keys
+    are required:
+
+    - name
+    - dst
+    - heap_accumulation_threshold
```
bmerry, line 148: Doesn't seem to be required (there is a default), although see my earlier comment suggesting that we make it required.

src/katgpucbf/xbgpu/main.py
```diff
@@ -114,6 +177,14 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         metavar="KEY=VALUE[,KEY=VALUE...]",
         help="Add a half-beam output (may be repeated). The required keys are: name, dst.",
     )
+    parser.add_argument(
+        "--corrprod",
+        type=parse_corrprod,
+        default=[],
+        action="append",
+        metavar="KEY=VALUE[,KEY=VALUE...]",
+        help="Add a baseline-correlation-product output (may be repeated). The required keys are: name, dst.",
```
bmerry, line 186:
```suggestion
        help="Add a baseline-correlation-products output (may be repeated). The required keys are: name, dst.",
```

bmerry (changes requested):
Looking pretty good.

src/katgpucbf/xbgpu/main.py
amishatishpatel: [As requested](https://skaafrica.atlassian.net/browse/NGC-988).

src/katgpucbf/xbgpu/main.py
```diff
@@ -48,13 +49,140 @@
...
+        metavar="KEY=VALUE[,KEY=VALUE...]",
+        help="Add a half-beam output (may be repeated). The required keys are: name, dst.",
+    )
+    parser.add_argument(
+        "--corrprod",
+        type=parse_corrprod,
+        default=[],
+        action="append",
+        metavar="KEY=VALUE[,KEY=VALUE...]",
+        help="Add a baseline-correlation-products output (may be repeated). The required keys are: name, dst.",
```
bmerry, line 184:
```suggestion
        help="Add a baseline-correlation-products output (may be repeated). The required keys are: name, dst, heap_accumulation_threshold.",
```

bmerry (changes requested):
Just one place you missed documenting that heap_accumulation_threshold is now required. I'm happy apart from that, and with that fixed you can assume I accept without another round of review.

amishatishpatel:
@james-smith-za let me know if I should wait for your review before merging.

james-smith-za (approved):
I like it. Some nit-picks, but they're completely stylistic. Feel free to ignore.

src/katgpucbf/xbgpu/main.py
```diff
@@ -210,13 +333,23 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
...
+    args.outputs = []
+    for output_group in [args.corrprod, args.beam]:
+        for output in output_group:
+            name = output.name
+            if name in used_names:
+                parser.error(f"output name {name} already used.")
+            used_names.add(name)
+            args.outputs.append(output)
+
+    args.beam = [BOutput(**output) for output in args.beam]
```
bmerry, line 352: I think this line is a leftover?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -282,23 +279,24 @@ def __init__(
         super().__init__(katcp_host, katcp_port)
         self._cancel_tasks: list[asyncio.Task] = []  # Tasks that need to be cancelled on shutdown
 
+        # B-engine doesn't work yet
+        assert len(outputs) == 1
+        assert isinstance(outputs[0], XOutput)
```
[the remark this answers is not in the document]
bmerry: My feeling is that assertions are fine if you believe that it can never fail, no matter what the user does, and you want to document that belief (and the computer telling you that your believe is wrong is a bonus). That way if it's eliminated by `python -O` you don't really care.

I don't think that's applicable in this case, because if the user doesn't pass `--corrprod`, this assertion will fail. So it would be a little better to replace this with an exception that explains that `--corrprod` is required. But also as James says, this is a stepping stone and we don't have to make things perfect if we know we're about to rip them out anyway.

bmerry (changes requested):
By the way, don't forget that you'll need to update katsdpcontroller to pass `--corrprod`.

amishatishpatel:
> By the way, don't forget that you'll need to update katsdpcontroller to pass `--corrprod`.

This is the kind of reminder I needed after this rollercoaster of a PR. Just kept going with the suggestions and kinda forgot the greater context of operation. As always, please and thank you!

PR #576, opened 2023-05-26
Fix qualification tests

bmerry:
I'd updated pytest-check in #575, which broke the qualification tests because pytest-check moved around some of the internals that we're monkey-patching.


bmerry:
I haven't attached a qualification report since it looks the same as before, but I've successfully run qualification manually.

james-smith-za (approved):
Odd change of API.

bmerry:
> Odd change of API.

Not sure it qualifies as "API" - I think we're poking the internals, and they got rearranged.

PR #582, opened 2023-06-01
Run wideband+narrowband in test_engine

bmerry:
Instead of only running the desired output stream (which does not test the real-world use of running both concurrently), run both and only do tests against one at a time.

The main functional change in the test is that previously --dst-chunk-jones was set differently for wide and narrow, but now they need to have a single value. I kept the value picked for narrowband just because that made all the tests pass.

Closes NGC-955.

### DIFF
--- a/test/fgpu/test_engine.py
+++ b/test/fgpu/test_engine.py
@@ -20,6 +20,7 @@ import logging
 from collections.abc import Sequence
 from dataclasses import dataclass
 from functools import partial
+from ipaddress import IPv4Network
 
 import aiokatcp
 import numpy as np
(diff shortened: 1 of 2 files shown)


test/fgpu/test_engine.py
```diff
@@ -47,18 +48,23 @@
...
-CHUNK_JONES = 1048576
+CHUNK_JONES = 131072
 MAX_DELAY_DIFF = 16384  # Needs to be lowered because CHUNK_SAMPLES is lowered
 PACKET_SAMPLES = 4096
 TAPS = 16
 FENG_ID = 42
 ADC_SAMPLE_RATE = 1712e6
 
+WIDEBAND_ARGS = f"name=test_wideband,dst=239.10.11.0+15:7149,taps={TAPS}"
+# Centre frequency is a multiple of the channel width
```
james-smith-za, line 63: Is that true regardless of the number of channels?

test/fgpu/test_engine.py
```diff
@@ -112,30 +118,31 @@ def frac_channel(output: Output, channel: float) -> float:
...
+        return f"{NARROWBAND_ARGS},channels={channels}"
+
+    @pytest.fixture(params=["wideband", "narrowband"])
+    def output(self, wideband_args: str, narrowband_args: str, request: pytest.FixtureRequest) -> Output:
+        """The output to run tests against."""
+        if request.param == "wideband":
+            return parse_wideband(wideband_args)
         else:
-            return parse_narrowband(arg_value)
+            return parse_narrowband(narrowband_args)
```
james-smith-za, line 137: So if I'm understanding correctly, this test is more or less the same as it previously was, except that
- it's now repeated twice,
- now both wide and narrowband are running, but
- only one output is checked each time.

Do I have it more or less correct?

test/fgpu/test_engine.py
bmerry: Probably not - just for the specific CHANNELS constant defined here.

test/fgpu/test_engine.py
bmerry: Yes.

james-smith-za (approved):

amishatishpatel (approved):

PR #583, opened 2023-06-01
Sort qualification test results and correlators

bmerry:
This provides a more sensible and predictor ordering for results than what naturally falls out of pytest. Only the report generation is changed - the tests themselves still run in the order pytest decides is best.

Results are split between antenna_channelised_voltage and baseline_correlation_products tests, then ordered by the correlator used. Correlators are sorted by band, channels and antennas.

Closes NGC-967.

- [x] If qualification tests are changed: attach a sample qualification report: [report-new.pdf](https://github.com/ska-sa/katgpucbf/files/11623512/report-new.pdf)

### DIFF
--- a/qualification/report/generate_pdf.py
+++ b/qualification/report/generate_pdf.py
@@ -113,6 +113,18 @@ def make_correlator_mode_str(config: dict, *, expand: bool = False) -> str:
     return config_mode
 
 
+def correlator_sort_key(config: dict) -> tuple:
+    """Generate a sort key to sort correlators."""
+    # It seems at least some of these get stringified, so ensure they are numeric
+    return (
+        float(config["bandwidth"]),
+        int(config["channels"]),
+        int(config["antennas"]),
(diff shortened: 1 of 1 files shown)


james-smith-za (approved):
I like it.

Probably not within the scope of this ticket, but now that it's sorted, I almost think it would be nice if we can have another layer of headings, like
- Detailed results
  - 4-L-4096
    - Channel Shape
    - Baselines
    - etc.

Possibly even hiding the lowest level in the table of contents.

bmerry:
> I like it.
> 
> Probably not within the scope of this ticket, but now that it's sorted, I almost think it would be nice if we can have another layer of headings, like
> 
>     * Detailed results
>       
>       * 4-L-4096
>         
>         * Channel Shape
>         * Baselines
>         * etc.
> 
> 
> Possibly even hiding the lowest level in the table of contents.

Nice idea. I've opened NGC-985 for it.

PR #584, opened 2023-06-01
Fix up initial binding of Compute slots

bmerry:
There are two issues fixed:
- Some slots had memory allocated to them before it was needed, and those slots were later re-bound, making the initial allocation unnecessary and a waste of memory (NGC-978).
- Some slots were only bound once the data arrived, slowing down the startup and possibly leading to more lost heaps than necessary.

These changes don't seem to affect the achievable throughput, but should reduce memory usage, and might reduce startup losses.

Closes NGC-978 (although I might still need to reduce the memory allocation in katsdpcontroller).

### DIFF
--- a/src/katgpucbf/fgpu/compute.py
+++ b/src/katgpucbf/fgpu/compute.py
@@ -212,8 +212,6 @@ class Compute(accel.OperationSequence):
             "gains": ["postproc:gains"],
         }
         aliases = {}
-        if template.ddc is not None:
-            compounds["ddc_weights"] = [f"ddc{pol}:weights" for pol in range(N_POLS)]
         for pol in range(N_POLS):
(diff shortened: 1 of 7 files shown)


src/katgpucbf/fgpu/compute.py
```diff
@@ -212,8 +212,6 @@ def __init__(
             "gains": ["postproc:gains"],
         }
         aliases = {}
-        if template.ddc is not None:
-            compounds["ddc_weights"] = [f"ddc{pol}:weights" for pol in range(N_POLS)]
```
bmerry, line 216: Not directly related to this PR; this is no longer useful because the DDC weights are now managed internally by the DDC class and are not slots.

james-smith-za (approved):

PR #585, opened 2023-06-02
Support ± symbol in qualification report

bmerry:
pytest.approx generates it, and lstlistings needs some help to parse it.

Closes NGC-989

### DIFF
--- a/qualification/report/preamble.tex
+++ b/qualification/report/preamble.tex
@@ -8,8 +8,8 @@
 % http://www.bollchen.de/blog/2011/04/good-looking-line-breaks-with-the-listings-package/
 \lstset{prebreak=\raisebox{0ex}[0ex][0ex]{\ensuremath{\hookleftarrow}}}
 \lstset{postbreak=\raisebox{0ex}[0ex][0ex]{\ensuremath{\hookrightarrow\space}}}
-% Make degree sign in listings (e.g. from a backtrace) work.
-\lstset{inputencoding=utf8,literate={°}{{\textdegree}}1}
+% Make degree sign and plus-minus sign in listings (e.g. from a backtrace) work.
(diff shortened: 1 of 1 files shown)


qualification/report/preamble.tex
```diff
@@ -8,8 +8,8 @@
 % http://www.bollchen.de/blog/2011/04/good-looking-line-breaks-with-the-listings-package/
 \lstset{prebreak=\raisebox{0ex}[0ex][0ex]{\ensuremath{\hookleftarrow}}}
 \lstset{postbreak=\raisebox{0ex}[0ex][0ex]{\ensuremath{\hookrightarrow\space}}}
-% Make degree sign in listings (e.g. from a backtrace) work.
-\lstset{inputencoding=utf8,literate={°}{{\textdegree}}1}
+% Make degree sign and plus-minus sign in listings (e.g. from a backtrace) work.
+\lstset{inputencoding=utf8,literate={°}{{\textdegree}}1{±}{{\ensuremath{\pm}}}1}
```
amishatishpatel, line 12: My LaTeX Fu isn't that strong, so this syntax seems odd to me. Also worth asking, does your keyboard support these characters? Or do you grab them from a resource elsewhere?

amishatishpatel (approved):

qualification/report/preamble.tex
bmerry: > My LaTeX Fu isn't that strong, so this syntax seems odd to me.

It is rather unusual for LaTeX, but seems to be what the listings package wants.

> Also worth asking, does your keyboard support these characters? Or do you grab them from a resource elsewhere?

I generally type those ones. Take a look at https://fsymbols.com/keyboard/linux/compose/.

PR #586, opened 2023-06-05
Fix phase offset for narrowband delays

bmerry:
It was failing to account for the phase shift caused by the mixer.

The mixer is now initially referenced to the start of the input chunk, rather than timestamp 0, which simplifies some calculations. When applying phase correction, the reference point is shifted to the start of the PFB window. Thus, the output phase is determined by the set of samples in the PFB window and the delay, but it is not affected by the absolute timestamp.

Some unit tests were adjusted to properly test this (when the centre frequency is not a multiple of the channel width), and a new test was added to help check my sanity.

This doesn't actually matter for baseline-correlation-products output, as the error will cancel out between the two antennas, but it will affect antenna-channelised-voltage and tied-array-channelised-voltage streams.

Qualification tests are still a work in progress.

Closes NGC-928.

### DIFF
--- a/src/katgpucbf/fgpu/compute.py
+++ b/src/katgpucbf/fgpu/compute.py
@@ -21,7 +21,6 @@ Allocations of memory for input, intermediate and output are also handled here.
 
 from collections.abc import Sequence
 from dataclasses import dataclass
-from fractions import Fraction
 
 import numpy as np
 from katsdpsigproc import accel, fft
@@ -243,14 +242,8 @@ class Compute(accel.OperationSequence):
             self.bind(**{f"in{pol}": samples[pol]})
             self.ensure_bound(f"subsampled{pol}")
         for pol in range(N_POLS):
-            # TODO: could run these in parallel, but that would require two
-            # command queues.
-            # Compute the fractional part of first_sample * mix_frequency.
-            # Using Fraction avoids the serious rounding errors that would
-            # occur using floating point.
-            phase = Fraction(self.ddc[pol].mix_frequency) * first_sample
-            phase -= round(phase)
-            self.ddc[pol].mix_phase = float(phase)
+            # TODO: eliminate this property entirely
+            self.ddc[pol].mix_phase = 0.0
             self.ddc[pol]()
 
     def _run_frontend_common(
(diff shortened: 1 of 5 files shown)


bmerry:
The more I think about this, the less clear I am on what the ideal behaviour is. I can see two possible ideologies:

1. Let's say the input is a CW with frequency f, and that the time step between output spectra is T. Then the output phase should advance by 2πfT from one spectrum to the next. That's what you'll get if this PR is merged. Or:
2. Consider just one channel bin, with centre frequency c. The effect of the PFB can be seen as shifting c to DC, applying a low-pass filter and decimating in time. Then the output phase should advance by 2π(f-c)T from one spectrum to the next. If I'm not mistaken, that's the situation prior to this PR.

@andrewvanderbyl @amartens any insight into what happens with MeerKAT? @mgouws any feelings about what *ought* to happen? It won't make any difference to correlation products (because it'll cancel out between two antennas), but I imagine it might affect any USE that does a second-stage channelisation, and possibly also VLBI.

andrewvanderbyl:
@bmerry - I don't think I'm 100% clear on exactly what you're asking but here goes. From what I can tell you are asking what point in the MK NB FEng we take as a reference when performing phase correction. In MK NB the DDS for the mixer is w.r.t the incoming timestamp (i'd have to check how many bits are sliced and used) in order to generate the correct mixing cos and sin for the complex mixing process. Phase corrections in FD are applied based on values supplied (written to registers periodically). All we do is interpolate between updates when applying. How these values are computed (or which point they are referenced to) I don't really know! @mgouws is likely to know if you are interested in how the update values are computed.

james-smith-za (approved):
One tiny bit of bike-shedding regarding the docstring, but otherwise I'm happy with the logic.

Though this is of course subject to the input from the people who know better what the expected behaviour should be.

james-smith-za (approved):

bmerry:
Feedback from @mgouws:
> Each PFB channel is equivalent to a bandpass filter + mixer to DC + decimation. If the PFB input is a tone at a channel's centre frequency, then the output is complex DC at that channel. This is true for every channel, not just the centre one.

That means that the original behaviour is the correct one, so I'm closing this PR without merging.

PR #590, opened 2023-06-07
Test narrowband modes in qualification testing

bmerry:
Add narrowband support to existing qualification tests (no narrowband-specific tests).

The config can control which narrowband modes to test via the inifile (`narrowband_decimation`); Jenkins is set up to test 1/8 and 1/16, with 32768 channels. The changes are mostly to account for either the bandwidth being different to the wideband bandwidth, the centre frequency not being the middle of the digitised band, or the centre frequency not being a multiple of the channel width.

- [x] If qualification tests are changed: attach a sample qualification report: [report-full.pdf](https://github.com/ska-sa/katgpucbf/files/11677244/report-full.pdf)

Closes NGC-984.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -189,7 +189,7 @@ async def test_delay_enable_disable(
 
 @pytest.mark.requirements("CBF-REQ-0187,CBF-REQ-0188")
 @pytest.mark.xfail(
-    reason="requirement cannot be met for all mdoes as delays cannot be updated more than once per spectrum"
+    reason="requirement cannot be met for all modes as delays cannot be updated more than once per spectrum"
 )
 async def test_delay_application_rate(correlator: CorrelatorRemoteControl, pdf_report: Reporter) -> None:
     """Test that delay and phase polynomials are applied at the required rate.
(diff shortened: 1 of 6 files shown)


qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -188,6 +188,9 @@ async def set_delays(delays: list[str]) -> None:
 
 
 @pytest.mark.requirements("CBF-REQ-0187,CBF-REQ-0188")
+@pytest.mark.xfail(
+    reason="requirement cannot be met for all modes as delays cannot be updated more than once per spectrum"
+)
```
james-smith-za, line 193: Does x-fail cause problems if some subset of the tests pass?

james-smith-za (approved):

qualification/antenna_channelised_voltage/test_delay.py
bmerry: Those are pass are reported as "xpass" when pytest summarises how many tests did what, but that counts as a pass. You can make xpass a failure by specifying `strict=True` on the mark.

I would have preferred all the xfails to be conditional on the specific modes that we know fail, but I don't see a simple way to give it a filter that has access to the test parameters.

PR #596, opened 2023-06-14
Do not return gains when setting them

bmerry:
That behaviour was due to implementing the word of the MK ICD, but the MK implementation doesn't return the gains when they are being set. Removing this should speed up gain setting.

- [x] If design has changed: ensure documentation is up to date: ICD has a proposed change
- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match: see https://github.com/ska-sa/katsdpcontroller/pull/697

Closes NGC-999.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -1585,16 +1585,16 @@ class Engine(aiokatcp.DeviceServer):
         if len(values) not in {0, 1, output.channels}:
             raise aiokatcp.FailReply(f"invalid number of values provided (must be 0, 1 or {output.channels})")
         if not values:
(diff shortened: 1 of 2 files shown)


james-smith-za (commented):
Looks good, but the thought occurs to me now to wonder whether we should include a test of the behaviour if no argument is passed, i.e. returning the existing gains.

bmerry:
> Looks good, but the thought occurs to me now to wonder whether we should include a test of the behaviour if no argument is passed, i.e. returning the existing gains.

Fair enough. Will take a look. I think that might be needed for katsdpcontroller too.

james-smith-za (approved):

PR #598, opened 2023-06-20
Update ingest script

james-smith-za:
It was written quite a while ago and relied on some assumptions that are no longer true. First among these is outdated sensor names, but also some logic from the derived classes of Chunks in katgpucbf seems to have found its way here. Which isn't how the base spead2 Chunks worked.

I'm not quite sure how it got to that state, but it's fixed now. I was able to ingest and plot spectra.

Checklist:

Closes NGC-1007


james-smith-za:
An example of spectrum plots, just for completeness:

bmerry (approved):

PR #600, opened 2023-06-28
Refactor xbgpu to support multiple output streams

amishatishpatel:
Much like #537, the result of this refactoring (and added unit testing endeavours) means the plumbing is in and (mostly?) verified - at least for the case of running multiple XPipelines.

The main change is around moving the gpu-processing and sender loops into an base Pipeline class, to be implemented by subclasses. As mentioned above, I've created an XPipeline (as per NGC-996) which contains most of the existing xbgpu codebase - just a bit reorganised and tweaked where necessary.
* A single `XBEngine` manages multiple `Pipeline`s,
* Each Pipeline corresponds to an `Output`,
* Each Pipeline also has its own processing and download command queues.

Currently only the `XPipeline` is supported (its also the only one that exists).

As mentioned in the last commit before raising this PR, `test_engine` requires a bit of cleaning up, but the structure is there. I might take further inspiration from @bmerry and opt to update relevant documentation at the end of this adventure.

As always, please and thank you.

- [ ] (N/A) If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`
- [ ] (N/A) If modules are added/removed: use sphinx-apidoc to update files in `doc/`
- [ ] (N/A) If qualification tests are changed: attach a sample qualification report

Contributes to NGC-929 and Closes NGC-996.

### DIFF
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -36,12 +36,12 @@ import time
 
 import aiokatcp
 import katsdpsigproc
-import katsdpsigproc.abc
 import katsdpsigproc.resource
 import numpy as np
 import spead2.recv
 from aiokatcp import DeviceServer
 from katsdpsigproc import accel
+from katsdpsigproc.abc import AbstractContext
 
 from .. import (
     DESCRIPTOR_TASK_NAME,
@@ -115,6 +115,349 @@ class TxQueueItem(QueueItem):
         super().reset(timestamp=timestamp)
         self.present_ants.fill(True)  # Assume they're fine until told otherwise
         self.batches = 0
+
+
+# TODO: Create a BasePipeline class, followed with {X, B}Pipelines
+
+
+class XPipeline:
+    """Processing pipeline for a single baseline-correlation-products stream."""
+
+    def __init__(self, output: Output, engine: "XBEngine", context: AbstractContext) -> None:
+        assert isinstance(output, XOutput), "Only baseline-correlation-products are supported"
+        self.engine = engine
+        self.output = output
+
+        self.timestamp_increment_per_accumulation = output.heap_accumulation_threshold * engine.rx_heap_timestamp_step
+
+        # NOTE: This value staggers the send so that packets within a heap are
+        # transmitted onto the network across the entire time between dumps.
+        # Care needs to be taken to ensure that this rate is not set too high.
+        # If it is too high, the entire pipeline will stall needlessly waiting
+        # for packets to be transmitted too slowly.
+        self.dump_interval_s = self.timestamp_increment_per_accumulation / engine.adc_sample_rate_hz
+
+        self._proc_command_queue = context.create_command_queue()
+        self._download_command_queue = context.create_command_queue()
+
+        correlation_template = CorrelationTemplate(
+            context=context,
+            n_ants=engine.n_ants,
+            n_channels=engine.n_channels_per_stream,
+            n_spectra_per_heap=engine.src_layout.n_spectra_per_heap,
+        )
+        self.correlation = correlation_template.instantiate(
+            self._proc_command_queue, n_batches=engine.src_layout.heaps_per_fengine_per_chunk
+        )
+
+        # NOTE: The n_rx_items and n_tx_items each wrap a GPU buffer. Setting
+        # these values too high results in too much GPU memory being consumed.
+        # There needs to be enough of them that the different processing
+        # functions do not get starved waiting for items. The low single digits
+        # is suitable. n_free_chunks wraps buffer in system RAM. This can be
+        # set quite high as there is much more system RAM than GPU RAM. It
+        # should be higher than max_active_chunks.
+        # These values are not configurable as they have been acceptable for
+        # most tests cases up until now. If the pipeline starts bottlenecking,
+        # then maybe look at increasing these values.
+        # TODO: Declare these constants elsewhere
+        n_rx_items = 3  # Too high means too much GPU memory gets allocated
+        n_tx_items = 2  # TODO: Will likely need to be == n_rx_items for BPipeline
+
+        # These queues are extended in the monitor class, allowing for the
+        # monitor to track the number of items on each queue.
+        # * The _rx_item_queue passes items from the _receiver_loop function to
+        #   the _gpu_proc_loop function.
+        # * The _tx_item_queue passes items from the _gpu_proc_loop to the
+        #   _sender_loop function.
+        # Once the destination function is finished with an item, it will pass
+        # it back to the corresponding _(rx/tx)_free_item_queue to ensure that
+        # all allocated buffers are in continuous circulation.
+        self._rx_item_queue: asyncio.Queue[RxQueueItem | None] = engine.monitor.make_queue("rx_item_queue", n_rx_items)
+        self._tx_item_queue: asyncio.Queue[TxQueueItem | None] = engine.monitor.make_queue("tx_item_queue", n_tx_items)
+        self._tx_free_item_queue: asyncio.Queue[TxQueueItem] = engine.monitor.make_queue(
+            "tx_free_item_queue", n_tx_items
+        )
+
+        allocator = accel.DeviceAllocator(context=context)
+        for _ in range(n_tx_items):
+            buffer_device = self.correlation.slots["out_visibilities"].allocate(allocator, bind=False)
+            saturated = self.correlation.slots["out_saturated"].allocate(allocator, bind=False)
+            present_ants = np.zeros(shape=(engine.n_ants,), dtype=bool)
+            tx_item = TxQueueItem(buffer_device, saturated, present_ants)
+            self._tx_free_item_queue.put_nowait(tx_item)
+
+    def _populate_sensors(self) -> None:
+        sensors = self.engine.sensors
+        # Static sensors
+        sensors.add(
+            aiokatcp.Sensor(
+                str,
+                "chan-range",
+                "The range of channels processed by this X-engine, inclusive",
+                default=f"({self.engine.channel_offset_value},"
+                f"{self.engine.channel_offset_value + self.engine.n_channels_per_stream - 1})",
+                initial_status=aiokatcp.Sensor.Status.NOMINAL,
+            )
+        )
+        # Dynamic sensors
+        sensors.add(
+            aiokatcp.Sensor(
+                bool,
+                "rx.synchronised",
+                "For the latest accumulation, was data present from all F-Engines.",
+                default=False,
+                initial_status=aiokatcp.Sensor.Status.ERROR,
+                status_func=lambda value: aiokatcp.Sensor.Status.NOMINAL if value else aiokatcp.Sensor.Status.ERROR,
+            )
+        )
+        sensors.add(
+            aiokatcp.Sensor(
+                int,
+                "xeng-clip-cnt",
+                "Number of visibilities that saturated",
+                default=0,
+                initial_status=aiokatcp.Sensor.Status.NOMINAL,
+            )
+        )
+
+    def add_rx_item(self, item: RxQueueItem | None) -> None:
+        """Append a newly-received :class:`RxQueueItem`."""
+        self._rx_item_queue.put_nowait(item)
+
+    async def _flush_accumulation(self, tx_item: TxQueueItem, next_accum: int) -> TxQueueItem:
+        """Emit the current `tx_item` and prepare a new one."""
+        if tx_item.batches == 0:
+            # We never actually started this accumulation. We can just
+            # update the timestamp and continue using it.
+            tx_item.timestamp = next_accum * self.timestamp_increment_per_accumulation
+            return tx_item
+
+        # present_ants only takes into account batches that have
+        # been seen. If some batches went missing entirely, the
+        # whole accumulation is bad.
+        if tx_item.batches != self.output.heap_accumulation_threshold:
+            tx_item.present_ants.fill(False)
+
+        # Update the sync sensor (converting np.bool_ to Python bool)
+        self.engine.sensors["rx.synchronised"].value = bool(tx_item.present_ants.all())
+
+        self.correlation.reduce()
+        tx_item.add_marker(self._proc_command_queue)
+        await self._tx_item_queue.put(tx_item)
+
+        # Prepare for the next accumulation (which might not be
+        # contiguous with the previous one).
+        tx_item = await self._tx_free_item_queue.get()
+        await tx_item.async_wait_for_events()
+        tx_item.timestamp = next_accum * self.timestamp_increment_per_accumulation
+        self.correlation.bind(out_visibilities=tx_item.buffer_device, out_saturated=tx_item.saturated)
+        self.correlation.zero_visibilities()
+        return tx_item
+
+    async def gpu_proc_loop(self) -> None:
+        """
+        Perform all GPU processing of received data in a continuous loop.
+
+        This function performs the following steps:
+        1. Retrieve an rx_item from the _rx_item_queue
+        2.1 Apply the correlation kernel to small subsets of the data
+            until all the data has been processed.
+        2.2 If sufficient correlations have occured, transfer the correlated
+            data to a tx_item, pass the tx_item to the _tx_item_queue and get a
+            new item from the _tx_free_item_queue.
+
+        The ratio of rx_items to tx_items is not one to one. There are expected
+        to be many more rx_items in for every tx_item out.
+
+        The above steps are performed in a loop until the running flag is set
+        to false.
+
+        .. todo::
+
+            Add B-Engine processing in this function.
+        """
+        rx_item: RxQueueItem | None
+
+        def do_correlation() -> None:
+            """Apply correlation kernel to all pending batches."""
+            first_batch = self.correlation.first_batch
+            last_batch = self.correlation.last_batch
+            if first_batch < last_batch:
+                self.correlation()
+                # Update the present ants tracker one last time
+                assert rx_item is not None
+                tx_item.present_ants[:] &= rx_item.present[first_batch:last_batch, :].all(axis=0)
+                tx_item.batches += last_batch - first_batch
+                self.correlation.first_batch = last_batch
+
+        tx_item = await self._tx_free_item_queue.get()
+        await tx_item.async_wait_for_events()
+
+        # Indicate that the timestamp still needs to be filled in.
+        tx_item.timestamp = -1
+        self.correlation.bind(out_visibilities=tx_item.buffer_device, out_saturated=tx_item.saturated)
+        self.correlation.zero_visibilities()
+        while True:
+            # Get item from the receiver function.
+            # - Wait for the HtoD transfers to complete, then
+            # - Give the chunk back to the receiver for reuse.
+            rx_item = await self._rx_item_queue.get()
+            if rx_item is None:
+                break
+            await rx_item.async_wait_for_events()
+            assert rx_item.chunk is not None  # mypy doesn't like the fact that the chunk is "optional".
+            rx_item.chunk.recycle()
+
+            current_timestamp = rx_item.timestamp
+            if tx_item.timestamp < 0:
+                # First heap seen. Round the timestamp down to the previous
+                # accumulation boundary
+                tx_item.timestamp = (
+                    current_timestamp
+                    // self.timestamp_increment_per_accumulation
+                    * self.timestamp_increment_per_accumulation
+                )
+
+            self.correlation.bind(in_samples=rx_item.buffer_device)
+            # Initially no work to do; as each batch is examined, last_batch
+            # is extended.
+            self.correlation.first_batch = 0
+            self.correlation.last_batch = 0
+            for i in range(self.engine.heaps_per_fengine_per_chunk):
+                # NOTE: The timestamp representing the end of an
+                # accumulation does not necessarily line up with the chunk
+                # timestamp. It will line up with a specific batch within a
+                # chunk though, this is why this check has to happen for each
+                # batch. This check is the equivalent of the MeerKAT SKARAB
+                # X-Engine auto-resync logic.
+                current_accum = current_timestamp // self.timestamp_increment_per_accumulation
+                tx_accum = tx_item.timestamp // self.timestamp_increment_per_accumulation
+                if current_accum != tx_accum:
+                    do_correlation()
+                    tx_item = await self._flush_accumulation(tx_item, current_accum)
+                self.correlation.last_batch = i + 1
+                current_timestamp += self.engine.rx_heap_timestamp_step
+
+            do_correlation()
+            # If the last batch of the chunk was also the last batch of the
+            # accumulation, we can flush it now without waiting for more data.
+            # This is mostly a convenience for unit tests, since in practice
+            # we'd expect to see more data soon.
+            current_accum = current_timestamp // self.timestamp_increment_per_accumulation
+            tx_accum = tx_item.timestamp // self.timestamp_increment_per_accumulation
+            if current_accum != tx_accum:
+                tx_item = await self._flush_accumulation(tx_item, current_accum)
+
+            rx_item.reset()
+            rx_item.add_marker(self._proc_command_queue)
+            self.engine.free_rx_item(rx_item)
+        # When the stream is closed, if the sender loop is waiting for a tx item,
+        # it will never exit. Upon receiving this NoneType, the sender_loop can
+        # stop waiting and exit.
+        logger.debug("gpu_proc_loop completed")
+        self._tx_item_queue.put_nowait(None)
+
+    async def sender_loop(self, send_stream: XSend) -> None:
+        """
+        Send heaps to the network in a continuous loop.
+
+        This function does the following:
+        1. Get an item from the _tx_item_queue.
+        2. Wait for all the events on this item to complete.
+        3. Wait for an available heap buffer from the send_stream.
+        4. Transfer the GPU buffer in the item to the heap buffer in system RAM.
+        5. Wait for the transfer to complete.
+        6. Transmit data in heap buffer out into the network.
+        7. Place the tx_item on _tx_item_free_queue so that it can be reused.
+
+        The above steps are performed in a loop until the running flag is set to false.
+
+        NOTE: The transfer from the GPU to the heap buffer and the sending onto
+        the network could be pipeline a bit better, but this is not really
+        required in this loop as this whole process occurs at a much slower
+        pace than the rest of the pipeline.
+        """
+        old_time_s = time.time()
+        old_timestamp = 0
+
+        while True:
+            item = await self._tx_item_queue.get()
+            if item is None:
+                break
+            await item.async_wait_for_events()
+
+            heap = await send_stream.get_free_heap()
+
+            # NOTE: We do not expect the time between dumps to be the same each
+            # time as the time.time() function checks the wall time now, not
+            # the actual time between timestamps. The difference between dump
+            # timestamps is expected to be constant.
+            new_time_s = time.time()
+            time_difference_between_heaps_s = new_time_s - old_time_s
+
+            logger.debug(
+                "Current output heap timestamp: %#x, difference between timestamps: %#x, "
+                "wall time between dumps %.2f s",
+                item.timestamp,
+                item.timestamp - old_timestamp,
+                time_difference_between_heaps_s,
+            )
+
+            # NOTE: As the output packets are rate limited in
+            # such a way to match the dump rate, receiving data too quickly
+            # will result in data bottlenecking at the sender, the pipeline
+            # eventually stalling and the input buffer overflowing.
+            if time_difference_between_heaps_s * 1.05 < self.dump_interval_s:
+                logger.warning(
+                    "Time between output heaps: %.2f which is less the expected %.2f. "
+                    "If this warning occurs too often, the pipeline will stall "
+                    "because the rate limited sender will not keep up with the input rate.",
+                    time_difference_between_heaps_s,
+                    self.dump_interval_s,
+                )
+
+            old_time_s = new_time_s
+            old_timestamp = item.timestamp
+
+            item.buffer_device.get_async(self._download_command_queue, heap.buffer)
+            item.saturated.get_async(self._download_command_queue, heap.saturated)
+            event = self._download_command_queue.enqueue_marker()
+            await katsdpsigproc.resource.async_wait_for_events([event])
+
+            if not np.any(item.present_ants):
+                # All Antennas have missed data at some point, mark the entire dump missing
+                logger.warning("All Antennas had a break in data during this accumulation")
+                heap.buffer[...] = MISSING
+                incomplete_accum_counter.inc(1)
+            elif not item.present_ants.all():
+                affected_baselines = Correlation.get_baselines_for_missing_ants(item.present_ants, self.engine.n_ants)
+                for affected_baseline in affected_baselines:
+                    # Multiply by four as each baseline (antenna pair) has four
+                    # associated correlation components (polarisation pairs).
+                    affected_baseline_index = affected_baseline * 4
+                    heap.buffer[:, affected_baseline_index : affected_baseline_index + 4, :] = MISSING
+
+                incomplete_accum_counter.inc(1)
+            # else: No F-Engines had a break in data for this accumulation
+
+            heap.timestamp = item.timestamp
+            if send_stream.tx_enabled:
+                # Convert timestamp for the *end* of the heap (not the start)
+                # to a UNIX time for the sensor update. NB: this should be done
+                # *before* send_heap, because that gives away ownership of the
+                # heap.
+                end_adc_timestamp = item.timestamp + self.timestamp_increment_per_accumulation
+                end_timestamp = self.engine.time_converter.adc_to_unix(end_adc_timestamp)
+                clip_cnt_sensor = self.engine.sensors["xeng-clip-cnt"]
+                clip_cnt_sensor.set_value(clip_cnt_sensor.value + int(heap.saturated), timestamp=end_timestamp)
+            send_stream.send_heap(heap)
+
+            item.reset()
+            await self._tx_free_item_queue.put(item)
+
+        await send_stream.send_stop_heap()
+        logger.debug("sender_loop completed")
 
 
 class XBEngine(DeviceServer):
@@ -274,7 +617,7 @@ class XBEngine(DeviceServer):
         rx_reorder_tol: int,
         tx_enabled: bool,
         monitor: Monitor,
-        context: katsdpsigproc.abc.AbstractContext,
+        context: AbstractContext,
     ):
         super().__init__(katcp_host, katcp_port)
         self._cancel_tasks: list[asyncio.Task] = []  # Tasks that need to be cancelled on shutdown
@@ -305,7 +648,17 @@ class XBEngine(DeviceServer):
         self._src_buffer = src_buffer
         self._src_comp_vector = src_comp_vector
 
+        self._dst_interface = dst_interface
+        self._dst_ttl = dst_ttl
+        self._dst_ibv = dst_ibv
+        self._dst_packet_payload = dst_packet_payload
+        self._dst_affinity = dst_affinity
+        self._dst_comp_vector = dst_comp_vector
+        self._send_rate_factor = send_rate_factor
+
         self._init_tx_enabled = tx_enabled
+
+        self._pipelines = [XPipeline(output, self, context) for output in outputs]
 
         # NOTE: The n_rx_items and n_tx_items each wrap a GPU buffer. Setting
         # these values too high results in too much GPU memory being consumed.
@@ -318,7 +671,6 @@ class XBEngine(DeviceServer):
         # most tests cases up until now. If the pipeline starts bottlenecking,
         # then maybe look at increasing these values.
         n_rx_items = 3  # Too high means too much GPU memory gets allocated
-        n_tx_items = 2
 
         # Multiply this _step by 2 to account for dropping half of the
         # spectrum due to symmetric properties of the Fourier Transform. While
@@ -326,6 +678,7 @@ class XBEngine(DeviceServer):
         # configure the receiver, we pass it as a seperate argument to the
         # reciever for cases where the n_channels_per_stream changes across
         # streams (likely for non-power-of-two array sizes).
+        # NOTE: Should be the same for both X- and B-engines, right?
         self.rx_heap_timestamp_step = n_samples_between_spectra * n_spectra_per_heap
 
         self.populate_sensors(
@@ -354,7 +707,7 @@ class XBEngine(DeviceServer):
             self.max_active_chunks, name="recv_data_ringbuffer", task_name=RECV_TASK_NAME, monitor=monitor
         )
         free_ringbuffer = spead2.recv.ChunkRingbuffer(n_free_chunks)
-        self._src_layout = recv.Layout(
+        self.src_layout = recv.Layout(
             n_ants=n_ants,
             n_channels_per_stream=n_channels_per_stream,
             n_spectra_per_heap=n_spectra_per_heap,
@@ -363,7 +716,7 @@ class XBEngine(DeviceServer):
             heaps_per_fengine_per_chunk=self.heaps_per_fengine_per_chunk,
         )
         self.receiver_stream = recv.make_stream(
-            layout=self._src_layout,
+            layout=self.src_layout,
             data_ringbuffer=data_ringbuffer,
             free_ringbuffer=free_ringbuffer,
             src_affinity=src_affinity,
@@ -376,18 +729,6 @@ class XBEngine(DeviceServer):
         # command queue can either be implemented as an OpenCL command queue or
         # a CUDA stream depending on the context.
         self._upload_command_queue = self.context.create_command_queue()
-        self._proc_command_queue = self.context.create_command_queue()
-        self._download_command_queue = self.context.create_command_queue()
-
-        correlation_template = CorrelationTemplate(
-            self.context,
-            n_ants=n_ants,
-            n_channels=n_channels_per_stream,
-            n_spectra_per_heap=n_spectra_per_heap,
-        )
-        self.correlation = correlation_template.instantiate(
-            self._proc_command_queue, n_batches=heaps_per_fengine_per_chunk
-        )
 
         # These queues are extended in the monitor class, allowing for the
         # monitor to track the number of items on each queue.
@@ -400,25 +741,16 @@ class XBEngine(DeviceServer):
         # all allocated buffers are in continuous circulation.
         self._rx_item_queue: asyncio.Queue[RxQueueItem | None] = self.monitor.make_queue("rx_item_queue", n_rx_items)
         self._rx_free_item_queue: asyncio.Queue[RxQueueItem] = self.monitor.make_queue("rx_free_item_queue", n_rx_items)
-        self._tx_item_queue: asyncio.Queue[TxQueueItem | None] = self.monitor.make_queue("tx_item_queue", n_tx_items)
-        self._tx_free_item_queue: asyncio.Queue[TxQueueItem] = self.monitor.make_queue("tx_free_item_queue", n_tx_items)
 
-        allocator = accel.DeviceAllocator(self.context)
+        rx_data_shape = (heaps_per_fengine_per_chunk, n_ants, n_channels_total, n_spectra_per_heap)
         for _ in range(n_rx_items):
-            buffer_device = self.correlation.slots["in_samples"].allocate(allocator, bind=False)
+            buffer_device = accel.DeviceArray(context, rx_data_shape, dtype=np.int8)
             present = np.zeros(shape=(self.heaps_per_fengine_per_chunk, n_ants), dtype=np.uint8)
             rx_item = RxQueueItem(buffer_device, present)
             self._rx_free_item_queue.put_nowait(rx_item)
 
-        for _ in range(n_tx_items):
-            buffer_device = self.correlation.slots["out_visibilities"].allocate(allocator, bind=False)
-            saturated = self.correlation.slots["out_saturated"].allocate(allocator, bind=False)
-            present_ants = np.zeros(shape=(n_ants,), dtype=bool)
-            tx_item = TxQueueItem(buffer_device, saturated, present_ants)
-            self._tx_free_item_queue.put_nowait(tx_item)
-
         for _ in range(n_free_chunks):
-            buf = self.correlation.slots["in_samples"].allocate_host(self.context)
+            buf = buffer_device.empty_like()
             present = np.zeros(n_ants * self.heaps_per_fengine_per_chunk, np.uint8)
             chunk = recv.Chunk(data=buf, present=present, stream=self.receiver_stream)
             chunk.recycle()  # Make available to the stream
@@ -455,37 +787,7 @@ class XBEngine(DeviceServer):
 
     def populate_sensors(self, sensors: aiokatcp.SensorSet, rx_sensor_timeout: float) -> None:
         """Define the sensors for an XBEngine."""
-        # Static sensors
-        sensors.add(
-            aiokatcp.Sensor(
-                str,
-                "chan-range",
-                "The range of channels processed by this XB-engine, inclusive",
-                default=f"({self.channel_offset_value},{self.channel_offset_value + self.n_channels_per_stream - 1})",
-                initial_status=aiokatcp.Sensor.Status.NOMINAL,
-            )
-        )
         # Dynamic sensors
-        sensors.add(
-            aiokatcp.Sensor(
-                bool,
-                "rx.synchronised",
-                "For the latest accumulation, was data present from all F-Engines.",
-                default=False,
-                initial_status=aiokatcp.Sensor.Status.ERROR,
-                status_func=lambda value: aiokatcp.Sensor.Status.NOMINAL if value else aiokatcp.Sensor.Status.ERROR,
-            )
-        )
-        sensors.add(
-            aiokatcp.Sensor(
-                int,
-                "xeng-clip-cnt",
-                "Number of visibilities that saturated",
-                default=0,
-                initial_status=aiokatcp.Sensor.Status.NOMINAL,
-            )
-        )
-
         for sensor in recv.make_sensors(rx_sensor_timeout).values():
             sensors.add(sensor)
 
@@ -494,6 +796,18 @@ class XBEngine(DeviceServer):
         time_sync_task = add_time_sync_sensors(sensors)
         self.add_service_task(time_sync_task)
         self._cancel_tasks.append(time_sync_task)
+
+    def free_rx_item(self, item: RxQueueItem) -> None:
+        """Return an :class:`RxQueueItem` to the free queue."""
+        self._rx_free_item_queue.put_nowait(item)
+
+    def _add_rx_item(self, item: RxQueueItem | None) -> None:
+        """Push an :class:`RxQueueItem` to all the pipelines.
+
+        Allow for None to be passed through to indicate shutdown procedure.
+        """
+        for pipeline in self._pipelines:
+            pipeline.add_rx_item(item)
 
     async def _receiver_loop(self) -> None:
         """
@@ -510,7 +824,7 @@ class XBEngine(DeviceServer):
         """
         async for chunk in recv.recv_chunks(
             self.receiver_stream,
-            self._src_layout,
+            self.src_layout,
             self.sensors,
             self.time_converter,
         ):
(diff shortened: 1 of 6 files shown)


src/katgpucbf/xbgpu/engine.py
```diff
@@ -318,14 +749,14 @@ def __init__(
...
         n_rx_items = 3  # Too high means too much GPU memory gets allocated
-        n_tx_items = 2
 
         # Multiply this _step by 2 to account for dropping half of the
         # spectrum due to symmetric properties of the Fourier Transform. While
         # we can workout the timestamp_step from other parameters that
         # configure the receiver, we pass it as a seperate argument to the
         # reciever for cases where the n_channels_per_stream changes across
         # streams (likely for non-power-of-two array sizes).
+        # NOTE: Should be the same for both X- and B-engines, right?
```
james-smith-za, line 759: The thought occurs to me to wonder whether this comment actually relates to the code anymore. It may be worth digging back into the history. There's no 2 here, was there at one point?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -396,96 +817,34 @@ def __init__(
...
+            heaps_per_fengine_per_chunk,
+            n_ants,
+            n_channels_per_stream,
+            n_spectra_per_heap,
+            N_POLS,
+            COMPLEX,
         )
+        for _ in range(n_rx_items):
+            # TODO: Abstract dtype, as per NGC-1000
+            buffer_device = accel.DeviceArray(context, rx_data_shape, dtype=np.int8)
```
james-smith-za, line 834: You've changed the way that the `buffer_device` is assigned here. Why?

Do you want to avoid reaching into one of the Pipeline objects and using the slots' `allocate_host` function?

james-smith-za (changes requested):
I am broadly happy with the approach taken, but I think we need a little bit of tidying up.

There might be one or two points of discussion here, perhaps we can do a zoom at some point to clarify things, but for the time being you should have some things to work on.

src/katgpucbf/xbgpu/engine.py
bmerry: And yes, n_channels_per_stream will be the same for all streams because it's determined by the input that the engine subscribes to (and non-power-of-two array sizes have nothing to do with it).

src/katgpucbf/xbgpu/engine.py
```diff
@@ -396,96 +817,34 @@ def __init__(
         # * The _tx_item_queue passes items from the _gpu_proc_loop to the
         #   _sender_loop function.
```
bmerry, line 818: There is no tx_item_queue here any more.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -396,96 +817,34 @@ def __init__(
...
-                interface_ip=dst_interface,
-                ttl=dst_ttl,
-                use_ibv=dst_ibv,
-                affinity=dst_affinity,
-                comp_vector=dst_comp_vector,
-                stream_config=stream_config,
-                buffers=buffers,
-            ),
-            tx_enabled=self._init_tx_enabled,
+        # it back to the corresponding _(rx/tx)_free_item_queue to ensure that
```
bmerry, line 820: No tx_free_item_queue here any more.

src/katgpucbf/xbgpu/engine.py
bmerry: I'm happy with this change. Once we have B pipelines, it will be more difficult to generically access the appropriate slots within the pipelines.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -495,6 +854,39 @@ def populate_sensors(self, sensors: aiokatcp.SensorSet, rx_sensor_timeout: float
...
+    async def _push_recv_chunk(item: RxQueueItem) -> None:
+        """Return chunk to the stream once `events` have fired."""
+        await item.async_wait_for_events()
+        item.chunk.recycle()  # type: ignore
+        item.chunk = None
+
+    async def _add_rx_item(self, item: RxQueueItem) -> None:
+        """Push an :class:`RxQueueItem` to all the pipelines.
+
+        Allow for None to be passed through to indicate shutdown procedure.
```
bmerry, line 883: Is this meant to be a TODO? It doesn't match the type annotation or the body of the function.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -528,254 +920,45 @@ async def _receiver_loop(self) -> None:
...
+    async def request_capture_start(self, ctx, stream_name: str) -> None:
+        """Start transmission of a stream.
 
-        The above steps are performed in a loop until the running flag is set to false.
-
-        NOTE: The transfer from the GPU to the heap buffer and the sending onto
-        the network could be pipeline a bit better, but this is not really
-        required in this loop as this whole process occurs at a much slower
-        pace than the rest of the pipeline.
+        The request requires a stream name to enable
```
bmerry, line 949: The generated help for katcp requests doesn't show the function signature, so it's recommended to use a `Parameters` section to formally document the parameter.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -528,254 +920,45 @@ async def _receiver_loop(self) -> None:
...
-    async def request_capture_stop(self, ctx) -> None:
-        """Stop transmission of this baseline-correlation-products stream."""
-        self.send_stream.tx_enabled = False
+        pipeline = self._request_pipeline(stream_name)
+        # TODO: Introduce `make_send_stream` helper function for each pipeline
+        #       to address this 'has no attribute' mypy error
+        pipeline.send_stream.tx_enabled = True  # type: ignore
+
+    async def request_capture_stop(self, ctx, stream_name: str) -> None:
+        """Stop transmission of a stream."""
```
bmerry, line 957: Again, document the parameter.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -528,254 +920,45 @@ async def _receiver_loop(self) -> None:
...
-    async def request_capture_start(self, ctx) -> None:
-        """Start transmission of this baseline-correlation-products stream."""
-        self.send_stream.tx_enabled = True
-
-    async def request_capture_stop(self, ctx) -> None:
-        """Stop transmission of this baseline-correlation-products stream."""
-        self.send_stream.tx_enabled = False
+        pipeline = self._request_pipeline(stream_name)
+        # TODO: Introduce `make_send_stream` helper function for each pipeline
+        #       to address this 'has no attribute' mypy error
```
bmerry, line 953: Not quite sure I follow. But rather than exposing the XSend object or its B equivalent directly to the Engine, what about have the base class provide an abstract function to enable or disable transmission?

bmerry (changes requested):
I've looked over the src/ tree so far. As James said, the overall structure seems sound, but a few things need fixing and there is a bit of polishing.

I'll come back for the test/ tree tomorrow, so you can leave the ticket assigned to me for now.

test/xbgpu/test_engine.py
```diff
@@ -506,46 +523,49 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+
+        assert (
+            prom_diff.get_sample_diff("output_x_incomplete_accs_total", {"stream": "bcp1"}) == incomplete_accums_counter
+        )
+        assert prom_diff.get_sample_diff("output_x_heaps_total", {"stream": "bcp1"}) == n_total_accumulations
         # Could manually calculate it here, but it's available inside the send_stream
-        assert prom_diff.get_sample_diff("output_x_bytes_total") == (
-            xbengine.send_stream.heap_payload_size_bytes * n_total_accumulations
+        assert prom_diff.get_sample_diff("output_x_bytes_total", {"stream": "bcp1"}) == (
+            xbengine._pipelines[0].send_stream.heap_payload_size_bytes * n_total_accumulations
```
bmerry, line 563: Could use `_request_pipeline` to do the lookup by name, to ensure that you're querying the right pipeline i.e. the one matching `{"stream": "bcp1"}`.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -23,7 +23,7 @@
 
 .. todo::
 
-    - The B-Engine logic has not been implemented yet - this needs to be added
+    * The B-Engine logic has not been implemented yet - this needs to be added
```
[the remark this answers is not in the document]
amishatishpatel: Apologies, I had actually changed them in an attempt to resolve some doc building errors (which still exist).

Will roll back to `-` convention.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+        """Perform all GPU processing of received data in a continuous loop.
+
+        This function should do the following:
+        * Get an RxQueueItem off the rx_item_queue
+        * Ensure it is not a NoneType value (indicating shutdown sequence)
+        * await any outstanding events associated with the RxQueueItem
+        * TODO: Figure out whether the `is not None` check on the Chunk is necessary
+        * Apply GPU processing to data in the RxQueueItem
+            * Bind input buffer(s) accordingly
+        * Obtain a free TxQueueItem of the tx_free_item_queue
```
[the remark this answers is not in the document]
amishatishpatel: Yeah, I likely meant **off** the queue, but **from** the queue sounds better.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+        2.1 Apply the correlation kernel to small subsets of the data
+            until all the data has been processed.
+        2.2 If sufficient correlations have occured, transfer the correlated
+            data to a tx_item, pass the tx_item to the _tx_item_queue and get a
+            new item from the _tx_free_item_queue.
+
+        The ratio of rx_items to tx_items is not one to one. There are expected
+        to be many more rx_items in for every tx_item out.
+        """
+        rx_item: RxQueueItem | None
```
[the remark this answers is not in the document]
amishatishpatel: Trust me, I didn't enjoy adding it either 😅 Something in the pre-commit was complaining it didn't know what `rx_item` is below/in the helper function.

The removal results in "Cannot determine type of `rx_item`"
* Maybe I'll add a `type: ignore` below?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+        if tx_item.batches == 0:
+            # We never actually started this accumulation. We can just
+            # update the timestamp and continue using it.
+            tx_item.timestamp = next_accum * self.timestamp_increment_per_accumulation
+            return tx_item
+
+        # present_ants only takes into account batches that have
+        # been seen. If some batches went missing entirely, the
+        # whole accumulation is bad.
+        if tx_item.batches != self.output.heap_accumulation_threshold:  # type: ignore
```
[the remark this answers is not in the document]
amishatishpatel: I added it here because mypy complained
> "Output" has no attribute "heap_accumulation_threshold"

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+        while True:
+            # Get item from the receiver function.
+            # * Wait for the HtoD transfers to complete, then
+            # * Give the chunk back to the receiver for reuse.
+            rx_item = await self._rx_item_queue.get()
+            if rx_item is None:
+                break
+            await rx_item.async_wait_for_events()
+            # TODO: Get rid of this explicit reference to the Chunk
+            assert rx_item.chunk is not None  # mypy doesn't like the fact that the chunk is "optional".
```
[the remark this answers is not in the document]
amishatishpatel: Too right, too right. Thanks for the heads-up.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+
+            rx_item.add_marker(self._proc_command_queue)
+            self.engine.free_rx_item(rx_item)
+        # When the stream is closed, if the sender loop is waiting for a tx item,
+        # it will never exit. Upon receiving this NoneType, the sender_loop can
+        # stop waiting and exit.
+        logger.debug("gpu_proc_loop completed")
+        self._tx_item_queue.put_nowait(None)
+
+    async def sender_loop(self) -> None:  # noqa: D102
```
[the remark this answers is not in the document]
amishatishpatel: It does not, I'd added it in my debugging of why <something> was complaining that I hadn't implemented the @abstractmethod properly. I fixed that but failed to remove my "throw everything at the problem to fix it" - removing now.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -495,6 +854,39 @@ def populate_sensors(self, sensors: aiokatcp.SensorSet, rx_sensor_timeout: float
...
+                    self._push_recv_chunk(item),
+                    name="Receive Chunk Recycle Task",
+                )
+            )
+            item.reset()
+            self._rx_free_item_queue.put_nowait(item)
+
+    @staticmethod
+    async def _push_recv_chunk(item: RxQueueItem) -> None:
+        """Return chunk to the stream once `events` have fired."""
```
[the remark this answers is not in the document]
amishatishpatel: I'd left it in to allude to the `await` below it. I gather this backtick-manner of mentioning things is more specific to function args.

Will update accordingly (to e.g. "once an item's events have fired").

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Ah, legacy from my own reminder (now done with the `shutdown`). Will remove.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
bmerry: While it's not pretty, having mypy know the types of things is definitely preferable to sprinkling on `type: ignore`.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
bmerry: That's probably because `self.output` is defined in the base class as `Output`, so it doesn't know that it's actually an XOutput. That problem should vanish once you remove `output` from the base class constructor and handle it in XPipeline.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Rather embarassingly, I didn't know you could `?help <request_name>` like you can with ipython `thing.function?`.

Updating now!

src/katgpucbf/xbgpu/engine.py
amishatishpatel: mypy, as always, is upset because it doesn't know about the `send_stream` attribute that we're accessing below.

Your suggestion is a good one indeed, and probably a bit cleaner than reaching straight into a `pipeline` and toggling a value.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -495,6 +854,39 @@ def populate_sensors(self, sensors: aiokatcp.SensorSet, rx_sensor_timeout: float
         self.add_service_task(time_sync_task)
         self._cancel_tasks.append(time_sync_task)
 
+    def free_rx_item(self, item: RxQueueItem) -> None:
+        """Return an RxQueueItem to the free queue if its refcount hits zero."""
+        item.refcount -= 1
+        if item.refcount == 0:
+            # All Pipelines are done with this item
+            self._active_in_sem.release()
+            # Need to wait for any outstanding events before recycling
```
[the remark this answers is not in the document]
amishatishpatel: As usual, fair points all around. I don't like answering a question with another question, _but_ I think it may be an issue because
* Does the item still have an event to a/wait from `rx_item.add_marker(self._proc_command_queue)`?
* It's called/invoked just before the call to `free_rx_item` in the gpu_proc_loop.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -495,6 +854,39 @@ def populate_sensors(self, sensors: aiokatcp.SensorSet, rx_sensor_timeout: float
...
+            )
+            item.reset()
+            self._rx_free_item_queue.put_nowait(item)
+
+    @staticmethod
+    async def _push_recv_chunk(item: RxQueueItem) -> None:
+        """Return chunk to the stream once `events` have fired."""
+        await item.async_wait_for_events()
+        item.chunk.recycle()  # type: ignore
+        item.chunk = None
```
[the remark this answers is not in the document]
amishatishpatel: Ooff, good catch. I think I was being lazy in giving it the entire RxQueueItem. Perhaps I was also naive in ending `free_rx_item` with a `put_nowait`.
* I'm going to split the difference between this comment and the [earlier one](https://github.com/ska-sa/katgpucbf/pull/600/files#r1246647148) (in free_rx_item) for the next round of reviewing.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
bmerry: That captures the GPU processing done on the RxItem (the GPU memory), but you don't need to wait for that to reuse the chunk (CPU memory). Once you've copied the chunk to the GPU, the chunk is ready for reuse.

test/xbgpu/test_engine.py
```diff
@@ -365,31 +384,21 @@ def n_samples_between_spectra(self, n_channels_total: int) -> int:  # noqa: D102
         # Will need to be updated for narrowband
```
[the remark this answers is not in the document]
amishatishpatel: Too right, I've created [NGC-1014](https://skaafrica.atlassian.net/browse/NGC-1014).

test/xbgpu/test_engine.py
```diff
@@ -404,6 +413,15 @@ async def xbengine(
...
         ]
+
+    @pytest.fixture
+    async def xbengine(
+        self,
+        context: AbstractContext,
+        engine_arglist: list[str],
+    ) -> AsyncGenerator[XBEngine, None]:
+        """Create and start an engine based on the fixture values."""
+        arglist = list(engine_arglist)  # Copy, to ensure we don't alter the original
```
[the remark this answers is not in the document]
amishatishpatel: > Do you have any reason to think you might be altering the original simply by parsing it?

Honestly? No. Your comment in fgpu/test made sense to me (ethically, morally), so I figured might as well employ the same practice here.

test/xbgpu/test_engine.py
```diff
@@ -506,46 +523,49 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
-                # checked here is simply the first in the range.
-                # - Even though :meth:`generate_expected_output` returns a
-                #   zeroed array for an incomplete accumulation, we still need
-                #   to maintain programmatic sense in the values generated here.
-                num_batches_in_current_accumulation = HEAPS_PER_FENGINE_PER_CHUNK
-                incomplete_accums_counter += 1
-            else:
-                num_batches_in_current_accumulation = heap_accumulation_threshold
-                if missing_antenna is not None:
+        for i in range(len(mock_send_stream)):
```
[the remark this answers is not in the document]
amishatishpatel: Indeed indeed. As mentioned during the consult yesterday, this wasn't quite a complete representation of the required `test_engine` updates.

amishatishpatel:
Alrighty, I think I've addressed all the comments under `src/`. Bruce's comments/observations under `test/` are, as always, entirely valid, but will take some time to implement (mainly around testing different `heap_accumulation_threshold`s for each `--corrprod`).

With that in mind, would you both (@james-smith-za and @bmerry ) be happy to review the `src/` updates while I continue working on `test_engine` et al?
* I'm going to request a re-review anyway,
* As usual, ignore until you see fit to interrogate.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+        # Prepare for the next accumulation (which might not be
+        # contiguous with the previous one).
+        tx_item = await self._tx_free_item_queue.get()
+        await tx_item.async_wait_for_events()
+        tx_item.timestamp = next_accum * self.timestamp_increment_per_accumulation
+        self.correlation.bind(out_visibilities=tx_item.buffer_device, out_saturated=tx_item.saturated)
+        self.correlation.zero_visibilities()
+        return tx_item
+
+    async def gpu_proc_loop(self) -> None:  # noqa: D102
```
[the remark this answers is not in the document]
james-smith-za: Ok so you haven't changed the `noqa` here but I suspect that's because you moved the docstring to the base class. Which is probably fine, but then this function will get nothing of its own in the module documentation.
That may be OK, I haven't considered the implications too deeply.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
james-smith-za: I'm surprised that mypy can't figure it out because it knows what the type of `_rx_item_queue` is, so it should know what kind of thing `get()` produces.

Would it not be possible to add the type hint where `rx_item` is below? like
```
rx_item: RxQueueItem | None = await self._rx_item_queue.get()
```

I didn't test that at all, this is just how I would have done it. What are the down-sides to what I've suggested compared with what you've got here?

(This is a really minor item. Ignore if you even have the slightest inkling of not wanting to bother.)

test/xbgpu/test_engine.py
```diff
@@ -177,6 +178,14 @@ def valid_end_to_end_combination(combo: dict) -> bool:
 class TestEngine:
     r"""Grouping of unit tests for :class:`.XBEngine`\'s various functionality."""
 
+    @pytest.fixture
+    def mock_send_stream_network(self) -> IPv4Network:
+        """Filter the output queues to just those corresponding to the --corrprods.
```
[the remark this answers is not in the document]
james-smith-za: @amishatishpatel, did you address this comment at all, perhaps during your real-time feedback session?

test/xbgpu/test_engine.py
```diff
@@ -177,6 +178,14 @@ def valid_end_to_end_combination(combo: dict) -> bool:
...
+
+        Hardcoded to override the default implementation in conftest.py.
```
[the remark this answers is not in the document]
james-smith-za: ditto here

src/katgpucbf/xbgpu/engine.py
```diff
@@ -305,27 +721,25 @@ def __init__(
...
-        # is suitable. n_free_chunks wraps buffer in system RAM. This can be
-        # set quite high as there is much more system RAM than GPU RAM. It
-        # should be higher than max_active_chunks.
+        # NOTE: n_free_chunks dictates the number of buffers in system RAM.
+        # This can be set quite high as there is much more system RAM than GPU
+        # RAM. It should be higher than max_active_chunks.
         # These values are not configurable as they have been acceptable for
         # most tests cases up until now. If the pipeline starts bottlenecking,
         # then maybe look at increasing these values.
+        # TODO: This is the same value as in Pipeline above. Abstract into a single declaration.
```
james-smith-za, line 740: It would be good to attend to this before you merge, methinks.

james-smith-za (changes requested):
It's getting close. There are some nit-picks and some outstanding things that you may have missed from previous reviews, which I'd like to be addressed before we merge.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -72,24 +73,29 @@
...
-    the copy is complete to reuse resources.
+    The RxQueueItem holds a reference to the received Chunk because its data is
+    copied into the GPU buffer asynchronously. Once the GPU processing loop is
+    done with the Chunk's data, it hands it back to the receiving loop to reuse
+    the resource. It is for this reason that the Chunk's heap presence is
+    stored separately.
+
+    The RxQueueItem also holds a reference count of the number of pipelines still
+    using this item. This is to ensure the item isn't freed before each data
+    product has had a chance to use the received data.
```
bmerry, line 86: A spot of bike-shedding:
```suggestion
    using this item. This is to ensure the item isn't freed before each
    pipeline has had a chance to use the received data.
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -72,24 +73,29 @@
...
         self.buffer_device = buffer_device
         self.present = present
         self.refcount = 0
-        self.chunk: recv.Chunk | None = None
         super().__init__(timestamp)
 
     def reset(self, timestamp: int = 0) -> None:
         """Reset the timestamp, events and chunk."""
         super().reset(timestamp=timestamp)
+        self.chunk: recv.Chunk | None = None
```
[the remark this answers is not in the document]
bmerry: `__init__` does call reset, via calling `super().__init__(...)`.

It would probably work to put just the type annotation (without the assignment) in `__init__`, and have just the assignment here.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -164,14 +167,8 @@ def __init__(self, output: Output, engine: "XBEngine", context: AbstractContext,
...
 
-    # TODO: Is it fair to impose this on the user?
-    @abstractmethod
-    def _populate_sensors(self) -> None:
-        """Define any Pipeline-specific sensors."""
-        raise NotImplementedError  # pragma: nocover
-
     def add_rx_item(self, item: RxQueueItem) -> None:
-        """Append a newly-received :class:`RxQueueItem`."""
+        """Append a newly-received :class:`RxQueueItem` to the `rx_item_queue`."""
```
bmerry, line 171:
```suggestion
        """Append a newly-received :class:`RxQueueItem` to the :attr:`_rx_item_queue`."""
```
(haven't actually tested that that renders/links correctly)

src/katgpucbf/xbgpu/engine.py
```diff
@@ -183,21 +180,23 @@ async def gpu_proc_loop(self) -> None:
...
         This function should do the following:
-        * Get an RxQueueItem off the rx_item_queue
-        * Ensure it is not a NoneType value (indicating shutdown sequence)
-        * await any outstanding events associated with the RxQueueItem
-        * TODO: Figure out whether the `is not None` check on the Chunk is necessary
-        * Apply GPU processing to data in the RxQueueItem
-            * Bind input buffer(s) accordingly
-        * Obtain a free TxQueueItem of the tx_free_item_queue
-            * Add event marker to wait for the proc_command_queue
-            * Put the prepared TxQueueItem on the tx_item_queue
```
bmerry: Need a blank line here - it's not rendering properly. Might also need a blank line before "- Add event marker..."

src/katgpucbf/xbgpu/engine.py
```diff
@@ -206,17 +205,29 @@ async def sender_loop(self) -> None:
...
 
+    @abstractmethod
+    def capture_start(self) -> None:
+        """Enable the transmission of this data product's stream."""
+        raise NotImplementedError  # pragma: nocover
+
+    @abstractmethod
+    def capture_stop(self) -> None:
+        """Disable the transmission of this data product's stream."""
+        raise NotImplementedError  # pragma: nocover
```
bmerry, line 229: You could probably reduce the amount of boilerplate you deal with by combining these into one function that takes a boolean.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -228,8 +239,7 @@ def __init__(
         context: AbstractContext,
         init_tx_enabled: bool,
```
bmerry, line 240: This parameter seems to be unused now.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
bmerry: Sphinx is configured to replicate the base class documentation in the derived class when it renders.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -860,28 +841,14 @@ def free_rx_item(self, item: RxQueueItem) -> None:
...
             self._active_in_sem.release()
-            # Need to wait for any outstanding events before recycling
-            self.add_service_task(
-                asyncio.create_task(
-                    self._push_recv_chunk(item),
-                    name="Receive Chunk Recycle Task",
-                )
-            )
+            # TODO: Don't recycle the chunk in the reset of the Item itself
+            # Need to separate the concerns
```
bmerry, line 845: Is there still something TODO?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -860,28 +841,14 @@ def free_rx_item(self, item: RxQueueItem) -> None:
...
+            item.chunk.recycle()  # type: ignore
```
bmerry, line 846: Rather than suppressing mypy, help it out with something like `assert item.chunk is not None`.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -892,11 +859,11 @@ async def _receiver_loop(self) -> None:
...
 
         This function does the following:
-        1. Wait for a chunk to be assembled on the receiver.
-        2. Get a free rx item off of the _rx_free_item_queue.
-        3. Initiate the transfer of the chunk from system memory to the buffer
-           in GPU RAM that belongs to the rx_item.
-        4. Place the rx_item on _rx_item_queue so that it can be processed downstream.
+        - Wait for a chunk to be assembled on the receiver.
+        - Get a free rx item off of the _rx_free_item_queue.
+        - Initiate the transfer of the chunk from system memory to the buffer
```
bmerry, line 864: I'm not so sure I like conversion of all the numbered lists to bullet lists. Numbered lists imply that things are happening in a specific order, while this bullet list could be taken as an unordered collection.

This documentation doesn't seem to get rendered (presumably because it is a private member), but I think you need a blank line before the list starts to make it correct.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -554,11 +547,11 @@ class XBEngine(DeviceServer):
     Data is passed between these items using :class:`asyncio.Queue`\s. The three
     processing functions are as follows:
 
-      1. :func:`_receiver_loop` - Receive chunks from network and initiate
-         transfer to GPU.
-      2. :func:`_gpu_proc_loop` - Perform the correlation operation.
-      3. :func:`_sender_loop` - Transfer correlated data to system RAM and then
-         send it out on the network.
+    - :func:`_receiver_loop` - Receive chunks from network and initiate
+        transfer to GPU.
```
bmerry, line 551: This is rendering funny, maybe due to the indentation.

Also `:func:` should be `:meth:` for methods.

bmerry (changes requested):

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
james-smith-za: I'd feel better about that solution.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: I'm absolutely with you; I was (still am) a big proponent of the numbered list. I changed it because I recall (way back) we had a discussion around numbered-list comments and general discomfort around maintaining/keeping them in parity with what the code is actually doing.

Will roll back and update to add a blank line.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
amishatishpatel: I'd mentioned in my [overall comment](https://github.com/ska-sa/katgpucbf/pull/600#issuecomment-1624877017) before re-requesting a review that I haven't implemented changes/updates to `test_engine.py` just yet (it might take another chunk of time Re: Testing different `heap_accumulation_threshold`s).

The more [top-level aspect](https://github.com/ska-sa/katgpucbf/pull/600#discussion_r1246734554) of this has been addressed, abstracting the common (base?) fixture from fgpu and xbgpu into `test/conftest.py`.

src/katgpucbf/xbgpu/engine.py
bmerry: Yes, there is a balance. But I think that discussion was more about the numbered comments sprinkled through the code, where adding/removing a step would require renumbering things at distant points, and there were also multiple levels of numbering. With this docstring the list is more self-contained.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: As far as I remember, this solution of `recycle()` just before/alongside the `item.reset()` isn't the most desirable?
* It was around the mention of "perhaps the Chunk will need to be reused before the RxQueueItem is done with it"
* Basically, the Chunk doesn't 'belong to' the RxQueueItem, so it should be "allowed to recycle" independently from the resetting of the item.

If that makes sense... Either way, perhaps not a great concern for right now?

bmerry: I just didn't want the call to `recycle` to be *inside* `reset`. I'm happy enough with the way things are.

We are delaying the release of chunks somewhat, so I would suggest running a test correlator at S-band rate before merging just to make sure this isn't going to cause any performance problems.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Please and thank you for the heads-up. Interestingly, I was able to correctly link gpu-proc and sender loops, but the receiver-loop just goes to bold. Is this the expected behaviour?
image

bmerry: I think when I last looked, the `_receiver_loop` method wasn't appearing in the built documentation (and hence it can't be linked to), but I didn't investigate why.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: This one is proving a tough nut to crack. It's syntactically similar to e.g. [this reference in fgpu](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/engine.py#L144), but fails to include the actual/needed hypertext anchor to `#katgpucbf.xbgpu.engine.Pipeline._rx_item_queue`
* Perhaps the `_rx_item_queue` needs to be mentioned as a class attribute like `n_rx/tx_items`?

Perhaps++ this is part of the reasoning behind fgpu's `Pipeline.add_in_item` docstring simply saying
> Append a newly-received :class:`~.InItem`.

* I should also add, the mention of [fgpu's _out_queue](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/engine.py#L926) is only ever done in 'plain old comments' - no rST/sphinx directives.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Is it showing up for you at the end of `self.send_stream = XSend(...)`?
* Arg is `tx_enabled=init_tx_enabled`

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +121,426 @@ def reset(self, timestamp: int = 0) -> None:
...
+    # These values are not configurable as they have been acceptable for
+    # most tests cases up until now. If the pipeline starts bottlenecking,
+    # then maybe look at increasing these values.
+    # TODO: Declare these constants elsewhere
+    n_rx_items = 3  # Too high means too much GPU memory gets allocated
+    n_tx_items = 2  # TODO: Will likely need to be == n_rx_items for BPipeline
+
+    def __init__(self, output: Output, engine: "XBEngine", context: AbstractContext, init_tx_enabled: bool) -> None:
+        self.engine = engine
+        self.output = output
```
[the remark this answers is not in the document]
amishatishpatel: Indeed, apologies for that. Do you think it's worth making this base Pipeline default to accepting `list[Output]`
* Then, in XPipeline, only take the first element of that list?
* This might add unnecessary complexity/force me to think about number of/relationship between `rx_item_queue`s vs `tx_item_queue`s.

I'm separately trying to figure out the neatest way of letting the base Pipeline (and `add_rx_item`, `shutdown`) know about the `rx_item_queue` - without adding a/nother `type: ignore`.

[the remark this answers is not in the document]
bmerry: > Do you think it's worth making this base Pipeline default to accepting list[Output]

It depends. Does the base class (or anything that interacts with pipelines without knowing what sort they are) need to access the (list of) outputs? If so, then your approach is what I'd go with. If not, you could just eliminate the output member from the base class entirely.

> I'm separately trying to figure out the neatest way of letting the base Pipeline (and add_rx_item, shutdown) know about the rx_item_queue

I'm not sure I follow. The `_rx_item_queue` is assigned in `Pipeline.__init__` - does that not achieve it?

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
bmerry: In that case I was immediately altering the copy. I don't feel it's necessary here if you're not altering the list.

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
amishatishpatel: Agreed on all accounts, but I think it's best if we chat through this one for a few minutes. I'll take you up on your offer of a consult tomorrow morning.

Basically,
* The creation of the item queues requires the `output.name`.
* It doesn't really make sense (to me) to remove the `output` arg (the origin of this comment) and maintain creation of item queues in `Pipeline.__init__`
    * It would require passing through just the `output.name`
    * Which isn't quite desired given idea of one BPipeline to many BOutputs.

amishatishpatel:
I think I've addressed the outstanding comments, namely:
* removing `output` arg from `Pipeline.__init__` (and replacing with a Pipeline `name` arg), and
* abstracting `n_{rx, tx}_items` constants into `xbgpu/__init__.py`
    * I'd chosen `xbgpu/__init__.py` instead of the top of `engine.py` because I'd seen `DIG_RMS_DBFS` defaults also placed in `fgpu/__init__` and used in `engine.py` - feel free to tell me to move it.

I want to make some headway on `test_engine.py` before requesting another round of reviewing.

james-smith-za (commented):
One RFQ. Otherwise I am actually quite happy with this.

I can't figure out where the 1-A problem is but I have a suspicion. I'll do a couple more tests then post an update on JIRA later.

test/xbgpu/test_engine.py
amishatishpatel: I've changed the way these final checks are done to only loop over XPipelines in `xbengine._pipelines` - I use `pipeline.output.name` when looking up stats.

test/xbgpu/conftest.py
```diff
@@ -27,17 +29,35 @@ def n_src_streams() -> int:  # noqa: D401
...
 
     def constructor(thread_pool, endpoints, config, *args, **kwargs):
-        return spead2.send.asyncio.InprocStream(thread_pool, [queue], config)
+        stream_queues = [spead2.InprocQueue() for _ in endpoints]
+        queues.extend(
+            queue
+            for queue, endpoint in zip(stream_queues, endpoints)
+            if IPv4Address(endpoint[0]) in mock_send_stream_network
+        )
+        return spead2.send.asyncio.InprocStream(thread_pool, stream_queues, config)
```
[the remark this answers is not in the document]
amishatishpatel: Argh, this has come back to bite me. There are similarities, but they are not identical. I'm tracing through the changes/blame to see what came from where. `test/fgpu/` still references the old/singular **mock_recv_stream** and not *mock_recv_streams*.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
bmerry: I'm not following why it's necessary to override the base implementation. Are there UDP send streams being created that you're trying to filter out? Or is it just future-proofing for when you add beamforming?

test/xbgpu/test_engine.py
```diff
@@ -236,116 +271,157 @@ def _create_heaps(
...
-        n_ants, n_channels_per_substream, n_spectra_per_heap
+        n_ants, n_channels_per_substream, n_spectra_per_heap, frequency
             See :meth:`_create_heaps` for more info.
 
         Returns
         -------
         device_results
             Array of all GPU-generated data of shape
-            - (n_total_accumulations, n_channels_per_substream, n_baselines, COMPLEX)
+            - (len(corrprod_outputs), n_total_accumulations, n_channels_per_substream, n_baselines, COMPLEX)
```
bmerry, line 331: This shape is confusing. Each product can have a different n_total_accumulations, so the actual data doesn't fit into a neat rectangular array. I'd suggest you rather return a list of arrays (one for each corrprod_output).

test/xbgpu/test_engine.py
```diff
@@ -236,116 +271,157 @@ def _create_heaps(
...
+        n_accumulations_completed
+            Number of completed accumulations for each corrprod_output
         """
         max_packet_size = n_spectra_per_heap * N_POLS * COMPLEX * SAMPLE_BITWIDTH // 8 + PREAMBLE_SIZE
         max_heaps = n_ants * HEAPS_PER_FENGINE_PER_CHUNK * 10
-        feng_stream = self._make_feng(mock_recv_stream, max_packet_size, max_heaps)
+        feng_stream = self._make_feng(mock_recv_streams, max_packet_size, max_heaps)
+
+        acc_index_counts: list[dict[int, int]] = []
```
bmerry, line 339: Would be helpful to have a comment indicating what this structure represents.

test/xbgpu/test_engine.py
```diff
@@ -236,116 +271,157 @@ def _create_heaps(
...
+                )
+            )
+
+        # NOTE: The data generation generates an extra heap or several for the XPipeline
+        # with the smallest heap-accumulation-threshold - resulting in an incomplete final
+        # accumulation. Incomplete accumulations are purposefully dictated by `batch_indices`
+        # constructed by the caller.
+        for i, corrprod_output in enumerate(corrprod_outputs):
+            for batch_index in batch_indices:
+                adjusted_index = batch_index // corrprod_output.heap_accumulation_threshold
```
bmerry, line 353: "adjusted" doesn't really describe what adjustment is being made. It looks like it is an accumulation index, so maybe name it `acc_index`.

test/xbgpu/test_engine.py
```diff
@@ -236,116 +271,157 @@ def _create_heaps(
...
 
-        mock_recv_stream.stop()
-
-        # It is expected that the first packet will be a descriptor.
-        ig_recv = spead2.ItemGroup()
-        heap = await recv_stream.get()
-        items = ig_recv.update(heap)
-        assert len(list(items.values())) == 0, "This heap contains item values not just the expected descriptors."
+        for q in mock_recv_streams:
+            q.stop()
```
bmerry, line 364: It's more common in katgpucbf tests to use `queue` for a variable holding an InprocQueue.

test/xbgpu/test_engine.py
```diff
@@ -236,116 +271,157 @@ def _create_heaps(
...
+
+        # NOTE: The data generation generates an extra heap or several for the XPipeline
+        # with the smallest heap-accumulation-threshold - resulting in an incomplete final
+        # accumulation. Incomplete accumulations are purposefully dictated by `batch_indices`
+        # constructed by the caller.
+        for i, corrprod_output in enumerate(corrprod_outputs):
+            for batch_index in batch_indices:
+                adjusted_index = batch_index // corrprod_output.heap_accumulation_threshold
+                acc_index_counts[i][adjusted_index] += 1
+            # `adjusted_index` should be the last accumulation index for this corrprod
```
bmerry, line 355: That comment isn't helping me understand what the code below it is doing. It seems like if the last accumulation is incomplete you remove it from the count array? Why?

test/xbgpu/test_engine.py
```diff
@@ -452,15 +527,34 @@ async def test_xengine_end_to_end(
...
+
+        # The arbitrarily-selected number above might not be at a convenient
+        # place. We want to be just before a LCM of the heap accumulation
+        # indices of all the corrprods, so that we can get an incomplete
+        # accumulation as the output first.
+        lowest_common_multiple = lcm(
+            *[corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]
+        )
+        advance = lowest_common_multiple - (first_batch_index % lowest_common_multiple)
+        first_batch_index += advance
```
bmerry, line 542: This is quite a complicated way to get an index that is a multiple of all the heap sizes. Since there is nothing special about batch 123, you could do something like
```python
first_batch_index = 123 * np.product([corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs])
```

test/xbgpu/test_engine.py
```diff
@@ -452,15 +527,34 @@ async def test_xengine_end_to_end(
...
+
+        max_heap_acc_threshold = max(
+            corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs
+        )
+        min_full_accumulations = (
+            lcm(*[corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs])
```
bmerry, line 548: This is the third time you've used `[corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]`. Things will be more readable if you assign it to a variable once.

test/xbgpu/test_engine.py
```diff
@@ -452,15 +527,34 @@ async def test_xengine_end_to_end(
...
+        )
+
         timestamp_step = n_samples_between_spectra * n_spectra_per_heap
         missing_antennas = set() if missing_antenna is None else {missing_antenna}
 
-        range_start = n_channels_per_substream * CHANNEL_OFFSET
+        range_start = frequency
         range_end = range_start + n_channels_per_substream - 1
-        assert xbengine.sensors["chan-range"].value == f"({range_start},{range_end})"
+        assert xbengine.sensors[f"{corrprod_outputs[0].name}.chan-range"].value == f"({range_start},{range_end})"
```
bmerry, line 557: Check it for every corrprod, not just the first.

test/xbgpu/test_engine.py
```diff
@@ -470,7 +564,7 @@ def sensor_observer(sync_sensor: aiokatcp.Sensor, sensor_reading: aiokatcp.Readi
             """Record sensor updates in a list for later comparison."""
             actual_sensor_updates.append((sensor_reading.value, sensor_reading.status))
 
-        xbengine.sensors["rx.synchronised"].attach(sensor_observer)
+        xbengine.sensors[f"{corrprod_outputs[0].name}.rx.synchronised"].attach(sensor_observer)
```
bmerry, line 567: Check it for every corrprod, not just the first.

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
             # received batch is from the middle of an accumulation.
-            batch_start_index = (
-                first_accumulation_index + 1
-            ) * heap_accumulation_threshold - HEAPS_PER_FENGINE_PER_CHUNK
-            batch_end_index = (first_accumulation_index + 1 + n_full_accumulations) * heap_accumulation_threshold
-            device_results = await self._send_data(
-                mock_recv_stream,
-                recv_stream,
-                heap_accumulation_threshold=heap_accumulation_threshold,
+            batch_start_index = (first_batch_index) * max_heap_acc_threshold - HEAPS_PER_FENGINE_PER_CHUNK
```
bmerry, line 585: Multiplying a batch index by a heap_acc_threshold doesn't make sense.

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+            batch_end_index = (first_batch_index + min_full_accumulations) * max_heap_acc_threshold
```
bmerry, line 586: Should probably be
```suggestion
            batch_end_index = first_batch_index + min_full_accumulations * max_heap_acc_threshold
```
Also since `min_full_accumulations * max_heap_acc_threshold` is (currently) just the LCM, this could probably be simplified.

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
-                heap_accumulation_threshold,
-                n_channels_per_substream,
-                n_ants,
-                n_spectra_per_heap,
-                missing_antenna,
+        xpipelines: list[XPipeline] = [pipeline for pipeline in xbengine._pipelines if isinstance(pipeline, XPipeline)]
+        for i, pipeline in enumerate(xpipelines):
+            output_name = pipeline.output.name
+            assert (
+                prom_diff.get_sample_diff("output_x_incomplete_accs_total", {"stream": f"{output_name}"})
```
bmerry, line 639:
```suggestion
                prom_diff.get_sample_diff("output_x_incomplete_accs_total", {"stream": output_name})
```

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
-        # Could manually calculate it here, but it's available inside the send_stream
-        assert prom_diff.get_sample_diff("output_x_bytes_total") == (
-            xbengine.send_stream.heap_payload_size_bytes * n_total_accumulations
-        )
-        assert prom_diff.get_sample_diff("output_x_visibilities_total") == (
-            n_channels_per_substream * n_baselines * n_total_accumulations
-        )
-        assert prom_diff.get_sample_diff("output_x_clipped_visibilities_total") == 0
+            assert (
+                prom_diff.get_sample_diff("output_x_heaps_total", {"stream": f"{output_name}"})
```
bmerry, line 643:
```suggestion
                prom_diff.get_sample_diff("output_x_heaps_total", {"stream": output_name})
```

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+                == n_accumulations_completed[i]
+            )
+            # Could manually calculate it here, but it's available inside the send_stream
+            assert prom_diff.get_sample_diff("output_x_bytes_total", {"stream": f"{output_name}"}) == (
```
bmerry, line 647:
```suggestion
            assert prom_diff.get_sample_diff("output_x_bytes_total", {"stream": output_name}) == (
```

test/xbgpu/test_engine.py
```diff
@@ -487,64 +582,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+                pipeline.send_stream.heap_payload_size_bytes * n_accumulations_completed[i]
+            )
+            assert prom_diff.get_sample_diff("output_x_visibilities_total", {"stream": f"{output_name}"}) == (
```
bmerry, line 650:
```suggestion
            assert prom_diff.get_sample_diff("output_x_visibilities_total", {"stream": output_name}) == (
```

test/xbgpu/test_engine.py
```diff
@@ -553,9 +659,9 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
         # Depending on the `missing_antenna` parameter, the full accumulations
         # will either be all complete or incomplete.
         if missing_antenna is not None:
-            expected_sensor_updates += [(False, aiokatcp.Sensor.Status.ERROR)] * n_full_accumulations
+            expected_sensor_updates += [(False, aiokatcp.Sensor.Status.ERROR)] * (max(n_accumulations_completed) - 1)
```
bmerry, line 662: At present, actual_sensor_updates is collecting from output 0, so this should probably use `n_accumulations_completed[0]` rather than the maximum.

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -590,30 +697,34 @@ async def test_saturation(
         n_baselines = n_ants * (n_ants + 1) * 2
 
         def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
-            timestamp = timestamp_step * batch_index
+            timestamp = batch_index * timestamp_step
```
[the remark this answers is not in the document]
amishatishpatel: Indeed indeed, mostly because it bugged me that the `heap_factory` in `test_xengine_end_to_end` (~Line 570) had this timestamp calculation as
> `timestamp = batch_index * timestamp_step`

And this was the other way around (for no apparent reason?).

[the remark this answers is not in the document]
bmerry: That's a good reason to swap it around then.

test/xbgpu/test_engine.py
amishatishpatel: I didn't expect this to be a quick fix, but I also didn't expect to run into the issue I have. Decided to update this to fill a dictionary (key == sensor name) and the value is a list of tuples.
* Each update is seemingly called once (correctly) - verified with logging statements,
* But the eventual dictionary shows updates from both sensors present in both lists.

Same issue when generating the `expected_sensor_updates` later on. Will keep plugging at it, but just had to share this strangeness.

amishatishpatel: Yeah I figured I hadn't discovered a new branch of quantum entanglement. Thank you for the reminder.

test/xbgpu/test_engine.py
james-smith-za: This was mine actually, it was part of the tweaking that I did to make sure that we were starting the test in a sane place.

I wouldn't suggest further work on this right now - your comment is valid, but rather than delaying this PR further I'd suggest just creating a small technical-debt ticket once we've got it merged.

test/xbgpu/test_engine.py
```diff
@@ -336,25 +336,36 @@ async def _send_data(
...
+        # each accumulation index (for each corrprod output). If a heap count
+        # for a corrprod's final accumulation is not equal to its
+        # `heap_accumulation_threshold`, the accumulation index is discarded
+        # from the list (used later on). If this function attempted to receive
+        # data for an XPipeline's final, incomplete accumulation, it would
+        # result in a _spead2.Stopped error as that accumulation would never
+        # have been processed/transmitted by the XPipeline.
+        # - i.e. the test would attempt to get one more heap than the stream
+        #   contained.
+        acc_index_heap_counts: list[dict[int, int]] = []
```
james-smith-za, line 354: Is the whole exercise of determining the LCM of the two heap accumulation thresholds not to ensure that this doesn't happen? Or did I miss something?

test/xbgpu/test_engine.py
```diff
@@ -663,16 +668,21 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
             )
             assert prom_diff.get_sample_diff("output_x_clipped_visibilities_total", {"stream": output_name}) == 0
 
-        expected_sensor_updates: list[tuple[bool, aiokatcp.Sensor.Status]] = []
+        # NOTE: Much like initialising `actual_sensor_updates`, a similar
+        # approach must be taken when geenrating `expected_sensor_updates`.
```
james-smith-za, line 672: typo here I think ("generating")

test/xbgpu/test_engine.py
amishatishpatel: As always, fair points. However I'm gona plead the Fifth and cite [James's comment above](https://github.com/ska-sa/katgpucbf/pull/600#discussion_r1296857624) to make this part of a technical-debt-type overhaul.

Your call is still important to us and we thank you for your patience.

james-smith-za: When you create a ticket Amish please just link to these comments. It'll help with the context.

test/xbgpu/test_engine.py
amishatishpatel: I'd added in this logic to resolve the `spead2 ring buffer has been stopped` error, before we delved further into the 1A debug. I can look into it further now if you want (to see if it's still necessary), but I'm leaning towards making this part of the eventual technical-debt ticket for test cleanup.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
amishatishpatel: Promise I haven't been ignoring this comment. Yes, largely for future-proofing reasons. Admittedly, I may have included this pre-emptively.

test/xbgpu/test_engine.py
james-smith-za: Yeah IMO just leave it for now. It currently works. We will probably have to make several tickets but that's OK.

test/xbgpu/test_engine.py
amishatishpatel: For the record, I've created [NGC-1048](https://skaafrica.atlassian.net/browse/NGC-1048).

test/xbgpu/test_engine.py
james-smith-za: Maybe I'm missing something, but ... why not?

test/xbgpu/test_engine.py
amishatishpatel: This one grew some tentacles and allowed me to remove the `n_accumulations_completed` list that gets returned alongside this. I've just finished neatening up how it's used in the data verification in the unit test itself, so it might require a bit more engagement during the final review.

Apologies in advance and thank you for this suggestion.

test/xbgpu/test_engine.py
```diff
@@ -307,44 +316,54 @@ async def _send_data(
...
+                    f"actual: {hex(ig_recv['timestamp'].value)}."
+                )
+
+                assert ig_recv["frequency"].value == n_channels_per_stream * CHANNEL_OFFSET, (
+                    "Output channel offset not correct. "
+                    f"Expected: {n_channels_per_stream * CHANNEL_OFFSET}, "
+                    f"actual: {ig_recv['frequency'].value}."
+                )
+
+                device_results[i][j] = ig_recv["xeng_raw"].value
```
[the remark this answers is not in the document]
amishatishpatel: Just btw, `device_results` itself is not a numpy array anymore (as of a few mins ago).

test/xbgpu/test_engine.py
james-smith-za: Is `max_heap_acc_threshold` not heaps per batch?
Doesn't change the overall thrust of your argument though, in that you then end up with units of heaps.

So there does seem to be at least some naming inconsistency there.

bmerry: I don't feel confident that the test is doing what it did in `main` (in terms of how it aligns the batches that get sent with the accumulations) since I can't easily work my way through the logic when I don't know what units the variables are in.

If you're confident that the test is doing what it's supposed to do then go ahead and fix it up later (but not much later).

test/xbgpu/test_engine.py
```diff
@@ -602,7 +613,8 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
             # Or assert if incomplete_accs_total == incomplete_accums_counter * len(xbengine._pipelines)
             incomplete_accums_counter = 0
             base_batch_index = batch_start_index
-            for j in range(n_accumulations_completed[i]):
+            # for j in range(n_accumulations_completed[i]):
```
bmerry, line 616: Remove code rather than commenting it out.

bmerry (changes requested):

test/xbgpu/test_engine.py
amishatishpatel: Argh, my apologies, thought I cleaned up properly.

test/xbgpu/test_engine.py
```diff
@@ -327,58 +327,69 @@ async def _send_data(
...
-                n_channels_per_substream,
-                n_baselines,
-                COMPLEX,
-            ),
-            dtype=np.int32,
-        )
+        device_results = [
+            np.zeros(
+                shape=(
+                    len(acc_indices.keys()),  # n_accumulations for this XPipeline
```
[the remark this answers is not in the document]
bmerry: I was quite happy with the explicit `.keys` when iterating, since the fact that iterating a dictionary iterates the keys can be a bit non-intuitive. But `len` isn't the same thing as iteration - it just tells you how big the dictionary is (or as you've written it, how big the set of keys in the dictionary is).

Anyway, as mentioned it's just bike-shedding; feel free to leave as is if it makes more sense to you.

test/xbgpu/test_engine.py
bmerry: I think a comment here would still be helpful, even though it's now a list of sets.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +127,417 @@ def reset(self, timestamp: int = 0) -> None:
...
+
+    def shutdown(self) -> None:
+        """Start a graceful shutdown after the final call to :meth:`add_rx_item`."""
+        self._rx_item_queue.put_nowait(None)
+
+    @abstractmethod
+    async def gpu_proc_loop(self) -> None:
+        """Perform all GPU processing of received data in a continuous loop.
+
+        This function should do the following:
```
bmerry, line 192: Need a blank line here to make the list render correctly.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +127,417 @@ def reset(self, timestamp: int = 0) -> None:
...
+          first heap output.
+        - It also provides an opportunity to bind buffers before processing is queued.
+        """
+        raise NotImplementedError  # pragma: nocover
+
+    @abstractmethod
+    async def sender_loop(self) -> None:
+        """Send heaps to the network in a continuous loop.
+
+        This function should do the following:
```
bmerry, line 218: Need a blank line here to make the list render properly.

src/katgpucbf/xbgpu/engine.py
bmerry, line 192:
```suggestion
        This method does the following:
```

I assume you've written "should do" because this is the abstract declaration, but the same docstring gets inherited by the implementation, and it's confusing when consulting the XPipeline documentation. Same applies to the other abstract methods in this class.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -117,6 +127,417 @@ def reset(self, timestamp: int = 0) -> None:
...
+
+        - Transmit heap buffer onto the network
+        - Place the TxQueueItem back on the tx_free_item_queue once complete
+        """
+        raise NotImplementedError  # pragma: nocover
+
+    @abstractmethod
+    def capture_enable(self, enable: bool = True) -> None:
+        """Enable/Disable the transmission of this data product's stream."""
+        raise NotImplementedError
```
bmerry, line 235: Since you've used `#pragma: nocover` a few lines further up, you probably want it here too.

src/katgpucbf/xbgpu/engine.py
bmerry: The mis-indentation is still messing up the rendering.

bmerry (changes requested):
No more code change comments - just some requests to fix up documentation.

amishatishpatel:
@bmerry Jenkins is happy with the updates and I can verify the problematic indentation has been fixed (screenshot below). I'm going to hold off on pinging you officially for a review because I want to make sure there are/aren't any code updates required (here) for the `?capture-{start, stop}` change in katsdpcontroller. 

image

bmerry (approved):

PR #609, opened 2023-08-01
Bump dependencies to latest versions

bmerry:
A couple of packages are kept back:
- mypy is limited to 1.4.1, since that's the latest version at https://github.com/pre-commit/mirrors-mypy
- pyparsing is kept back at 3.0.9, because matplotlib has a dependency on pyparsing<3.1 which prevents upgrading to 3.1.1.
- numba is limited to 0.56.* because 0.57 is logging warnings about the TBB library on Ubuntu 22.04 being too old. Using 0.56 restricts numpy to 1.23.*.

- [ ] Run a qualification test to ensure it still works

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
       - id: no-commit-to-branch # without arguments, master/main will be protected.
       - id: trailing-whitespace
   - repo: https://github.com/psf/black
-    rev: 23.3.0
+    rev: 23.7.0
     hooks:
       - id: black
   - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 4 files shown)


src/katgpucbf/fgpu/delay.py
```diff
@@ -219,6 +219,7 @@ def _popleft_until(self, start: int) -> None:
                 f"Timestamp {start} is before start of first linear model "
                 f"at {self._models[0].start} - possibly due to non-monotonic queries",
                 NonMonotonicQueryWarning,
+                stacklevel=3,
```
james-smith-za, line 222: I'm curious as to what this warning was.

james-smith-za (approved):

src/katgpucbf/fgpu/delay.py
bmerry: > B028: No explicit stacklevel argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

I went with 3 rather than 2 because this is a private helper function rather than the function the user actually calls.

PR #611, opened 2023-08-01
Add pre-tuning support

bmerry:
A pre-tuning database is added to the repository, and the Dockerfile
inserts it into the image so that it will be used at runtime. This
avoids every container re-doing autotuning (provided the parameters are
in the space that's pre-tuned), which could be problematic when run by
multiple containers simultaneously (skewing tuning results).

- [x] Update to a release of katsdpsigproc supporting KATSDPSIGPROC_TUNE_DB

Closes NGC-983.


docker/autotune.py
```diff
@@ -0,0 +1,69 @@
...
+        device_filter=lambda device: device.is_cuda,
+    ) as context, tempfile.NamedTemporaryFile(
+        prefix="tuning",
+        suffix=".db",
+        dir=args.database.parent,
+        delete=False,
+    ) as temp_db:
+        should_delete = True
+        try:
+            os.environ["KATSDPSIGPROC_TUNE_DB"] = temp_db.name
```
james-smith-za, line 53: Does the autotune function know to store stuff here but not read from it?

bmerry: It will read from there too, but that's desirable. We want to be isolated from whatever is in the user's existing tuning DB at e.g. `~/.cache/katsdpsigproc/tuning.db`.

james-smith-za: Right, but presumably since you've just created the file, it'll be empty so the template won't be able to pull values from there, so it knows it'll need to run the autotune.

I think I follow the logic. Sounds reasonable to me.

james-smith-za (approved):

PR #615, opened 2023-08-14
xbgpu: Call QueueItem.reset when item is acquired, not released

bmerry:
This gives an opportunity to await any events that came with the item on the free item queue. It fixes a race condition introduced as part of the work on NGC-929, where events were getting lost when `reset()` cleared the event list (there is a bigger issue that this is a foot-gun, but that's a problem for another day).

I also changed tx_items to work the same way, although it should be a non-issue since all work on the TxItem is synchronously completed before it goes onto the free queue.

Tests are still failing with an out of resources error, but I think that's unrelated.

- [ ] If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`
- [ ] If modules are added/removed: use sphinx-apidoc to update files in `doc/`

Closes NGC-xxxx.

### DIFF
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -354,14 +354,14 @@ class XPipeline(Pipeline):
         self.engine.sensors[f"{self.output.name}.rx.synchronised"].value = bool(tx_item.present_ants.all())
 
         self.correlation.reduce()
-        self._proc_command_queue.finish()
+        tx_item.add_marker(self._proc_command_queue)
         self._tx_item_queue.put_nowait(tx_item)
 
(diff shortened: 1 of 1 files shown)


amishatishpatel (approved):
So was there a premature reset on the RxQueueItem? Or does your change on line 364 actually solve the race condition? (Or a combination of all updates)

bmerry:
> So was there a premature reset on the RxQueueItem? Or does your change on line 364 actually solve the race condition? (Or a combination of all updates)

The original reset on line 531 was causing the race condition, by emptying the events list without waiting for the events.

bmerry:
I've created https://skaafrica.atlassian.net/browse/NGC-1042 to capture the technical debt issue.

PR #617, opened 2023-08-25
Clamp dig-rms-dbfs sensor to DBL_MIN in case of zero power

james-smith-za:
Or rather strictly speaking it's to np.finfo(np.float64).min, which I think may be the same as DBL_MIN under the hood, though it's probably implementation dependent.

The context to this is that previously in the case of zero input power we'd assign a -inf value. This is OK for the IEEE double specification, but the KATCP spec only allows decimal values for float sensors, so we weren't compliant in this sense. Issues arose with CAM's NATS system which didn't like the -inf values.

I couldn't find any suitable unit test where these sensors were exercised, so I added a simple one.

Closes NGC-1059

### DIFF
--- a/test/fgpu/test_engine.py
+++ b/test/fgpu/test_engine.py
@@ -958,7 +958,9 @@ class TestEngine:
 
     # It's easier to use a constant voltage. Also need to check the case were
     # the input power is zero.
-    @pytest.mark.parametrize("inout", [(0, np.finfo(np.float64).min), (64, -15.034518566376697)])
+    @pytest.mark.parametrize(
+        "input_voltage,output_power_dbfs", [(0, np.finfo(np.float64).min), (64, pytest.approx(-15.0345186))]
+    )
     async def test_dig_rms_dbfs_sensors(
         self,
         mock_recv_stream: spead2.InprocQueue,
@@ -966,13 +968,13 @@ class TestEngine:
         engine_server: Engine,
         engine_client: aiokatcp.Client,
         output: Output,
-        inout: tuple[int, float],
+        input_voltage: int,
+        output_power_dbfs: float,
     ) -> None:
         """Test that the ``dig-rms-dbfs`` sensors are set correctly."""
         sensors = [engine_server.sensors[f"input{pol}.dig-rms-dbfs"] for pol in range(N_POLS)]
         sensor_update_dict = self._watch_sensors(sensors)
         n_samples = 10 * CHUNK_SAMPLES
-        input_voltage, output_power_dbfs = inout
         dig_data = np.full((2, n_samples), input_voltage, np.int16)
 
         await self._send_data(
--- a/test/fgpu/test_engine.py
+++ b/test/fgpu/test_engine.py
@@ -956,6 +956,40 @@ class TestEngine:
         assert [r.value for r in sensor_update_dict[sensors[1].name]] == [0, 0] + [10000] * 7
         assert [r.timestamp for r in sensor_update_dict[sensors[1].name]] == expected_timestamps
 
+    # It's easier to use a constant voltage. Also need to check the case were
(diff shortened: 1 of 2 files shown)


test/fgpu/test_engine.py
```diff
@@ -956,6 +956,40 @@ async def test_dig_clip_cnt_sensors(
         assert [r.value for r in sensor_update_dict[sensors[1].name]] == [0, 0] + [10000] * 7
         assert [r.timestamp for r in sensor_update_dict[sensors[1].name]] == expected_timestamps
 
+    # It's easier to use a constant voltage. Also need to check the case were
+    # the input power is zero.
+    @pytest.mark.parametrize("inout", [(0, np.finfo(np.float64).min), (64, -15.034518566376697)])
```
bmerry, line 961: How about
```suggestion
    @pytest.mark.parametrize("input_voltage,output_power_dbfs", [(0, np.finfo(np.float64).min), (64, -15.034518566376697)])
```
so that pytest will destructure the tuple for you.

test/fgpu/test_engine.py
```diff
@@ -956,6 +956,40 @@ async def test_dig_clip_cnt_sensors(
...
+        engine_client: aiokatcp.Client,
+        output: Output,
+        inout: tuple[int, float],
+    ) -> None:
+        """Test that the ``dig-rms-dbfs`` sensors are set correctly."""
+        sensors = [engine_server.sensors[f"input{pol}.dig-rms-dbfs"] for pol in range(N_POLS)]
+        sensor_update_dict = self._watch_sensors(sensors)
+        n_samples = 10 * CHUNK_SAMPLES
+        input_voltage, output_power_dbfs = inout
+        dig_data = np.ones((2, n_samples), np.int16) * input_voltage
```
bmerry, line 976:
```suggestion
        dig_data = np.full((2, n_samples), input_voltage, np.int16)
```

bmerry (changes requested):

test/fgpu/test_engine.py
bmerry, line 961: It might be worth using something like `pytest.approx(-15.0345186)` so that a change in the implementation that causes slightly different rounding doesn't break the test. You'd just need to check that the test ID (in `pytest -v`) still looks sane.

By the way, how is that magic value derived? Did you calculate it (in which case it might make sense to put the calculation in the code to make it self-documenting) or just observe what the value happened to be?

james-smith-za: That looks neater, I like it.

james-smith-za: I started off trying to calculate it, but the fact that it was Friday afternoon got to me and I decided just to run through and see what value came out. Then I made note of it.

The parameterisation looks like this:
```
test/fgpu/test_engine.py::TestEngine::test_dig_rms_dbfs_sensors[wideband-NVIDIA GeForce RTX 3080 Ti (CUDA)-0--1.7976931348623157e+308] PASSED            [ 25%]
test/fgpu/test_engine.py::TestEngine::test_dig_rms_dbfs_sensors[wideband-NVIDIA GeForce RTX 3080 Ti (CUDA)-64-output_power_dbfs1] PASSED                 [ 50%]
test/fgpu/test_engine.py::TestEngine::test_dig_rms_dbfs_sensors[narrowband-NVIDIA GeForce RTX 3080 Ti (CUDA)-0--1.7976931348623157e+308] PASSED          [ 75%]
test/fgpu/test_engine.py::TestEngine::test_dig_rms_dbfs_sensors[narrowband-NVIDIA GeForce RTX 3080 Ti (CUDA)-64-output_power_dbfs1] PASSED               [100%]
```
which seems OK to me.

bmerry (approved):

PR #619, opened 2023-09-04
Redesign QueueItem event management

bmerry:
The primary motivation is to address NGC-1042 by not throwing away events (which might not have been awaited) in `reset`. It also keeps the event list bounded (assuming a bounded number of command queues) by retaining just a latest event per comment queue.

I think this also fixes some race conditions when using vkgdr or PeerDirect in fgpu:
- When using PeerDirect, the saturation and power data was downloaded without waiting for the computation to complete.
- When using vkgdr, `_push_recv_chunks` was called asynchronously with a reference to the item's events, which could potentially change before the actual await.

Closes NGC-1042.


src/katgpucbf/queue_item.py
```diff
@@ -42,37 +42,51 @@ class QueueItem:
...
-    def add_marker(self, command_queue: AbstractCommandQueue) -> None:
+    def add_marker(self, command_queue: AbstractCommandQueue) -> AbstractEvent:
         """Add an event to the list of events in the QueueItem.
 
         The event represents all previous work enqueued to `command_queue`.
         """
-        self.events.append(command_queue.enqueue_marker())
+        marker = command_queue.enqueue_marker()
+        self._events[command_queue] = marker
+        return marker
```
amishatishpatel, line 60: 1. Not to be too pedantic, but does the docstring need to be updated? Specifically around "Add an event to the *list of events*.
    * Admittedly I don't have a suggestion, but technically we're not working with lists anymore.
    * Feel free to ignore this request.
2. What happens if the caller doesn't do anything with the returned `marker`?
    * e.g. fgpu/engine.py Lines [667](https://github.com/ska-sa/katgpucbf/blob/NGC-1042-reset-events/src/katgpucbf/fgpu/engine.py#L667), [720](https://github.com/ska-sa/katgpucbf/blob/NGC-1042-reset-events/src/katgpucbf/fgpu/engine.py#L720) and [850](https://github.com/ska-sa/katgpucbf/blob/NGC-1042-reset-events/src/katgpucbf/fgpu/engine.py#L850) (few more examples in xbgpu/engine.py as well)

amishatishpatel (commented):
I think I'm following along correctly. Just the ~one comment for clarification.

src/katgpucbf/queue_item.py
bmerry: > 1. Not to be too pedantic, but does the docstring need to be updated? Specifically around "Add an event to the _list of events_.

Fair point - @james-smith-za any suggestions on how to reword it, that avoids saying anything about the internal representation?

>     2. What happens if the caller doesn't do anything with the returned `marker`?

That's perfectly fine - it's just returned in case the caller wants to know (there was one place where I needed it).

james-smith-za: Something like "Add an (awaitable possibly?) event to the QueueItem" would probably be fine IMO.

james-smith-za (approved):
LGTM once the wording of the docstring is sorted out.

james-smith-za (approved):
I like it.

bmerry:
IIRC this passed Jenkins before I reworded a comment, so I'm going to assume it is okay.

PR #622, opened 2023-09-07
Drastically reduce gc pauses

bmerry:
Use gc.freeze to move the objects allocated at startup to a permanent generation that is not subject to the GC. This makes GC traversal much faster, and eliminates some 57ms-long GC pauses.

The fgpu src buffer is reduced to 64MiB (I've tested it as low as 16MiB, although not for very long).

xbgpu gets the same gc.freeze treatment. I haven't touched the buffer size since I haven't done any benchmarking on it.

- [x] Run a largish correlator in the lab

Closes NGC-1060.

### DIFF
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -23,6 +23,7 @@ actual running of the processing.
 
 import argparse
 import asyncio
+import gc
 import logging
 import math
 from collections.abc import Callable, Sequence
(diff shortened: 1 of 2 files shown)


james-smith-za (approved):
I'm curious as to why you reduce the buffer size though. It's not as though we're short on memory?

bmerry:
> I'm curious as to why you reduce the buffer size though. It's not as though we're short on memory?

Using less memory can mean less TLB thrashing and better cache use. I'm not sure if it'll make any difference in this specific case, but it always bugged me that we had such a big buffer.

PR #625, opened 2023-09-10
Install buildx in Jenkins Docker image

bmerry:
Closes NGC-1088.


james-smith-za (approved):
Thanks for digging into this. I've subscribed to the relevant bug so hopefully there'll be a fix soon.

PR #620, opened 2023-09-05
xbgpu documentation cleanup after PR #600

amishatishpatel:
As the subject suggests, tackling the documentation-related tickets that are part
of NGC-1048. I've made the commits 1:1 with the ticket names/numbers, with the
exception of NGC-1054 - that took two commits. Reviewing each commit on its
own might be easier to verify the ticket it addresses.

Can't imagine this needs two reviewers, but observations were raised by you both
so... please and thank you.

Checklist (if not applicable, edit to add `(N/A)` and mark as done):

- [x] (N/A) If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`
- [x] (N/A) If modules are added/removed: use sphinx-apidoc to update files in `doc/`
- [x] Ensure copyright notices are present and up-to-date
- [x] (N/A) If qualification tests are changed: attach a sample qualification report
- [ ] If design has changed: ensure documentation is up to date
- [x] (N/A) If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match

Closes the following:
* NGC-1053,
* NGC-1054,
* NGC-1055,
* NGC-1056, 
* NGC-1066,
* NGC-1067,
* NGC-1068.

PR #626, opened 2023-09-11
xbgpu/test_engine cleanup after PR #600

amishatishpatel:
As the subject suggests, and following #620, this addresses some issues raised
in xbgpu's test_engine.py. See commits for more details.

Took me some non-zero time to re-re-wrap my head around the terminology and its origins.
The corresponding tickets have a bit of discussion that might be handy in reviewing.

I don't think it requires two reviewers, but Bruce made the original observations on #600 and
James created the tickets. Whenever you can, please and thank you.

Closes the following:
* NGC-1050,
* NGC-1051, and
* NGC-1069.


amishatishpatel:
I see Jenkins tells me `docker build and push failed for the ffg reason. I'm going to merge `main` into here and rebuild.
```
ERROR: BuildKit is enabled but the buildx component is missing or broken.

       Install the buildx component to build images with BuildKit:

       https://docs.docker.com/go/buildx/

script returned exit code 1
```

test/xbgpu/test_engine.py
```diff
@@ -565,11 +540,23 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+            #   middle of an accumulation.
+            # - More accurately, it forces the first accumulation (to be
+            #   processed) to only contain one batch of data. This ensures
+            #   we test that output dumps are aligned correctly, despite
+            #   the first data processed not being on an accumulation
+            #   boundary.
+            heap_acc_thresholds = [corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]
+            first_batch_index = 12 * np.product(heap_acc_thresholds)
+            batch_start_index = first_batch_index - HEAPS_PER_FENGINE_PER_CHUNK
+            batch_end_index = first_batch_index + np.product(heap_acc_thresholds)
```
james-smith-za, line 558: I'd suggest just a bit of a style tweak, and a comment to clarify the selection of the arbitrary constant (12).

```suggestion
            n_heaps = np.product([corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs])
            first_batch_index = 12 * n_heaps  # Just somewhere arbitrarily that isn't zero
            batch_start_index = first_batch_index - HEAPS_PER_FENGINE_PER_CHUNK
            batch_end_index = first_batch_index + n_heaps
```

james-smith-za (approved):
I like it. One suggestion to make things (I think) a bit more elegant, but overall I am happy.

test/xbgpu/test_engine.py
bmerry, line 558: It's a bit confusing to have variables called both `first_batch_index` and `batch_start_index` that are different. Since `first_batch_index` doesn't seem be used again, how about something like this (can also be combined with James's suggestions):
```suggestion
            batch_start_index = 12 * np.product(heap_acc_thresholds)
            batch_end_index = batch_start_index + np.product(heap_acc_thresholds)
            # Add an extra batch before the first full accumulation
            batch_start_index -= HEAPS_PER_FENGINE_PER_CHUNK
```

test/xbgpu/test_engine.py
```diff
@@ -565,11 +540,23 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+            #   in this test dictates the first received batch to be in the
+            #   middle of an accumulation.
+            # - More accurately, it forces the first accumulation (to be
+            #   processed) to only contain one batch of data. This ensures
+            #   we test that output dumps are aligned correctly, despite
+            #   the first data processed not being on an accumulation
+            #   boundary.
+            heap_acc_thresholds = [corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]
+            first_batch_index = 12 * np.product(heap_acc_thresholds)
+            batch_start_index = first_batch_index - HEAPS_PER_FENGINE_PER_CHUNK
```
bmerry, line 557: This doesn't match the comment, which says "one batch of data". Will the test even do the right thing if HEAPS_PER_FENGINE_PER_CHUNK is >= one of the heap accumulation thresholds?

bmerry (changes requested):

test/xbgpu/test_engine.py
amishatishpatel: Ah, yes, well, this was my original observation and I didn't really know whether to stick or twist with 
* adding logic to prevent that from happening, and/or
* making a note that we've chosen values ~carefully to ensure this does not happen.

Any recommendations/suggestions/requests? I could probably do something at the top of the test like
* `assert heap_acc_threshold > HEAPS_PER_FENGINE_PER_CHUNK` for each threshold?

test/xbgpu/test_engine.py
amishatishpatel: Reading your collective suggestions here, my perceived awkwardness if I'd left `first_batch_index` out is unfounded.

Will update accordingly.

test/xbgpu/test_engine.py
amishatishpatel: Explicitly requested a re-review from you mostly to catch your attention to my reply above.

james-smith-za: I'd be happy with an assert statement such as you propose, @amishatishpatel, perhaps accompanied by a comment explaining that we haven't verified the logic of the test for if that constraint isn't met.

bmerry: What is the reason for using HEAPS_PER_FENGINE_PER_CHUNK here instead of 1?

amishatishpatel: > What is the reason for using HEAPS_PER_FENGINE_PER_CHUNK [2] here instead of 1?

My guess is that it's to be explicit that an entire Chunk went missing.
- Previously the `heap_accumulation_threshold` was set to 4,
- So the comment of "testing `if the first received batch is from the **middle** of an accumulation`" held true because 2 heaps (1 Chunk) were halfway to a total of 4 heaps (2 Chunks) required for an accumulation.

I figured I'd do a little `git blame` to see where that came from, and wouldn't you know [the commit](https://github.com/ska-sa/katgpucbf/commit/db1bae921e4b3b2be9ac3cd9e2be661e543c4139) was made by one @bmerry. I gather your reasoning was similar?

Either way, I think I'm going to do the assert statement and comment to be more explicit about this bit of code.

bmerry: Hoisted by my own petard!

> Either way, I think I'm going to do the assert statement and comment to be more explicit about this bit of code.

Sounds sensible.

test/xbgpu/test_engine.py
```diff
@@ -553,9 +553,11 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
             #   the first data processed not being on an accumulation
             #   boundary.
             heap_acc_thresholds = [corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]
-            first_batch_index = 12 * np.product(heap_acc_thresholds)
-            batch_start_index = first_batch_index - HEAPS_PER_FENGINE_PER_CHUNK
-            batch_end_index = first_batch_index + np.product(heap_acc_thresholds)
+            n_heaps = np.product(heap_acc_thresholds)
+            batch_start_index = 12 * n_heaps  # Somewhere arbitrary that isn't zero
+            batch_end_index = batch_start_index + n_heaps
+            # Add an extra batch before the first full accumulation
```
bmerry, line 559:
```suggestion
            # Add an extra chunk before the first full accumulation
```

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -501,6 +501,17 @@ async def test_xengine_end_to_end(
...
+        # NOTE: `HEAPS_PER_FENGINE_PER_CHUNK` and the `heap_accumulation_threshold`s
+        # are chosen carefully for this test. We simulate the first accumulation
+        # being incomplete by ensuring there is only one chunk of data present.
+        # In doing so, we must ensure `HEAPS_PER_FENGINE_PER_CHUNK` is smaller
+        # than either pipeline's `heap_accumulation_threshold`. The logic of the
+        # test has *not* been verified if that constraint isn't met, it merely
+        # `assert`'s.
+        heap_acc_thresholds = [corrprod_output.heap_accumulation_threshold for corrprod_output in corrprod_outputs]
+        for heap_accumulation_threshold in heap_acc_thresholds:
+            assert heap_accumulation_threshold > HEAPS_PER_FENGINE_PER_CHUNK
```
bmerry, line 513: Could also do
```suggestion
        assert min(heap_accumulation_thresholds) > HEAPS_PER_FENGINE_PER_CHUNK
```
but that's just bikeshedding - feel free to keep it as is if it seems easier to understand that way.

bmerry (approved):

test/xbgpu/test_engine.py
amishatishpatel: Argh, too right. I think your suggestion is easier and/or exactly what I had in mind (just couldn't think of the syntax).

amishatishpatel:
Will wait for Jenkins before merging.

james-smith-za:
> Will wait for Jenkins before merging.

Please go ahead and merge. The failures seem to be flake8-related. A remote server that seems to be down at the moment.

PR #627, opened 2023-09-14
Add benchmark code for fgpu

bmerry:
Benchmarking for the whole of fgpu, using a noisy binary search to find the critical ADC sample rate at which things start to fall apart.

This will need some tweaking after I've finishing running calibration experiments, but those changes should be fairly isolated so this might as well get some reviewing now.

Not quite sure what to do about requirements; I could add another extra to setup.cfg and incorporate them into requirements-dev.txt, but on the other hand I'm not expecting this to be run frequently or by many people.

- [ ] If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`

Contributes to NGC-1024.


doc/benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+for lower core counts, the benchmark script may need to be tuned to provide
+sensible placement.
+
+The client machine will also need a few Python packages installed, including
+:mod:`aiohttp`, :mod:`asyncssh`, :mod:`numpy` and :mod:`prometheus_client`
+(it's easiest to just run the script then install whatever packages are
+missing).
+
+On the client machine, you will need to create a TOML_ file describing the
+servers you want to use. Each table in the TOML file describes one server. You
```
james-smith-za, line 36: TIL that what I always thought were called "sections" in TOML are actually called "tables".

doc/benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+
+The assignment of threads to processor cores on the servers has been
+optimised for AMD Epyc servers with at least 16 cores; for other servers or
+for lower core counts, the benchmark script may need to be tuned to provide
+sensible placement.
+
+The client machine will also need a few Python packages installed, including
+:mod:`aiohttp`, :mod:`asyncssh`, :mod:`numpy` and :mod:`prometheus_client`
+(it's easiest to just run the script then install whatever packages are
+missing).
```
james-smith-za, line 33: Is this in lieu of a currently-absent requirements file?

doc/benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+   interfaces = ['enp193s0f0np0', 'enp193s0f1np1']
+
+   [fgpu]
+   hostname = 'server02.domain'
+   username = 'myusername'
+   interfaces = ['enp193s0f0np0', 'enp193s0f1np1']
+
+.. _TOML: https://toml.io/
+
+The `interfaces` tables list the names of the ibverbs-capable network
```
james-smith-za, line 54: nit-pick: I think this would be a key/value pair, not a table.

doc/benchmarking.rst
```diff
@@ -0,0 +1,160 @@
...
+which every candidate rate is tested a large number of times and the fraction
+of successes is printed. This does not automatically feed this information
+back into the (hard-coded) model, but it provides some data to guide the
+developer in writing or updating the model.
+
+.. option:: --calibrate
+
+   Run the calibration mode instead of the usual search
+
+.. option:: --calibrate-repeat
```
james-smith-za, line 154:
```suggestion
.. option:: --calibrate-repeat  <N>
```
?

james-smith-za (approved):
I have a couple of minor nit-picks which you can merge after fixing, I don't think I'd need to re-review.

The overall logic seems sound (though I say that as someone not particularly credentialed in computer science). It would be nice to see some results, but I guess maybe the JIRA ticket would be the best place to store that.

doc/benchmarking.rst
bmerry: Yes.

doc/benchmarking.rst
bmerry: I was referring to the value (`['enp193s0f0np0', 'enp193s0f1np1']`), but actually that isn't a table - it's an array. Will fix.

doc/benchmarking.rst
bmerry: Good catch.

doc/benchmarking.rst
james-smith-za: I wouldn't be opposed to adding another `extra` to `setup.cfg` but I don't think it's necessary to roll them into `requirements-dev.txt` since as you point out this will seldom be used.

But leaving it as you've got it here is probably also fine.

bmerry: I think I'm going to add a separate requirements.in/txt in scratch/benchmarks, similar to qualification/requirements.txt. I won't add it to setup.cfg because these benchmarks are standalone from the katgpucbf package (unlike qualification, which imports some bits of it).

james-smith-za (approved):

PR #631, opened 2023-09-21
Add --dst-buffer option to fgpu

bmerry:
This complements the --src-buffer option. The default value matches the previous default, which came from spead2, but it is now more explicitly under the control of katgpucbf and the user.

- [x] (n/a) If modules are added/removed: use sphinx-apidoc to update files in `doc/`
- [x] Update spead2 and remove the TODO, once https://github.com/ska-sa/spead2/pull/282 is released

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -1093,7 +1093,7 @@ class Engine(aiokatcp.DeviceServer):
     src_packet_samples
         The number of samples per digitiser packet.
     src_buffer
-        The size of the network receive buffer (per polarisation).
+        The size of the network receive buffer.
     dst_interface
         IP addresses of the network devices to use for output.
     dst_ttl
@@ -1108,6 +1108,8 @@ class Engine(aiokatcp.DeviceServer):
     dst_comp_vector
         Completion vector for transmission, or -1 for polling.
         See :class:`spead2.send.UdpIbvConfig` for further information.
+    dst_buffer
+        Size of the network send buffer.
     outputs
         Output streams to generate. At present this must be a single
         WidebandOutput.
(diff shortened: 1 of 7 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -1093,7 +1093,7 @@ class Engine(aiokatcp.DeviceServer):
     src_packet_samples
         The number of samples per digitiser packet.
     src_buffer
-        The size of the network receive buffer (per polarisation).
+        The size of the network receive buffer.
```
bmerry, line 1096: The actual behaviour change occurred when stream-per-pol was replaced by a single stream, but the documentation was missed at the time.

src/katgpucbf/fgpu/send.py
```diff
@@ -252,14 +253,22 @@ def make_streams(
...
         streams = [
-            spead2.send.asyncio.UdpStream(
-                thread_pool, [(ep.host, ep.port) for ep in endpoints], config, ttl=ttl, interface_address=interface
+            spead2.send.asyncio.UdpStream(  # type: ignore
+                thread_pool,
+                [(ep.host, ep.port) for ep in endpoints],
+                config,
+                ttl=ttl,
+                interface_address=interface,
+                buffer_size=buffer // len(interfaces),
```
james-smith-za, line 271: So the buffer is shared between interfaces? I guess in most cases `len(interfaces)` will be 1, but in the high-bandwidth case it may be 2.

james-smith-za (approved):

src/katgpucbf/fgpu/send.py
bmerry: Yes. This makes it consistent with `--src-buffer`. On the other hand, I guess that means that with two interfaces, the buffer is smaller than it was. I think I'll just increase the default: the reason I added this support was there was some inconclusive evidence that increasing the buffer size might help, and I wanted to be able to test that easily.

james-smith-za (approved):

PR #634, opened 2023-09-29
fgpu compute: 4-bit output support

bmerry:
Support packed 4-bit output from the compute part of the F-engine (not yet supported by the rest of the engine). It's implemented with the real part in the high bits and imaginary in the low bits of each byte. That's consistent with SPEAD's interpretation of arrays with sub-byte types (big-endian), but I think the opposite of what TCC wants. We can cross that bridge later.

To achieve consistency between the modes, the trailing COMPLEX axis is removed from the arrays, and instead 8-bit mode uses a struct dtype with real and imaginary fields. The 4-bit mode uses an opaque 1-byte dtype instead.

Closes NGC-1099 and NGC-1100.


bmerry:
Off the top of my head I don't think there are any updates needed in the design doc, but I've left a checklist item unchecked for my to verify that.

src/katgpucbf/utils.py
```diff
@@ -322,3 +323,20 @@ def unix_to_adc(self, timestamp: float) -> float:
...
+    Parameters
+    ----------
+    bits
+        Number of bits in each real component
+    """
+    assert bits in {4, 8, 16, 32}
+    if bits == 4:
+        # 1-byte, packed with real high, imaginary low. Using void rather
+        # than e.g. uint8 avoids accidentally doing arithmetic on it.
+        return np.dtype("V1")
```
amishatishpatel, line 340: Did a shallow dive into `numpy.void` and found [this excellent explanation](https://stackoverflow.com/a/25247533), although I still must ask: What do you mean by [emphasis added]
> Using void rather than e.g. uint8 **avoids accidentally doing arithmetic on it**.

src/katgpucbf/fgpu/send.py
```diff
@@ -61,7 +63,7 @@ class Frame:
     timestamp
         Zero-dimensional array of dtype ``>u8`` holding the timestamp.
     data
-        Payload data for the frame, of shape (channels, spectra_per_heap, N_POLS, COMPLEX).
```
amishatishpatel, line 64: Does the `heap_data_shape` in `make_descriptor_heap` also need to be updated? Line 299.

src/katgpucbf/fgpu/send.py
```diff
@@ -234,10 +236,11 @@ def make_streams(
     differ only in the network interface used (there is one per interface).
     Thus, they can be used interchangeably for load-balancing purposes.
     """
-    dtype = chunks[0].data.dtype
+    dtype = chunks[0].data.dtype  # Type for each complex value
+    bandwidth = adc_sample_rate * 0.5
```
amishatishpatel, line 240: I'm not sure how else to ask this, but
* Where did this come from?

src/katgpucbf/fgpu/kernels/postproc.mako
```diff
@@ -22,8 +22,21 @@
...
+typedef uchar2 qjones;
+#define QMAX 7.0f
+#else
+typedef char2 qcomplex;
+typedef char4 qjones;
+#define QMAX 127.0f
+#endif
 
-<%transpose:transpose_data_class class_name="scratch_t" type="char4" block="${block}" vtx="${vtx}" vty="${vty}"/>
+<%transpose:transpose_data_class class_name="scratch_t" type="qjones" block="${block}" vtx="${vtx}" vty="${vty}"/>
```
amishatishpatel, line 39: Not sure I'm going to be much use in this domain, but worth asking where I can find any kind of pointers/documentation on understanding this syntax (and it's uses)? (I assume it's Mako? Naive Google didn't turn up anything.)

test/fgpu/test_postproc.py
```diff
@@ -64,46 +65,51 @@ def postproc_host_pol(
     phase = np.exp(m2jpi * fine_delay[:, np.newaxis] * channel_idx / (2 * channels) + 1j * fringe_phase[:, np.newaxis])
     assert phase.dtype == np.complex64
     # Apply delay, phase and gain
+    corrected: np.ndarray
```
amishatishpatel, line 68:
```bash
test/fgpu/test_postproc.py:70: error: Incompatible types in assignment
(expression has type "ndarray[Any, dtype[floating[_32Bit]]]", 
variable has type "ndarray[Any, dtype[bool_]]")  [assignment]

test/fgpu/test_postproc.py:80: error: Incompatible types in assignment 
(expression has type "ndarray[Any, dtype[complexfloating[_32Bit, _32Bit]]]", 
variable has type "ndarray[Any, dtype[bool_]]")  [assignment]
```

I see.

test/fgpu/test_postproc.py
```diff
@@ -113,6 +119,18 @@ def postproc_host(
     return np.stack(out, axis=3), np.stack(saturated, axis=1)
 
 
+def unpack_complex(data: np.ndarray) -> np.ndarray:
+    """Unpack array of Gaussian integers to complex dtype."""
+    if data.dtype.itemsize == 1:
+        # It's 4-bit packed. We assume that >> will sign extend on signed types
+        real = data.view(np.int8) >> 4
+        imag = data.view(np.int8) << 4 >> 4
```
amishatishpatel, line 127: Hang on, what is this sorcery? I gather it has to do with 
> We assume that >> will sign extend on signed types

Pasting the PR blurb for ease of reference, although I can't quite figure out how that influences the one-liner `<< 4 >> 4`
> It's implemented with the real part in the high bits and imaginary in the low bits of each byte.
That's consistent with SPEAD's interpretation of arrays with sub-byte types (big-endian)

amishatishpatel (commented):
Mostly comments for clarification, maybe one or two small changes therein.

src/katgpucbf/utils.py
bmerry: Actually that link might be confusing you - it's a different use case for `np.void` (which TBH I wasn't even aware counted as "void"). In this case I want the type to be *opaque*. Consider the following:
```
>>> a = np.array([10, 15, -17], np.int8)
>>> np.abs(a)
array([10, 15, 17], dtype=int8)
```
That makes sense if the bytes in `a` really are integers, but if they're packed 4-bit complex numbers, the result is nonsense, because you're not taking the abs of each component. By using an opaque type, we can ensure that such an operation fails instead of producing nonsense, so that accidentally trying to do things the naive way blows up in your face instead of causing subtle bugs:
```
>>> b = a.view("V1")
>>> b
array([b'\x0A', b'\x0F', b'\xEF'], dtype='|V1')
>>> np.abs(b)
UFuncTypeError: ufunc 'absolute' did not contain a loop with signature matching types <class 'numpy.dtypes.VoidDType'> -> None
```

src/katgpucbf/fgpu/send.py
bmerry: No, because that's specified by the ICD, and it uses int8 as the dtype rather than the (real, imag) struct. SPEAD supports sub-byte types (like int4) so it's not necessary to change anything there.

src/katgpucbf/fgpu/send.py
bmerry: Let's start with wideband. The formula for rate should always have been bandwidth * N_POLS * COMPLEX * bytes-per-real-component. It just so happens that `bandwidth * COMPLEX` equals adc_sample_rate, which is why it was correct before when the `dtype` was per-real-component. The `dtype` is now per-complex-value, so COMPLEX is now absorbed into `dtype.itemsize`.

For narrowband, this is actually wrong now (will fix). It should be
```suggestion
    bandwidth = adc_sample_rate * 0.5 * send_rate_factor
```
and then `send_rate_factor` will no longer appear in computing `rate`.

src/katgpucbf/fgpu/kernels/postproc.mako
bmerry: It's Mako. For the transposition macro specifically, https://katsdpsigproc.readthedocs.io/en/latest/user/macros.html is the reference.

test/fgpu/test_postproc.py
bmerry: Yeah I should add a comment to explain that this is to work around wonky type inference. I haven't figured out there the `bool_` is coming from.

test/fgpu/test_postproc.py
bmerry: Let's say that the value is 0bABCDEFGH i.e. the real part is 0bABCD and the imaginary part is 0bEFGH. Then `x << 4` is 0bEFGH0000 (the top 4 bits overflow and are discarded) and then x >> 4 is 0bEEEEEFGH, which is the imaginary part sign-extended to 8 bits.

src/katgpucbf/utils.py
amishatishpatel: Aha, I see what you mean. Excellent choice.

src/katgpucbf/fgpu/send.py
amishatishpatel: Too right, too right.

src/katgpucbf/fgpu/send.py
amishatishpatel: Excellent, I see the latest commits address my counter-question of 
> Will `send_rate_factor` be removed?

test/fgpu/test_postproc.py
amishatishpatel: *Interesting*, I (clearly) was not aware of the concept of "sign extension", appreciate the explanation.

amishatishpatel (approved):
I think I'll upskill on Mako separate from this PR. Still, appreciate pointing me to the relevant docs.

PR #638, opened 2023-10-04
Improvements to benchmarking code

bmerry:
Lots of improvements to the benchmarking script:

- Use a logistic curve to estimate error rates, rather than the previous piecewise linear model
- Fit the model to calibration data, instead of just guessing (but still a guess for n==2; will revisit later)
- Make 1024 channels the default, since it seems to be the worst case
- Adjust src buffer size to the number of engines (to keep total buffer size constant)
- Make calibration mode more robust, by making the rate the inner loop, so that machine load (e.g. installing updated packages) gets spread across the rates rather than all applying to a single rate
- Use 8 threads for n==1, for better performance
- Support n==2
- Always start the dsim first, so that fgpu starts with data already flowing (to ensure it can handle it)
- Have two verbosity levels, instead of just one
- Add assorted command-line options for better configurability

- [x] Calibrate the n==2 case

Relates to NGC-1024.


bmerry:
I've left the pre-commit checklist item open because I should check whether statsmodels or pandas have type hints.

bmerry:
@ludwigschwardt you might find this interesting purely for the logistic regression code in fit.py.

amishatishpatel:
> @ludwigschwardt you might find this interesting purely for the logistic regression code in fit.py.

I fear I might not be too helpful/useful in reviewing this, more in-depth maths/calculations. Not that I don't trust you, Bruce; would certainly appreciate an extra set of eyes on those updates.

amishatishpatel (commented):
Just a top-level comment before explicit approval (maybe one or two in the review itself).
* I've gone through everything I can make sense of,
* i.e. I haven't meaningfully engaged with `fit.py`.
* To that end, do you think it would be helpful/we would benefit from you running through a brief demo of the general process you go through? (When using these scripts).

Only if you have time, otherwise we'll wait to hear from Ludwig before approving (?).

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -390,13 +426,14 @@ async def main():  # noqa: D103
     parser.add_argument("--servers", type=str, default="servers.toml", help="Server description file [%(default)s]")
     parser.add_argument("--dsim-server", type=str, default="dsim", help="Server on which to run dsims [%(default)s]")
     parser.add_argument("--fgpu-server", type=str, default="fgpu", help="Server on which to run fgpu [%(default)s]")
-    parser.add_argument("-v", "--verbose", action="store_true", help="Emit stdout/stderr [no]")
+    parser.add_argument("-v", "--verbose", action="count", default=0, help="Increase verbosity [no]")
     parser.add_argument(
         "--calibrate", action="store_true", help="Run at multiple rates to calibrate expectations [%(default)s]"
```
ludwigschwardt, line 431: What is the difference between calibration and straight search? When you say that you expect to take 100 hours to do the search, how long does the calibration step take? I'm guessing 40-60 rates * 100 trials per rate * 1 minute = also about 100 hours.

scratch/benchmarks/fit.py
```diff
@@ -0,0 +1,46 @@
...
+    ind = []
+    for rate, row in orig_df.iterrows():
+        rates.extend([rate] * row["trials"])
+        ind.extend([1] * row["successes"])
+        ind.extend([0] * (row["trials"] - row["successes"]))
+    df = pd.DataFrame({"ind": ind, "rate": rates})
+
+    y, X = dmatrices("ind ~ np.log(rate)", data=df, return_type="dataframe")  # noqa: N806
+    model = sm.Logit(y, X)
+    result = model.fit()
```
ludwigschwardt, line 31: Impressively compact! :-)

ludwigschwardt (approved):
Except for general questions, I'm happy with the model fitting part. I did not check the noisy binary search. I'm still tempted to try a Gaussian process-based active search 😁

scratch/benchmarks/fgpu_calibration/calibrate-n1.txt
```diff
@@ -0,0 +1,41 @@
+8000000000.0 100 100
```
ludwigschwardt, line 1: Are these distribution functions independent of all the various parameters except `n`?

bmerry: They're almost certainly not. What I'm hoping is that the slope is mostly independent while the offset is dependent. In other words, changing parameters will change the achievable throughput, but not the uncertainty. Unfortunately I don't think it's feasible (in time spent running calibrations) to validate that assumption.

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -322,14 +344,23 @@ async def compare(adc_sample_rate: float) -> bool:
...
 
-    # Compute error estimate. This is a heuristic based on gut feel rather
-    # than measurement. Assume that results are 1-NOISE reliable when at least
-    # FUZZ away from the critical value, and vary logarithmically in between.
+    # Compute error estimate. The model is a logistic regression on the
+    # log of the sample rate (log is used mainly to make the slope invariant
+    # to the scale of the rates, rather than for the shape).
+    # The magic numbers are determined from fit.py.
+    slope = {
+        1: -342.212919,
```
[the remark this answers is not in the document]
bmerry: It's the coefficient in the linear part of the generalised linear model. You can see how it's applied at line 360.

scratch/benchmarks/benchmark_fgpu.py
bmerry: The 100 hours is for the calibration, which produces the calibrate-n*.txt files. The actual search is capped at 40 trials by default (the actual number depends on the `--interval` parameter controlling how tight a confidence interval you want).

bmerry (commented):
@ludwigschwardt do you have time to review the noisy binary search as well? It looks like it's doing sensible things when I run it, but it would be good to have some more eyes on the Bayesian update and the entropy calculation, for example.

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -322,14 +344,23 @@ async def compare(adc_sample_rate: float) -> bool:
...
+        2: -173.264274,
+        4: -582.668296,
+    }[args.n]
     mid_rates = 0.5 * (rates[:-1] + rates[1:])  # Rates in the middle of the intervals
     mid_rates = np.r_[args.low, mid_rates, args.high]
     l_rates = np.log(rates)[:, np.newaxis]
     l_mid_rates = np.log(mid_rates)[np.newaxis, :]
-    noise = np.clip(0.5 + (l_rates - l_mid_rates) / np.log1p(FUZZ) * (0.5 - NOISE), NOISE, 1 - NOISE)
+    noise = expit((l_mid_rates - l_rates) * slope)
```
ludwigschwardt, line 360: I'm not very clear on this noise term... I thought the expit would represent the probability of success at various (log) data rates, so that for a certain rate r0 we have P = 0.5 (the transition rate) and for r << r0 we have near certain success and r >> r0 we have near certain failure.

So why are you interested in the ratio of the "mid" rate and the rate below (?) it and passing that to expit instead?

bmerry: l_rates represents the rates at which we might choose to run trials. But the probability distribution is discretised as the probabilities that the transition rate is in each interval between two of those trial rates. Rather than mess around with integration, I'm just using the mid-point of each interval as a proxy for the whole interval.

Note that l_rates is a column vector and l_mid_rates is a row vector, so you're getting broadcasting here. In your notation, noise[i, j] is the probability of success (or failure, I forget which offhand) if log(r0) is l_mid_rates[j] and log(r) is l_rates[i].

I put everything into log-space just to make it scale-invariant e.g. the `slope` term shouldn't need to change if you run this on a system that's twice as fast or if you change the units of rate from Hz to MHz.

ludwigschwardt: Thanks, I missed the broadcast!

ludwigschwardt:
> @ludwigschwardt do you have time to review the noisy binary search as well? It looks like it's doing sensible things when I run it, but it would be good to have some more eyes on the Bayesian update and the entropy calculation, for example.

I can see the update step in the paper but I missed the entropy calculation... Specifically why cumsums are involved in noise_search.py, line 123.

bmerry:
> I can see the update step in the paper but I missed the entropy calculation... Specifically why cumsums are involved in noise_search.py, line 123.

In the paper they always do the next trial at a position where the current distribution gives it close to 50% chance of success. They then prove that this drives down entropy at a certain rate. I decided to just make a query that minimises *expected* entropy (I can't recall why, but maybe something to do with having a more general prior).

The cumsum is to estimate, for each possible query value, the probability of success/failure for that query (using the current distribution). But, now that you point it out, the calculation is incorrect: it's actually the probability that the query value is to the left/right of the transition point, but should take into account the prior as well. So I'll work on that.

bmerry:
I've fixed the calculation of expected entropy.

ludwigschwardt (approved):
Great, it's also more understandable now!

PR #639, opened 2023-10-04
Add types-setuptools to pre-commit config

bmerry:
I seem to randomly get errors about it being missing when I try to commit stuff. It was previously deleted because we're no longer using pkg-resources directly, but it is used by redis.


amishatishpatel (approved):
I see, I see.

PR #641, opened 2023-10-06
fgpu full support for 4-bit output types

bmerry:
Support 4-bit output from the command-line. This is a little bit of plumbing, with most of the PR being the addition of test code for fgpu.send (previously not tested in isolation).

Closes NGC-1101.

### DIFF
--- a/test/fgpu/test_send.py
+++ b/test/fgpu/test_send.py
@@ -189,13 +189,7 @@ async def test_send(
     sensors: aiokatcp.SensorSet,
     time_converter: TimeConverter,
 ) -> None:
-    """Test sending data via the :class:`katgpucbf.fgpu.send.Chunk` interface.
-
-    .. todo::
-
-       This is a very limited test that does not cover sensors, metrics or
-       missing data.
-    """
+    """Test sending data via the :class:`katgpucbf.fgpu.send.Chunk` interface."""
     # Send descriptors to all the streams
     descriptor_heap = make_descriptor_heap(
         channels_per_substream=N_CHANNELS // N_SUBSTREAMS,
--- a/test/fgpu/test_send.py
+++ b/test/fgpu/test_send.py
@@ -17,6 +17,7 @@
 """Unit tests for :mod:`katgpucbf.fgpu.send`."""
 
 import asyncio
+from typing import Sequence
 from unittest import mock
 
 import aiokatcp
@@ -41,7 +42,6 @@ N_CHANNELS = 1024
 N_SPECTRA_PER_HEAP = 32  # Small to make the test fast
 SPECTRA_SAMPLES = 2 * N_CHANNELS
 FENG_ID = 3
-INTERFACES = ["10.0.0.1", "10.0.0.2"]
 NAME = "foo"
 
 
@@ -50,15 +50,21 @@ def sample_bits(request) -> int:
     return request.param
 
 
+@pytest.fixture(params=[1, 2])
+def interfaces(request) -> Sequence[str]:
+    # request.param gives number of interfaces to use
+    return ["10.0.0.1", "10.0.0.2"][: request.param]
+
+
 @pytest.fixture
 def time_converter() -> TimeConverter:
     return TimeConverter(1234567890.0, 1e9)
 
 
 @pytest.fixture
-def queues() -> dict[str, list[spead2.InprocQueue]]:
+def queues(interfaces: Sequence[str]) -> dict[str, list[spead2.InprocQueue]]:
     """Create in-process queues to connect the test code to the test."""
-    return {iface: [spead2.InprocQueue() for _ in range(N_SUBSTREAMS)] for iface in INTERFACES}
+    return {iface: [spead2.InprocQueue() for _ in range(N_SUBSTREAMS)] for iface in interfaces}
 
 
 @pytest.fixture
@@ -104,7 +110,7 @@ def send_streams(
             output_name=NAME,
             thread_pool=spead2.ThreadPool(1),
             endpoints=expected_endpoints,
-            interfaces=INTERFACES,
+            interfaces=list(queues.keys()),
             ttl=4,
             ibv=False,
             packet_payload=8192,
@@ -124,7 +130,7 @@ def recv_streams(queues: dict[str, list[spead2.InprocQueue]]) -> dict[str, list[
     """Create streams for receiving the transmitted data."""
     config = spead2.recv.StreamConfig()
     streams: dict[str, list[spead2.recv.asyncio.Stream]] = {}
-    for iface in INTERFACES:
+    for iface in queues.keys():
         streams[iface] = []
         for queue in queues[iface]:
             tp = spead2.ThreadPool(1)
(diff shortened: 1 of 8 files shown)


amishatishpatel:
I see the branch build failed on building docs/finding an intersphinx inventory again, but at least this time it was a different package (coverage, not flake8).

bmerry:
> I see the branch build failed on building docs/finding an intersphinx inventory again, but at least this time it was a different package (coverage, not flake8).

I think this time it is a permanent rather than a temporary failure, but still not related to this PR. I'll fix it in a separate one.

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+            packet_payload=8192,
+            comp_vector=0,
+            buffer=65536,
+            bandwidth=1e9,
+            send_rate_factor=1.05,
+            feng_id=FENG_ID,
+            num_ants=64,
+            n_data_heaps=N_CHUNKS * N_FRAMES * N_SUBSTREAMS,
+            chunks=chunks,
+        )
```
amishatishpatel, line 125: I realise it's probably a bit difficult to declare these naked constants elsewhere. Could I at least request a one/two-liner comment around "these are typical values for each parameter", perhaps even touch on why e.g. `1e9` was chosen for `bandwidth`?

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+
+@pytest.fixture(params=[4, 8])
+def sample_bits(request) -> int:
+    return request.param
+
+
+@pytest.fixture(params=[1, 2])
+def interfaces(request) -> Sequence[str]:
+    # request.param gives number of interfaces to use
+    return ["10.0.0.1", "10.0.0.2"][: request.param]
```
amishatishpatel, line 56: Again, I didn't know of this "declare-and-use-immediately" MO. Very slick.

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+@pytest.fixture
+def time_converter() -> TimeConverter:
+    return TimeConverter(1234567890.0, 1e9)
```
amishatishpatel, line 61: I see `1e9` used again here. Maybe you didn't want to declare `ADC_SAMPLE_RATE` at the top and reuse it for the `bandwidth` arg (I think we had a back-and-forth about that on a 4-bit PR recently).

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+    return sensors
+
+
+def _fill_random(data: np.ndarray, rng: np.random.Generator) -> None:
+    """Fill an array with random bytes.
+
+    The underlying bytes are filled, rather than assuming any particular dtype.
+    """
+    view = data.view(np.uint8)
+    view[:] = rng.integers(0, 256, view.shape, dtype=np.uint8)
```
amishatishpatel, line 164: Man, so neat.

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+    chunks: list[Chunk],
+    sensors: aiokatcp.SensorSet,
+    time_converter: TimeConverter,
+) -> None:
+    """Test sending data via the :class:`katgpucbf.fgpu.send.Chunk` interface.
+
+    .. todo::
+
+       This is a very limited test that does not cover sensors, metrics or
+       missing data.
```
amishatishpatel, line 197: Either
* this `todo` needs to be updated (Re: `# Check the sensors and Prometheus metrics` at the end here), or
* I don't actually know what it means to test those things.

Both can be true, I guess.

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+    # Knock out the first few frames, to test partial transmission
+    chunks[0].present[:skip_frames] = False
+    data = np.concatenate([chunk.data for chunk in chunks])
+    saturated = np.concatenate([chunk.saturated for chunk in chunks])
+    saturated = np.sum(saturated[skip_frames:], axis=0, dtype=np.uint64)
+
+    with PromDiff(namespace=METRIC_NAMESPACE) as prom_diff:
+        # Send all the chunks, without waiting for the first one to complete
+        # transmission (to ensure that the send streams have sufficient
+        # capacity).
```
amishatishpatel, line 231: More for my own understanding, how does this ~shotgun-send ensure 
> that the send streams have sufficient capacity [?]

Would some underlying e.g. spead2 code complain if that weren't the case?

test/fgpu/test_send.py
```diff
@@ -0,0 +1,285 @@
...
+    for i in range(N_SUBSTREAMS):
+        seen_frames = np.zeros(N_CHUNKS * N_FRAMES, bool)
+        # Only one of the two interfaces should be receiving any data for
+        # this substream, but it's simpler to just iterate over both and
+        # find the data wherever it happened to fall.
+        for iface in queues.keys():
+            recv_stream = recv_streams[iface][i]
+            ig = spead2.ItemGroup()
+            frame = skip_frames
+            async for heap in recv_stream:
```
amishatishpatel, line 252: We should probably move away from `heap = await stream.get()` and towards this `async for`. Certainly looks neater.

EDIT: Although, maybe it depends on the use case? 
* e.g. the `# Check that we received all the data we expected` is a good example for this usage.
* We don't necessarily want to "only receive as much data as we know we sent".

amishatishpatel (commented):
The odd comment and a question about a `todo`. Overall, very nice.

test/fgpu/test_send.py
```diff
@@ -0,0 +1,256 @@
...
+pytest_mark = pytest.mark.parametrize("bits", [4, 8])
+N_SUBSTREAMS = 16
+N_CHUNKS = 5
+N_FRAMES = 7  # frames per chunk
+N_CHANNELS = 1024
+N_SPECTRA_PER_HEAP = 32  # Small to make the test fast
+SPECTRA_SAMPLES = 2 * N_CHANNELS
+FENG_ID = 3
+INTERFACES = ["10.0.0.1", "10.0.0.2"]
+NAME = "foo"
```
james-smith-za, line 45: Bikeshed:
Since we're writing in Python, this should really be something like "spam".

test/fgpu/test_send.py
james-smith-za: It looks like you're right, Amish, I do see checks for sensors and metrics down at the bottom, and missing frames in the next commit (I'm looking at this commit-by-commit).
Probably the TODO needs to be updated.

james-smith-za (changes requested):
Nit-pick about the TODO that @amishatishpatel pointed out.

I'm otherwise happy, so once that has been addressed, we can merge.

test/fgpu/test_send.py
bmerry: Indeed, it is just a stale TODO.

test/fgpu/test_send.py
bmerry: There are already 22 other instances of "foo" in katgpucbf and 0 of "spam", so I'm sticking with "foo".

test/fgpu/test_send.py
bmerry: Good point, I should. It turns out there was a bug, in that here it is ADC_SAMPLE_RATE and elsewhere it is used as the bandwidth. I'll introduce a constant and fix the bug.

test/fgpu/test_send.py
bmerry: Yes.

test/fgpu/test_send.py
james-smith-za: Touche. I'll allow it this time. ;)

james-smith-za (approved):

amishatishpatel (approved):

PR #642, opened 2023-10-09
Fix URLs for coverage.py

bmerry:
It seems they no longer use /en/stable, only /en/latest.


amishatishpatel (approved):
How did you pick this up? From the failing Jenkins builds?

bmerry:
> How did you pick this up? From the failing Jenkins builds?

Yes.

PR #650, opened 2023-10-18
Add BPipeline infrastructure

amishatishpatel:
As the title suggests, this aims to put in place some of the minimum infrastructure required to
run a Beamformer. Comments on NGC-1011 (and related tickets) will perhaps provide more detailed
info on some decisions. 
* This PR only addresses support for a single `--beam`; although,
* I have tried to make provision for multiple beams/outputs where it didn't require too much renovation. 
* Furthermore, it is currently configured to only transmit zeroed data via its BSend.
* There is a corresponding `test_bsend.py` to test this specific functionality (in its current, zeroed guise).

Areas of concern/seeking feedback include (but are not limited to)
* `saturated` array configuration (specifically its `shape`),
* How to 'verify' I've provisioned enough in the actual `spead2...AsyncStream` for high speed data transmission (e.g. `max_heaps`),
* A method of handling `capture-enable` for the eventual multiple `--beam`s/substreams within the BSend,

I've left quite a few `TODOs` in, mostly to grab the reviewer's attention in case I don't mention it here.

More admin-related, I'm not entirely trusting of `sphinx-apidoc -o doc/ src/` as it didn't pick up/automatically
generate a `.rst` file for `katgpucbf.xbgpu.bsend` (when I ran it on `gareth`).
* I didn't want to commit the file created by hand, but let me know if I should do that anyway.

EDIT: Also unsure if I should start with a corresponding katsdpcontroller update just yet (which is why I left it unchecked and without an `(N/A)`).

- [ ] If modules are added/removed: use sphinx-apidoc to update files in `doc/`

Closes NGC-1011 and NGC-1139.

### DIFF
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -121,6 +121,12 @@ class Chunk:
         self._timestamp_step = timestamp_step
         self._timestamps = (np.arange(n_frames) * self._timestamp_step).astype(">u8")
 
+        # NOTE: The future indicates when it is safe to modify the chunk,
+        # i.e. it is not being transmitted. At construction there is nothing to
+        # wait for, so we mark it ready.
+        self.future = asyncio.get_running_loop().create_future()
+        self.future.set_result(None)
+
         self._frames = [
             Frame(
                 self._timestamps[i, ...],
@@ -137,8 +143,8 @@ class Chunk:
         Timestamp of the first heap.
 
         Setting this property updates the timestamps stored in all the heaps.
-        This should not be done while a previous call to :meth:`send` is still
-        in progress.
+        This should only be done after obtaining a new chunk to ensure its
+        :attr:`future` has completed.
         """
         return self._timestamp
 
@@ -157,8 +163,8 @@ class Chunk:
         """
         Transmit a chunk's heaps over a SPEAD stream.
 
-        The chunk holds a reference to the gathered future, which needs to be
-        awaited before using again.
+        This method returns immediately and sends the data asynchronously. Before
+        modifying the chunk, first await :attr:`future`.
 
         .. todo::
 
@@ -172,10 +178,11 @@ class Chunk:
                     send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
                 )
             # TODO: Update counters and sensor with chunk.saturation
-            return asyncio.gather(*send_futures)
+            self.future = asyncio.gather(*send_futures)
         else:
             # TODO: Is it necessary to handle this case?
-            return asyncio.create_task(send_stream.stream.async_flush())
+            self.future = asyncio.create_task(send_stream.stream.async_flush())
+        return self.future
 
 
 class BSend:
@@ -246,7 +253,7 @@ class BSend:
         self.tx_enabled = [tx_enabled] * len(outputs)
         self.n_beams = len(outputs)
 
-        self._free_chunks_queue: asyncio.Queue[Chunk] = asyncio.Queue()
+        self._chunks_queue: asyncio.Queue[Chunk] = asyncio.Queue()
         buffers: list[np.ndarray] = []
 
         send_shape = (heaps_per_fengine_per_chunk, self.n_beams, n_channels_per_substream, spectra_per_heap, COMPLEX)
@@ -261,7 +268,7 @@ class BSend:
                 channel_offset=channel_offset,
                 timestamp_step=timestamp_step,
             )
-            self._free_chunks_queue.put_nowait(chunk)
+            self._chunks_queue.put_nowait(chunk)
             buffers.append(chunk.data)
 
         # TODO: Follow suit with other Engines to calculate `rate`
@@ -317,27 +324,28 @@ class BSend:
         self.tx_enabled[stream_id] = enable
 
     async def get_free_chunk(self) -> Chunk:
+        """Obtain a :class:`.Chunk` for transmission.
+
+        Raises
+        ------
+        asyncio.CancelledError
+            If the chunk's future is cancelled at any point.
+        Exception
+            For any other unexpected error.
         """
-        Obtain a :class:`.Chunk` for transmission.
-
-        .. todo::
-
-            Is this entirely necessary? Left in because it felt a bit strange
-            to dig into the :class:`.BSend` from the :class:`.XBEngine`.
-        """
-        chunk = await self._free_chunks_queue.get()
-        return chunk
-
-    async def send_chunk(self, chunk: Chunk, time_converter: TimeConverter, sensors: SensorSet) -> None:
-        """Send a chunk's data and put it on the queue."""
+        chunk = await self._chunks_queue.get()
         try:
-            await chunk.send(self, time_converter, sensors)
+            await chunk.future
         except asyncio.CancelledError:
             pass
         except Exception as ex:
-            raise ex
-        finally:
-            self._free_chunks_queue.put_nowait(chunk)
+            raise Exception(ex.args)
+        return chunk
+
+    def send_chunk(self, chunk: Chunk, time_converter: TimeConverter, sensors: SensorSet) -> None:
+        """Send a chunk's data and put it on the :attr:`_chunks_queue`."""
+        chunk.send(self, time_converter, sensors)
+        self._chunks_queue.put_nowait(chunk)
 
     async def send_stop_heap(self) -> None:
         """Send a Stop Heap over the spead2 transport."""
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -18,6 +18,7 @@
 
 import asyncio
 import logging
+from math import ceil
 from typing import Callable, Final, Sequence
 
 import katsdpsigproc.accel as accel
@@ -214,7 +215,7 @@ class BSend:
         Number of SPEAD heaps from one F-engine in a single received Chunk.
     n_tx_items
         Number of :class:`Chunk` to create.
-    n_channels, n_channels_per_substream, spectra_per_heap, channel_offset
+    adc_sample_rate, n_channels, n_channels_per_substream, spectra_per_heap, channel_offset
         See :class:`.XBEngine` for further information.
     timestamp_step
         The timestamp step between successive heaps, as dictated by the XBEngine.
@@ -243,6 +244,7 @@ class BSend:
         n_channels: int,
         n_channels_per_substream: int,
         spectra_per_heap: int,
+        adc_sample_rate: float,
         timestamp_step: int,
         send_rate_factor: float,
         channel_offset: int,
@@ -277,13 +279,24 @@ class BSend:
             self._chunks_queue.put_nowait(chunk)
             buffers.append(chunk.data)
 
-        # TODO: Follow suit with other Engines to calculate `rate` below
+        # Multicast stream parameters
+        heap_payload_size_bytes = n_channels_per_substream * spectra_per_heap * COMPLEX * SEND_DTYPE.itemsize
+
+        # Transport-agnostic stream information
+        packets_per_heap = ceil(heap_payload_size_bytes / packet_payload)
+        packet_header_overhead_bytes = packets_per_heap * BSend.header_size
+
+        heap_interval = timestamp_step / adc_sample_rate
+        send_rate_bytes_per_second = (
+            (heap_payload_size_bytes + packet_header_overhead_bytes) / heap_interval * send_rate_factor * self.n_beams
+        )
+
         stream_config = spead2.send.StreamConfig(
             max_packet_size=packet_payload + BSend.header_size,
             # + 1 below for the descriptor per beam
             max_heaps=(n_tx_items * heaps_per_fengine_per_chunk + 1) * self.n_beams,
             rate_method=spead2.send.RateMethod.AUTO,
-            rate=0.0,  # TODO: Update to use `send_rate_bytes_per_second`, this sends as fast as possible
+            rate=send_rate_bytes_per_second,
         )
         self.stream = stream_factory(stream_config, buffers)
         # Set heap count sequence to allow a receiver to ingest multiple
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -173,14 +173,14 @@ class Chunk:
             Also update its relevant counters and sensor values.
         """
         if any(send_stream.tx_enabled):
+            send_futures = []
             for frame in self._frames:
                 heaps_to_send = [heap for heap, enabled in zip(frame.heaps, send_stream.tx_enabled) if enabled]
-                send_future = send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
+                send_futures.append(
+                    send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
+                )
             # TODO: Update counters and sensor with chunk.saturation
-            # NOTE: We use the last future returned above as the data is sent
-            # in-order. Once the last frame has been transmitted, the rest have
-            # too.
-            self.future = send_future
+            self.future = asyncio.gather(*send_futures)
         else:
             # TODO: Is it necessary to handle this case?
             self.future = asyncio.create_task(send_stream.stream.async_flush())
@@ -350,6 +350,10 @@ class BSend:
 
     async def get_free_chunk(self) -> Chunk:
         """Obtain a :class:`.Chunk` for transmission.
+
+        We await the chunk's :attr:`future` to be sure we are not overwriting
+        data that is still being transmitted. If sending failed, it is no
+        longer being transmitted, and therefore safe to return the chunk.
 
         Raises
         ------
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -23,11 +23,13 @@ import katsdpsigproc.accel as accel
 import numpy as np
 import spead2
 import spead2.send.asyncio
+from aiokatcp import SensorSet
 from katsdpsigproc.abc import AbstractContext
 from katsdptelstate.endpoint import Endpoint
 
 from .. import COMPLEX, DEFAULT_PACKET_PAYLOAD_BYTES
-from ..spead import BF_RAW_ID, FLAVOUR, FREQUENCY_ID, IMMEDIATE_FORMAT, TIMESTAMP_ID
+from ..spead import BF_RAW_ID, FLAVOUR, FREQUENCY_ID, IMMEDIATE_FORMAT, TIMESTAMP_ID, make_immediate
+from ..utils import TimeConverter
 from .output import BOutput
 
 # NOTE: ICD suggests `beng_out_bits_per_sample`,
@@ -35,30 +37,118 @@ from .output import BOutput
 SEND_DTYPE = np.dtype(np.int8)
 
 
-class Heap:
-    """Hold all data for a heap.
+class Frame:
+    """Hold all data for heaps with a single timestamp.
 
-    # TODO: This will likely need to be updated to the 'fgpu.send.Chunk' methodology
-            - Especially to dictate 'Frames' within the Chunk
+    It does not own its memory - the backing store is in :class:`Chunk`.
+
+    Parameters
+    ----------
+    timestamp
+        Zero-dimensional array of dtype ``>u8`` holding the timestamp
+    data
+        Payload data for the frame with shape (n_channels_per_substream, spectra_per_heap, COMPLEX)
+    saturated
+        Total number of complex samples that saturated during requantisation
+    n_substreams
+        # TODO Re-word this
+        Number of beams requiring data to be transmitted
     """
 
     def __init__(
         self,
-        context: AbstractContext,
-        n_channels_per_substream: int,
-        n_spectra_per_heap: int,
+        timestamp: np.ndarray,
+        data: np.ndarray,
+        saturated: np.ndarray,
+        *,
+        channel_offset: int,
+        n_substreams: int,
     ) -> None:
-        self.buffer: Final = accel.HostArray(
-            (n_channels_per_substream, n_spectra_per_heap, COMPLEX), SEND_DTYPE, context=context
-        )
+        self.heaps = []
+        self.data = data
+        self.saturated = saturated
+        for i in range(n_substreams):
+            heap = spead2.send.Heap(flavour=FLAVOUR)
+            heap.repeat_pointers = True
+            heap.add_item(make_immediate(FREQUENCY_ID, channel_offset))
+            heap.add_item(make_immediate(TIMESTAMP_ID, timestamp))
+            # TODO: Update `self.data` to slice off beam data for each substream
+            heap.add_item(
+                spead2.Item(
+                    BF_RAW_ID,
+                    "bf_raw",
+                    "",
+                    shape=data.shape,
+                    dtype=data.dtype,
+                    value=data,
+                )
+            )
+            self.heaps.append(spead2.send.HeapReference(heap, substream_index=i))
+
+
+class Chunk:
+    """An array of :class:`Heaps`."""
+
+    def __init__(
+        self,
+        data: np.ndarray,
+        saturated: np.ndarray,
+        *,
+        channel_offset: int,
+        n_substreams: int,
+    ) -> None:
+        # data.shape[0] should be some outer element indicating n_heaps
+        # to fill a Chunk.
+        # NOTE: I think this value == heaps-per-feng-per-chunk????
+        n_heaps = data.shape[0]
+        n_channels_per_substream = data.shape[1]
+        spectra_per_heap = data.shape[2]
+        self.data = data  # data should have all the dimensions required
+        self.saturated = saturated
+
+        # TODO: Clarify whether the timestamp_step is the same as the
+        #       `engine.rx_heap_timestamp_step`.
+        self._timestamp = 0
+        self._timestamp_step = n_channels_per_substream * spectra_per_heap
+        self._timestamps = (np.arange(n_heaps) * self._timestamp_step).astype(">u8")
+
+        # Need a future for each heap to be sent
+        self.futures: list[asyncio.Future] = [asyncio.get_running_loop().create_future() for _ in range(n_heaps)]
+        for future in self.futures:
+            future.set_result(None)
+
+        self._frames = [
+            Frame(
+                self._timestamps[i, ...],
+                data[i],
+                saturated[i],
+                channel_offset=channel_offset,
+                n_substreams=n_substreams,
+            )
+            for i in range(n_heaps)
+        ]
+
+    @property
+    def timestamp(self) -> int:
+        """
+        Timestamp of the first heap.
+
+        Setting this property updates the timestamps stored in all the heaps.
+        This should not be done while a previous call to :meth:`send` is still
+        in progress.
+        """
+        return self._timestamp
+
+    @timestamp.setter
+    def timestamp(self, value: int) -> None:
+        delta = value - self.timestamp
+        # TODO: This specifically needs some attention Re: Casting error
+        self._timestamps = self._timestamps + delta
+        self._timestamp = value
 
 
 class BSend:
-    """
-    Class for turning tied array channelised voltage products into SPEAD heaps.
-
-    NOTE: Each BSend shouldn't need its own *version* of a descriptor heap
-    """
+    """Class for turning tied array channelised voltage products into SPEAD heaps."""
 
     descriptor_heap: spead2.send.Heap
     header_size: Final[int] = 64
@@ -66,6 +156,8 @@ class BSend:
     def __init__(
         self,
         output: BOutput,
+        heaps_per_fengine_per_chunk: int,
+        n_tx_items: int,
         n_channels_per_substream: int,
         spectra_per_heap: int,
         send_rate_factor: float,
@@ -81,23 +173,34 @@ class BSend:
         # - The `send` function will need to keep track of "which substream
         #   is enabled/disabled" Re: capture-{start, stop} <stream_name>
         self.enabled_stream_ids: list[int] = []
+        self.tx_enabled = tx_enabled
 
-        self._heaps_queue: asyncio.Queue[Heap] = asyncio.Queue()
-        buffers: list[accel.HostArray] = []
+        self._chunks_queue: asyncio.Queue[Chunk] = asyncio.Queue()
+        buffers: list[np.ndarray] = []
 
         # `n_heaps_to_send` is actually used to dictate the amount of buffers (in XSend)
         # So perhaps we need to change the number of buffers to be range(send_free_queue.maxsize)
-        n_heaps_to_send = len(buffers) // spectra_per_heap
+        # n_heaps_to_send = len(buffers) // spectra_per_heap
 
-        for _ in range(n_heaps_to_send):
-            heap = Heap(context, n_channels_per_substream, spectra_per_heap)
-            self._heaps_queue.put_nowait(heap)
-            buffers.append(heap.buffer)
+        send_shape = (heaps_per_fengine_per_chunk, n_channels_per_substream, spectra_per_heap, COMPLEX)
+        for _ in range(n_tx_items):
+            chunk = Chunk(
+                accel.HostArray(send_shape, SEND_DTYPE, context=context),
+                accel.HostArray(
+                    (heaps_per_fengine_per_chunk,),
+                    np.uint32,
+                ),
+                channel_offset=channel_offset,
+                n_substreams=1,  # TODO: Update once single-beam is working
+            )
+            self._chunks_queue.put_nowait(chunk)
+            buffers.append(chunk.data)
 
         # Multicast stream parameters
         self.heap_payload_size_bytes = n_channels_per_substream * spectra_per_heap * COMPLEX * SEND_DTYPE.itemsize
         # Transport-agnostic stream information
         # Used in XSend to calculate `send_rate_bytes_per_second`, do we need it here?
+        # TODO: Scope to move this calculation into a helper in utils
         # packets_per_heap = math.ceil(self.heap_payload_size_bytes / packet_payload)
         # packet_header_overhead_bytes = packets_per_heap * BSend.header_size
 
@@ -107,7 +210,7 @@ class BSend:
             rate_method=spead2.send.RateMethod.AUTO,
             rate=0,  # TODO: Update to use `send_rate_bytes_per_second`, this sends as fast as possible
         )
-        self.source_stream = stream_factory(stream_config, buffers)
+        self.stream = stream_factory(stream_config, buffers)
 
         item_group = spead2.send.ItemGroup(flavour=FLAVOUR)
         item_group.add_item(
@@ -134,9 +237,28 @@ class BSend:
 
         self.descriptor_heap = item_group.get_heap(descriptors="all", data="none")
 
-    async def send_heap(self, heap: Heap) -> None:
-        """Take in a buffer and send it as a SPEAD heap."""
-        pass
+    async def send_chunk(
+        self,
+        chunk: Chunk,
+        time_converter: TimeConverter,
+        sensors: SensorSet,
+    ) -> None:
+        """
+        Transmit a Chunk's heaps.
+
+        .. todo::
+            Also update its relevant counters and sensor values.
+
+            This might need to be moved outside BSend because it looks
+            a bit awkward. Stream already owns the Chunks.
+        """
+        if self.tx_enabled:
+            chunk.futures = [
+                self.stream.async_send_heaps(frame.heaps, mode=spead2.send.GroupMode.ROUND_ROBIN)
+                for frame in chunk._frames
+            ]
+            # await asyncio.gather(*chunk.futures)
+            self._chunks_queue.put_nowait(chunk)
 
     def enable_substream(self, stream_id: int, enable: bool = True) -> None:
         """Enable/Disable a substream's data transmission.
@@ -155,6 +277,23 @@ class BSend:
             disabled.
         """
         pass
+
+    async def get_free_chunk(self) -> Chunk:
+        """Return a Chunk once it has completed its send futures."""
+        chunk = await self._chunks_queue.get()
+        # TODO: Do we need to do this if we gather the futures in `send_chunk`?
+        await asyncio.wait(chunk.futures)
+        return chunk
+
+    async def send_stop_heap(self) -> None:
+        """Send a Stop Heap over the spead2 transport."""
+        stop_heap = spead2.send.Heap(FLAVOUR)
+        stop_heap.add_end()
+        # Flush just to ensure that we don't overflow the stream's queue.
+        # It's a heavy-handed approach, but we don't care about performance
+        # during shutdown.
+        await self.stream.async_flush()
+        await self.stream.async_send_heap(stop_heap)
 
 
 def make_stream(
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -54,9 +54,6 @@ class Frame:
         with shape (n_beams,).
     channel_offset
         The first frequency channel processed.
-    n_substreams
-        Number of beams requiring data to be transmitted.
-        NOTE: Hardcoded to 1 in the BSend constructor, for now.
     """
 
     def __init__(
@@ -66,11 +63,11 @@ class Frame:
         saturated: np.ndarray,
         *,
         channel_offset: int,
-        n_substreams: int,
     ) -> None:
-        self.heaps = []
+        self.heaps: list[spead2.send.HeapReference] = []
         self.data = data
         self.saturated = saturated
+        n_substreams = saturated.shape[0]
         for i in range(n_substreams):
             heap = spead2.send.Heap(flavour=FLAVOUR)
             heap.repeat_pointers = True
@@ -100,7 +97,6 @@ class Chunk:
         n_beams, n_channels_per_substream, n_spectra_per_heap, COMPLEX) and
         dtype `SEND_DTYPE`.
     saturated
-        TODO: Clarify the shape
         Storage for saturation counts, with shape (n_frames, n_beams) and dtype
         uint32.
     channel_offset
@@ -119,18 +115,11 @@ class Chunk:
     ) -> None:
         n_frames = data.shape[0]
         self.data = data  # data should have all the dimensions required
-        n_substreams = saturated.shape[1]
         self.saturated = saturated
 
         self._timestamp = 0
         self._timestamp_step = timestamp_step
         self._timestamps = (np.arange(n_frames) * self._timestamp_step).astype(">u8")
-
-        # NOTE: The chunk's future should hold the future returned from
-        # gathering all :meth:`spead2.send.asyncio.AsyncStream.async_send_heaps`
-        # calls for each frame's heaps.
-        self.future = asyncio.get_running_loop().create_future()
-        self.future.set_result(None)
 
         self._frames = [
             Frame(
@@ -138,7 +127,6 @@ class Chunk:
                 data[i],
                 saturated[i],
                 channel_offset=channel_offset,
-                n_substreams=n_substreams,
             )
             for i in range(n_frames)
         ]
@@ -165,7 +153,7 @@ class Chunk:
         send_stream: "BSend",
         time_converter: TimeConverter,
         sensors: SensorSet,
-    ) -> None:
+    ) -> asyncio.Future:
         """
         Transmit a chunk's heaps over a SPEAD stream.
 
@@ -175,19 +163,18 @@ class Chunk:
         .. todo::
             Also update its relevant counters and sensor values.
         """
-        if send_stream.tx_enabled:
-            send_futures = [
-                send_stream.stream.async_send_heaps(frame.heaps, mode=spead2.send.GroupMode.ROUND_ROBIN)
-                for frame in self._frames
-            ]
-            self.future = asyncio.gather(*send_futures)
-            send_stream._chunks_queue.put_nowait(self)
+        if any(send_stream.tx_enabled):
+            send_futures = []
+            for frame in self._frames:
+                heaps_to_send = [heap for heap, enabled in zip(frame.heaps, send_stream.tx_enabled) if enabled]
+                send_futures.append(
+                    send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
+                )
             # TODO: Update counters and sensor with chunk.saturation
+            return asyncio.gather(*send_futures)
         else:
-            # :meth:`.BSend.get_free_chunk` still needs to await some future
-            # before returning a :class:`.Chunk`.
-            self.future = asyncio.create_task(send_stream.stream.async_flush())
-            send_stream._chunks_queue.put_nowait(self)
+            # TODO: Is it necessary to handle this case?
+            return asyncio.create_task(send_stream.stream.async_flush())
 
 
 class BSend:
@@ -198,8 +185,8 @@ class BSend:
 
     Parameters
     ----------
-    output
-        The BOutput containing the beam name and destination address.
+    outputs
+        List of :class:`.output.BOutput`.
     heaps_per_fengine_per_chunk
         Number of SPEAD heaps from one F-engine in a single received Chunk.
     n_tx_items
@@ -228,7 +215,7 @@ class BSend:
 
     def __init__(
         self,
-        output: BOutput,
+        outputs: Sequence[BOutput],
         heaps_per_fengine_per_chunk: int,
         n_tx_items: int,
         n_channels_per_substream: int,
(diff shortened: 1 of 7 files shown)


src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+                    np.uint32,
+                ),
+                channel_offset=channel_offset,
+                timestamp_step=timestamp_step,
+            )
+            self._chunks_queue.put_nowait(chunk)
+            buffers.append(chunk.data)
+
+        # Multicast stream parameters
+        self.heap_payload_size_bytes = n_channels_per_substream * spectra_per_heap * COMPLEX * SEND_DTYPE.itemsize
```
bmerry, line 270: Do you need this outside the constructor? If not, it doesn't seem to be necessary to store it in the object.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+            TIMESTAMP_ID,
+            "timestamp",
+            "Timestamp provided by the MeerKAT digitisers and scaled to the digitiser sampling rate.",
+            shape=[],
+            format=IMMEDIATE_FORMAT,
+        )
+        item_group.add_item(
+            BF_RAW_ID,
+            "bf_raw",
+            "Channelised complex data.",  # TODO: Straight from the ICD, is this sufficient?
```
bmerry, line 303: If you want something more detailed, look at https://docs.google.com/spreadsheets/d/1XojAI9O9pSSXN8vyb2T97Sd875YCWqie8NY8L02gA_I/edit#gid=0

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+    def enable_substream(self, stream_id: int = 0, enable: bool = True) -> None:
+        """Enable/Disable a substream's data transmission.
+
+        :class:`.BSend` operates as a large single stream with multiple
+        substreams. Each substream is its own data product and is required
+        to be enabled/disabled independently.
+
+        .. todo::
+            Update to actually carry out substream enable/disable once
+            multiple substreams are supported.
```
bmerry, line 319: I think you'll need a blank line after the todo to keep Sphinx happy.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+        Parameters
+        ----------
+        stream_id
+            ID of the substream, corresponds to the <beam-id><pol>
+            convention, e.g. stream_id 3 has a stream_name ending in <1x>.
```
bmerry, line 324: This comment is confusing. The names are provided by the user and can be anything. This whole module doesn't deal in names anyway - everything is by index.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+        enable
+            Boolean indicating whether the `stream_id` should be enabled or
+            disabled.
+        """
+        self.tx_enabled = enable
+
+    async def get_free_chunk(self) -> Chunk:
+        """Return a Chunk once it has completed its send futures."""
+        chunk = await self._chunks_queue.get()
+        await chunk.future
```
bmerry, line 334: I'm slightly nervous about what happens if this await is cancelled - the chunk will probably get dropped on the floor, and we won't wait for it to complete transmission. On the other hand, that should only happen during shutdown, at which point we no longer care.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -148,8 +154,15 @@ class Pipeline:
...
 
+    .. todo::
+        Update the `output` parameter to be a list of `Output`s once multiple
+        beams are supported. This will require some updates to the plumbing
+        that uses `output` to e.g. capture-{start, stop}.
+
     Parameters
     ----------
+    output
+        Output config for data product (BOutput or XOutput).
```
bmerry, line 165: I'd recommend against adding this - you're likely to paint yourself into a corner where it becomes messy to remove it again to support multiple beams.

Instead, I'd suggest making it plural from the start (outputs: Sequence[Output]). In XPipeline you can then have `output` be an alias to `outputs[0]` (e.g. as a `@property`).

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
         raise NotImplementedError  # pragma: nocover
 
 
+class BPipeline(Pipeline):
+    """Processing pipeline for a collection of :class:`~output.BOutput`."""
```
bmerry, line 277: Check if that actually links correctly - I suspect it won't.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            output,
+            engine.heaps_per_fengine_per_chunk,
+            self.n_tx_items,
+            n_channels_per_substream=engine.n_channels_per_substream,
+            spectra_per_heap=engine.src_layout.n_spectra_per_heap,
+            timestamp_step=engine.rx_heap_timestamp_step,
+            send_rate_factor=engine.send_rate_factor,
+            channel_offset=engine.channel_offset_value,
+            context=context,
+            packet_payload=engine.dst_packet_payload,
```
bmerry, line 321: Hmm, it occurs to me that while we need the payload size to be same for all beams, there is no fundamental reason it has to be the same between X and B. It's not something to worry about too much now, but maybe file a ticket.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            buffer_device = accel.DeviceArray(context, shape=buffer_shape, dtype=bsend.SEND_DTYPE)
+            # TODO: Clarify shape of `saturated`
+            saturated = accel.DeviceArray(
+                context, shape=(engine.heaps_per_fengine_per_chunk, self.n_beams), dtype=np.uint32
+            )
+            present_ants = np.zeros(shape=(engine.n_ants,), dtype=bool)
+            tx_item = TxQueueItem(buffer_device, saturated, present_ants)
+            self._tx_free_item_queue.put_nowait(tx_item)
+
+        # TODO: The way this is imported will change once the OperationSequence is in place
```
bmerry, line 310: Not sure I follow why?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+        self.send_stream = bsend.BSend(
+            output,
+            engine.heaps_per_fengine_per_chunk,
+            self.n_tx_items,
```
bmerry, line 314: Given the number of parameters, you might as well pass them all by keyword.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+                "Number of complex samples that saturated.",
+                default=0,
+                initial_status=aiokatcp.Sensor.Status.NOMINAL,
+            )
+        )
+
+    def capture_enable(self, enable: bool = True) -> None:  # noqa: D102
+        # NOTE: Defaulting to handle its single output for now.
+        # Actual implementation will require some kind of 'stream_id' arg,
+        # which currently contravenes the base function signature.
```
bmerry, line 354: I'd make this a TODO rather than just a NOTE.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+            BF_RAW_ID,
+            "bf_raw",
+            "Channelised complex data.",  # TODO: Straight from the ICD, is this sufficient?
+            shape=buffers[0].shape[2:],
+            dtype=buffers[0].dtype,
+        )
+
+        self.descriptor_heap = item_group.get_heap(descriptors="all", data="none")
+
+    def enable_substream(self, stream_id: int = 0, enable: bool = True) -> None:
```
bmerry, line 310: I'd remove the default for `substream_id`. Any code that calls it without passing the substream ID is going to need to be updated sooner or later, and it would be better if it was obviously passing a hard-coded stream ID.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            # - Wait for the HtoD transfers to complete, then
+            # - Give the chunk back to the receiver for reuse.
+            rx_item = await self._rx_item_queue.get()
+            if rx_item is None:
+                break
+            await rx_item.async_wait_for_events()
+
+            tx_item = await self._tx_free_item_queue.get()
+            await tx_item.async_wait_for_events()
+            # TODO: Is it fine to update the timestamp like this?
```
bmerry, line 369: I'm happy with it.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            # Get item from the receiver loop.
+            # - Wait for the HtoD transfers to complete, then
+            # - Give the chunk back to the receiver for reuse.
+            rx_item = await self._rx_item_queue.get()
+            if rx_item is None:
+                break
+            await rx_item.async_wait_for_events()
+
+            tx_item = await self._tx_free_item_queue.get()
+            await tx_item.async_wait_for_events()
```
bmerry, line 368: Can just do
```suggestion
            tx_item.enqueue_wait_for_events(self._proc_command_queue)
```
That ensures the GPU waits, without holding up the Python code.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            tx_item.reset(rx_item.timestamp)
+
+            # Bind input buffers
+
+            # Queue GPU work
+            # - For now, just fill it with zeros
+            tx_item.buffer_device.zero(self._proc_command_queue)
+            tx_item.saturated.zero(self._proc_command_queue)
+
+            # Bind output buffers
```
bmerry, line 379: That should be before the GPU work.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -259,9 +273,171 @@ def capture_enable(self, enable: bool = True) -> None:
...
+            item.buffer_device.get_async(self._download_command_queue, chunk.data)
+            item.saturated.get_async(self._download_command_queue, chunk.saturated)
+
+            # Wait for the transfer
+            event = self._download_command_queue.enqueue_marker()
+            await katsdpsigproc.resource.async_wait_for_events([event])
+
+            # Set the Chunk timestamp
+            chunk.timestamp = item.timestamp
+            if self.send_stream.tx_enabled:
```
bmerry, line 419: Add a TODO here so that it's clear this still needs doing.

Also I reckon the clip-cnt sensors should be updated regardless of whether the data is being transmitted.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -794,9 +967,14 @@ def __init__(
...
-            XPipeline(output, self, context, tx_enabled) for output in outputs if isinstance(output, XOutput)
-        ]
+
+        self._pipelines: list[Pipeline]
+        x_outputs = [output for output in outputs if isinstance(output, XOutput)]
+        b_outputs = [output for output in outputs if isinstance(output, BOutput)]
+        self._pipelines = [XPipeline(output, self, context, tx_enabled) for output in x_outputs]
+        if len(b_outputs) == 1:
+            # TODO: Update once more BOutput's are supported
+            self._pipelines.append(BPipeline(b_outputs[0], self, context, tx_enabled))
```
bmerry, line 977: Looks like if you specify multiple b_outputs, it will silently ignore all of them?

test/xbgpu/test_engine.py
```diff
@@ -614,7 +614,9 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
                 np.testing.assert_equal(expected_output, device_result)
             incomplete_accums_counters.append(incomplete_accums_counter)
 
-        xpipelines: list[XPipeline] = [pipeline for pipeline in xbengine._pipelines if isinstance(pipeline, XPipeline)]
+        xpipelines: list[XPipeline] = [
+            pipeline for pipeline in xbengine._pipelines if isinstance(pipeline, XPipeline)  # type: ignore
```
bmerry, line 618: Hmm, in theory mypy should have enough information that it doesn't need to ignore. What is the error?

src/katgpucbf/xbgpu/engine.py
```diff
@@ -966,9 +1149,11 @@ async def start(self, descriptor_interval_s: float = SPEAD_DESCRIPTOR_INTERVAL_S
         # Create the descriptor task first to ensure descriptor will be sent
         # before any data makes its way through the pipeline.
         for pipeline in self._pipelines:
+            # TODO: Make this parametrisable somehow for XPipeline vs BPipeline
+            # BPipeline needs to stagger the sending of descriptors
             descriptor_sender = DescriptorSender(
-                pipeline.send_stream.stream,
-                pipeline.send_stream.descriptor_heap,
+                pipeline.send_stream.stream,  # type: ignore
+                pipeline.send_stream.descriptor_heap,  # type: ignore
```
bmerry, line 1156: Seems like we should have a base Send class that contains the attributes/functionality that is common between XSend and BSend.

bmerry (changes requested):
I still need to review the tests, but this should give you something to chew on so long.

Since n_beams = 1 gives only a half-beam, I think it is worth aiming to support n_beams > 1 from the start, even if we don't support beam steering. The good news is I don't think there very much needed to support that - a few singletons replaced by arrays and a sprinkling of for loops.

I wouldn't worry about katsdpcontroller at all yet. Let's get to the point where it does something useful in isolation before worrying about integration with katsdpcontroller.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,390 @@
...
+            # before returning a :class:`.Chunk`.
+            self.future = asyncio.create_task(send_stream.stream.async_flush())
+            send_stream._chunks_queue.put_nowait(self)
+
+
+class BSend:
+    """
+    Class for turning tied array channelised voltage products into SPEAD heaps.
+
+    This is some text.
```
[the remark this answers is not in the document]
amishatishpatel: Had it in there to remind me to add a second line of description, don't know how I missed it during clean-up!

src/katgpucbf/xbgpu/engine.py
james-smith-za: I would not be against that, but let's maybe do that via a technical debt ticket once this PR is complete.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,190 @@
...
+        More specifically, send `N_TX_ITEMS` Chunks, each of which contain
+        `HEAPS_PER_FENG_PER_CHUNK` heaps.
+        """
+        # Send the descriptors as the recv_stream object needs it to
+        # interpret the received heaps correctly.
+        await send_stream.stream.async_send_heap(send_stream.descriptor_heap)
+
+        for i in range(N_TX_ITEMS):
+            # Get a free chunk - there is not always a free one available. This
+            # function yields until one it available.
```
james-smith-za, line 82:
```suggestion
            # function blocks until one is available.
```

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,190 @@
...
+            assert items["timestamp"].value == i * heap_timestamp_step
+            assert items["frequency"].id == FREQUENCY_ID
+            assert items["frequency"].value == channel_offset
+            assert items["bf_raw"].id == BF_RAW_ID
+            assert items["bf_raw"].value.shape == (n_channels_per_substream, n_spectra_per_heap, COMPLEX)
+            assert items["bf_raw"].value.dtype == np.int8
+            np.testing.assert_equal(items["bf_raw"].value, zero_data)
+
+    @pytest.mark.parametrize("num_channels", test_parameters.num_channels)
+    @pytest.mark.parametrize("num_spectra_per_heap", test_parameters.num_spectra_per_heap)
```
james-smith-za, line 133: Can probably use combinations of these, will make the quick test run quicker.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,190 @@
...
+            Device context for allocating buffers.
+        num_channels
+            Total number of channels processed by a (theoretical) F-engine.
+        num_spectra_per_heap
+            Total number of packed spectra in every recevied channel.
+        """
+        # TODO: We don't do channels * 2 anymore, but n-samples-between-spectra
+        heap_timestamp_step = num_channels * 2 * num_spectra_per_heap
+        # Arbitrarily chosen, channels-per-substream is dictated by the
+        # F-engine anyway.
```
james-smith-za, line 164: Well, it's dictated by the number of XB-engines that we need for a given n_antennas.
It's passed as a parameter to the F-engine.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,190 @@
...
+            items = ig.update(heap)
+            assert set(items.keys()) == {"timestamp", "frequency", "bf_raw"}
+            assert items["timestamp"].id == TIMESTAMP_ID
+            assert items["timestamp"].value == i * heap_timestamp_step
+            assert items["frequency"].id == FREQUENCY_ID
+            assert items["frequency"].value == channel_offset
+            assert items["bf_raw"].id == BF_RAW_ID
+            assert items["bf_raw"].value.shape == (n_channels_per_substream, n_spectra_per_heap, COMPLEX)
+            assert items["bf_raw"].value.dtype == np.int8
+            np.testing.assert_equal(items["bf_raw"].value, zero_data)
```
james-smith-za, line 130: For the purposes of this file as it currently stands, this is fine.

In general though, I'd prefer to have these assertions in the body of the test itself, rather than the helper function. That way you could re-use the helper function for other tests within the file.

No reason to change anything at this stage, but if the file expands to include more tests at some point it would be better to rearrange things.

james-smith-za (commented):
Overall I am happy with the structure of this work.
I've added a bit to a few of the comments that @bmerry has made, which I mostly concur with. A few other small comments.
Should be quick enough to tidy this up and get it merged.

src/katgpucbf/xbgpu/engine.py
bmerry: `enqueue_wait_for_events` only blocks the GPU stream. It returns immediately on the CPU, unlike `await async_wait_for_events` which prevents the next line of code from running until the work is complete.

james-smith-za: OK, I had misunderstood the distinction there.
The way Amish has it now is the way it's done in the XPipeline, so if there's an issue it may be worth addressing there as well. (In a separate ticket of course.)

bmerry: Yes I think I've noticed that code in XPipeline and it's bugged me. It's not wrong, and it probably makes no difference, but it's not quite as efficient as it could be.

james-smith-za: I'd suggest then leaving this one as-is for the time being, and filing a technical debt ticket to improve both this one and the Xpipeline one.

I don't like it when a PR goes through lots of rounds of review with many many changes requested, it makes keeping track of individual issues quite difficult.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: I've created [NGC-1144](https://skaafrica.atlassian.net/browse/NGC-1144). Thanks for flagging it.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Fair points all around (I genuinely thought I'd addressed this in a previous commit-review by Bruce).

I've created [NGC-1145](https://skaafrica.atlassian.net/browse/NGC-1145) in the meantime.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: [NGC-1146](https://skaafrica.atlassian.net/browse/NGC-1146) created accordingly.

amishatishpatel:
A couple things before I explicitly request a re-review:
1. My update to allow for multiple `--beam`s from the get-go has broken `xbgpu/test_engine.py` - fixing that now.
2. I need to discuss the update to `chunk.send` and surrounds with Bruce as it moves towards fgpu's send and away from XSend.{send_heap, get_free_heap} - will chat to him this morning.
3. A more top-level comment I wasn't explicit that I wanted feedback on, in the PR blurb - whenever either of you can, please and thank you.
> More admin-related, I'm not entirely trusting of sphinx-apidoc -o doc/ src/ as it didn't pick up/automatically
generate a .rst file for katgpucbf.xbgpu.bsend (when I ran it on gareth).
I didn't want to commit the file created by hand, but let me know if I should do that anyway.

src/katgpucbf/xbgpu/bsend.py
amishatishpatel: This comment might get lost in the various updates, and I don't think we've explicitly discussed this. I think I'll wrap it in a `try...except...finally` like you've done in `chunk_send_and_cleanup` - also like in my previous update to `send_chunk`.

amishatishpatel:
I'd forgotten to update `test_bsend.py` to process (at least) two `BOutput`s. Naively added it in and, of course, it did not work straight away. Debugging that at the moment. Feel free to hold off on this round of reviewing until I ping you, or get it out of your way in the meantime.

Either way, please and thank you for your patience.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -254,25 +266,180 @@ async def sender_loop(self) -> None:
...
+        while True:
+            # Get a populated TxQueueItem from the tx_item_queue
+            item = await self._tx_item_queue.get()
+            if item is None:
+                break
+            # The CPU doesn't need to wait, but the GPU does to ensure it
+            # won't start the download before computation is done.
+            item.enqueue_wait_for_events(self._download_command_queue)
+
+            # Get a free Chunk from the send_stream
```
bmerry, line 395: Seems like a low-value comment - it's mostly words that appear in the next line of code just written in a different order.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -254,25 +266,180 @@ async def sender_loop(self) -> None:
...
+            chunk = await self.send_stream.get_free_chunk()
+
+            # Kick off any data downloads from the GPU
```
bmerry, line 398:
```suggestion
            # Kick off data downloads from the GPU
```
"Any" sounds like there is doubt about whether there be some or not.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -254,25 +266,180 @@ async def sender_loop(self) -> None:
...
+        self.send_stream.enable_substream(stream_id=stream_id, enable=enable)
+
+
 class XPipeline(Pipeline):
     """Processing pipeline for a single baseline-correlation-products stream."""
 
     def __init__(
         self,
-        output: XOutput,
+        outputs: Sequence[XOutput],
```
bmerry, line 433: Since XPipeline can only work on a single output, I'd suggest reverting this change, and passing `[output]` to the base class constructor.

A lot of other changes in the constructor (where you changed `output` to `self.output`) can then also be reverted.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -326,6 +493,13 @@ def __init__(
 
         self._populate_sensors()
 
+    @property
+    def output(self) -> XOutput:
+        """Alias the parent-level :class:`Output` list."""
```
bmerry, line 498: I think I'd find that docstring confusing if I didn't already know what was going on. How about
```suggestion
        """The single :class:`Output` produced by this pipeline."""
```
(it might need a noqa because it's not in imperative voice, which I think is fine for property docstrings since they're pretending to be data and not code).

src/katgpucbf/xbgpu/engine.py
```diff
@@ -326,6 +493,13 @@ def __init__(
...
+        # TODO: Type ignore below is due to outputs being of type Output,
+        # whereas the return type indicates XOutput expected.
+        return self.outputs[0]  # type: ignore
```
bmerry, line 501: You can probably fix that by writing
```
    outputs: Sequence[XOutput]
```
at the top of the class

src/katgpucbf/xbgpu/engine.py
```diff
@@ -622,7 +796,8 @@ class XBEngine(DeviceServer):
         this XB-Engine. Used to set the value in the XB-Engine output heaps for
         spectrum reassembly by the downstream receiver.
     outputs
-        Output streams to generate. Currently only XOutputs supported.
+        Output streams to generate. Currently XOutputs and a single BOutput is
```
bmerry, line 799:
```suggestion
        Output streams to generate. Currently XOutputs and a single BOutput are
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -909,19 +1085,21 @@ async def _receiver_loop(self) -> None:
...
         ------
         FailReply
             If the `stream_name` is not a known output.
         """
         for pipeline in self._pipelines:
-            if stream_name == pipeline.output.name:
-                return pipeline
+            for output in pipeline.outputs:
+                if stream_name == output.name:
+                    return pipeline, pipeline.outputs.index(output)
```
bmerry, line 1102: This is searching the list of outputs twice. How about:
```suggestion
            for i, output in enumerate(pipeline.outputs):
                if stream_name == output.name:
                    return pipeline, i
```

bmerry (changes requested):
As suggested, I'm leaving review of the test until you've got it working. Overall things are looking in good shape - it's mostly nitpicks and TODOs. Note that some comments from previous reviews are still open.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,403 @@
...
+        Exception
+            For any other unexpected error.
+        """
+        chunk = await self._chunks_queue.get()
+        try:
+            await chunk.future
+        except asyncio.CancelledError:
+            pass
+        except Exception as ex:
+            raise Exception(ex.args)
```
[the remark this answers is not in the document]
amishatishpatel: Argh, apologies. I was on auto-pilot a bit when taking care of this case and documenting it. I see I can simply `logger.exception("Error sending chunk")`.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,403 @@
...
+        asyncio.CancelledError
+            If the chunk's future is cancelled at any point.
+        Exception
+            For any other unexpected error.
+        """
+        chunk = await self._chunks_queue.get()
+        try:
+            await chunk.future
+        except asyncio.CancelledError:
+            pass
```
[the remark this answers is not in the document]
amishatishpatel: Too right, too right. I'm going to update the docstring to remove mentions of what it `Raises` and say something like
* If the chunk's :attr:`future` is cancelled, we allow it to propagate up to the :class:`.XBEngine`-level.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: I suppose I should apply this to any other ~repetitive one-liner comments below. Still, appreciate sharing your stance on this.

bmerry: I am a bit of a documentation minimalist, and sometimes I take that too far, so don't feel you need to delete any comments that you think might be helpful.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Related to the self-explanatory comments, I'd left this kind of comment in here mostly as placeholders/reminders. I don't think they actually need to persist beyond this PR - I'm sure we'll be able to figure out what goes where/order of operations later on.

src/katgpucbf/xbgpu/engine.py
james-smith-za: If you're using xgpu as a reference, Gareth's habit was to be very verbose in comments. Helpful if you're wading in a code-base that you're not familiar with at all, but can be taken too far.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -794,9 +938,13 @@ def __init__(
...
         self._active_in_sem = asyncio.BoundedSemaphore(1)
-        self._pipelines = [
-            XPipeline(output, self, context, tx_enabled) for output in outputs if isinstance(output, XOutput)
-        ]
+
+        self._pipelines: list[Pipeline] = []
+        x_outputs = [output for output in outputs if isinstance(output, XOutput)]
+        b_outputs = [output for output in outputs if isinstance(output, BOutput)]
+        self._pipelines = [XPipeline(x_output, self, context, tx_enabled) for x_output in x_outputs]
+        if len(b_outputs):
```
james-smith-za, line 946: I think you can just do 
```suggestion
        if b_outputs:
```

test/xbgpu/test_bsend.py
james-smith-za: I'd still suggest rewording this comment to something like
```suggestion
        # Arbitrarily chosen, channels-per-substream is dictated by the
        # top-level correlator configuration anyway.
```

james-smith-za (changes requested):
@amishatishpatel I have made a few nit-picks but they're not really important, I wouldn't insist on them.

The one change that I will insist on is the one that has to do with `get_free_chunk` where it looks like the functionality does not correspond to what the docstring says, and the docstring itself is incomplete.

Once that is addressed, and @bmerry is happy with the unit tests, then I'd say we can go ahead and merge.

src/katgpucbf/xbgpu/bsend.py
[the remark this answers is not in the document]
amishatishpatel: I could've sworn I updated the docstring right after I removed the `Raises` discussion. Very strange.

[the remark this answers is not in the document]
james-smith-za: Yes, that was the other thing I wanted addressed.
Does the function actually do what you want it to do.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,399 @@
...
+                    np.uint32,
+                    context=context,
+                ),
+                channel_offset=channel_offset,
+                timestamp_step=timestamp_step,
+            )
+            self._chunks_queue.put_nowait(chunk)
+            buffers.append(chunk.data)
+
+        # TODO: Follow suit with other Engines to calculate `rate`
```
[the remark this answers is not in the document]
amishatishpatel: Ah, fair point. I am referring to the respective calculations in [xbgpu/xsend](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/xbgpu/xsend.py#L257-L269) and [fgpu/send](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/send.py#L257-L259). I've updated to read 
* [...] to calculate `rate` below

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -327,13 +326,16 @@ def enable_substream(self, stream_id: int, enable: bool = True) -> None:
...
+        ------
+        asyncio.CancelledError
+            If the chunk's send future is cancelled.
         """
         chunk = await self._chunks_queue.get()
         try:
             await chunk.future
         except asyncio.CancelledError:
-            pass
+            raise asyncio.CancelledError("Sending chunk was cancelled")
```
james-smith-za, line 338: You can just do `raise` and it will send the exception on upwards unchanged.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -327,13 +326,16 @@ def enable_substream(self, stream_id: int, enable: bool = True) -> None:
...
         except Exception:
             logger.exception("Error sending chunk")
```
james-smith-za, line 340: This on the other hand suppresses any other exceptions.
Are you sure this is what you want to do?

src/katgpucbf/xbgpu/bsend.py
amishatishpatel: Too right, I sometimes forget "how useful" the tracebacks are when an Exception is raised. Happy to remove the ~custom message if you'd prefer it was more straightforward.

src/katgpucbf/xbgpu/bsend.py
amishatishpatel: Ah, well. What I actually (and usually) want to do is use the [traceback](https://docs.python.org/3/library/traceback.html) module, but that might be some outdated idea/expectation I have about debugging.

OTOH, what can I/we actually do here? 
* I'm genuinely not sure if a hard failure would cause this to execute, and whether we'd want to stop the BPipeline altogether if this did happen. Genuinely, open to suggestions.
* Perhaps I could at least `chunk.future.set_result(None)` and send it back on its way?

james-smith-za:
I'm actually not quite sure at this point. @bmerry do you have any opinions?

Disclaimer

The information contained in this communication from the sender is confidential. It is intended solely for use by the recipient and others authorized to receive it. If you are not the recipient, you are hereby notified that any disclosure, copying, distribution or taking action in relation of the contents of this information is strictly prohibited and may be unlawful.

src/katgpucbf/xbgpu/bsend.py
james-smith-za: Yeah I'd just have used a plain `raise`.
If the behaviour that you are after is raising this particular exception and catching and logging all other kinds.

bmerry:
> Once that is addressed, and @bmerry is happy with the unit tests, then I'd say we can go ahead and merge.

Before merging I think we should go over the TODOs and decide which need to be addressed first; I've been assuming that they're intended to be fixed as part of the review. Of the top of my head, I will insist that we properly calculate the output data rate, so that we don't spam the network and cause interference with other (non-beamformer) engines that happen to be running on the same nodes.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+from . import test_parameters
+
+HEAPS_PER_FENG_PER_CHUNK: Final[int] = 5
+N_TX_ITEMS: Final[int] = 2
+TOTAL_DATA_HEAPS_PER_SUBSTREAM: Final[int] = N_TX_ITEMS * HEAPS_PER_FENG_PER_CHUNK
+
+
+@pytest.fixture
+def time_converter() -> TimeConverter:
+    # TODO: Probably best to use adc_sample_rate from meerkat.py::BANDS
```
bmerry, line 45: On the other hand, you could pick an obviously artificial value, like 1234e6. This code is not supposed to be MK-specific after all.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+    return TimeConverter(123456789.0, 1712e6)
+
+
+@pytest.fixture
+def outputs() -> Sequence[BOutput]:
+    """Simulating `--beam` configuration."""
```
bmerry, line 51:
```suggestion
    """Simulate `--beam` configuration."""
```

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+
+from katgpucbf import COMPLEX
+from katgpucbf.spead import BF_RAW_ID, FREQUENCY_ID, TIMESTAMP_ID
+from katgpucbf.utils import TimeConverter
+from katgpucbf.xbgpu.bsend import BSend
+from katgpucbf.xbgpu.output import BOutput
+
+from . import test_parameters
+
+HEAPS_PER_FENG_PER_CHUNK: Final[int] = 5
```
bmerry, line 38: I don't think this is really the right name for this context. BSend has nothing to do with F-engines.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @staticmethod
+    async def _recv_data(
+        queues: list[spead2.InprocQueue],
+        channel_offset: int,
+        n_channels_per_substream: int,
+        n_spectra_per_heap: int,
+        heap_timestamp_step: int,
+    ) -> None:
+        """Receive data transmitted from :func:`_send_data`.
```
bmerry, line 116:
```suggestion
        """Receive data transmitted from :meth:`_send_data`.
```

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+        outputs
+            Fixture
+        queues
+            List of :class:`spead2.InprocQueue` used to transmit heaps
+            in :meth:`_send_data`.
+        channel_offset, n_channels_per_substream, n_spectra_per_heap, heap_timestamp_step
+            Variables declared by the calling unit test to verify
+            transmitted data.
+        """
+        out_config = spead2.recv.StreamConfig(max_heaps=100)
```
bmerry, line 131: Why override max_heaps? Since you're using a separate recv Stream per substream, I would expect the default to work fine.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,228 @@
...
+        in-process transport. The received heaps are then checked.
+
+        This test does not generate random data as it will take much more compute
+        to check that the random data is received correctly.
+
+        .. todo::
+
+            Update this test to make use of mock_{send, recv}_stream fixtures.
+            Perhaps just mock_send_stream, as mock_recv_stream might require
+            more refactoring.
```
bmerry, line 187: Why? At this level, a custom stream_factory is the appropriate way to make use of the inproc transport.

bmerry (changes requested):
Just reviewing the unit tests here. The code that has been written is fine, but I would like to see the payload be random data rather than zeros, together with verification that the right pieces of payload are being received.

I'd also like to see verification of the heap_cnt. I suspect that will fail since there isn't a call to set_cnt_sequence anywhere in the PR.

src/katgpucbf/xbgpu/bsend.py
bmerry: I agree
```suggestion
            raise
```

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,401 @@
...
+        asyncio.CancelledError
+            If the chunk's send future is cancelled.
+        """
+        chunk = await self._chunks_queue.get()
+        try:
+            await chunk.future
+        except asyncio.CancelledError:
+            raise asyncio.CancelledError("Sending chunk was cancelled")
+        except Exception:
+            logger.exception("Error sending chunk")
```
bmerry, line 340: I'm happy with this logging and otherwise suppressing the error. The point of waiting for the future is just to be sure that we're not going to overwrite data that is still being transmitted. If sending failed, then it's no longer being transmitted and it's safe to return the chunk.

test/xbgpu/test_bsend.py
amishatishpatel: Too right, I think I was too focused on the whole 'use the fixtures we have'. Also somehow concluded test_xsend.py was out of date since it hasn't been touched for a while. 

Will remove this `todo`.

test/xbgpu/test_bsend.py
amishatishpatel: Ah, I think I updated it way back when I was debugging 'Queue full' errors, and just put it to something >>larger than the amount of data being sent.

Will remove this override now.

test/xbgpu/test_bsend.py
amishatishpatel: 50/50 chance every time I do this, and I seem to get it wrong each time 😅

james-smith-za: My daughter also puts her shoes on the wrong feet more often than not.
Maybe it's a common human thing? 🤔

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,401 @@
...
+            Also update its relevant counters and sensor values.
+        """
+        if any(send_stream.tx_enabled):
+            for frame in self._frames:
+                heaps_to_send = [heap for heap, enabled in zip(frame.heaps, send_stream.tx_enabled) if enabled]
+                send_future = send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
+            # TODO: Update counters and sensor with chunk.saturation
+            # NOTE: We use the last future returned above as the data is sent
+            # in-order. Once the last frame has been transmitted, the rest have
+            # too.
```
bmerry, line 181: I've just realised there is a flaw in this idea - sorry for leading you on a wild goose chase (again!) It's fine if there are no errors, but there are a couple of problems if any of the sends fail:
1. If the *last* send fails because there isn't enough space in the queue (which should be impossible, but as you know bugs can happen), it will fail immediately, even though previous heaps are still in flight. That could allow data to be modified before it's fully transmitted.
2. If an earlier send fails, nothing catastrophic happens but the error is silently lost.

So unfortunately I'm going to need to ask you change this back to using gather :-(.

amishatishpatel: Ah, too right! Honestly, I was surprised by [your confidence in](https://github.com/ska-sa/katgpucbf/pull/650#discussion_r1372631225)
> [The data is sent in-order, so] once the last frame has been transmitted, the rest have too.

Genuinely, not a problem and appreciate revisiting this point.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        if any(send_stream.tx_enabled):
+            send_futures = []
+            for frame in self._frames:
+                heaps_to_send = [heap for heap, enabled in zip(frame.heaps, send_stream.tx_enabled) if enabled]
+                send_futures.append(
+                    send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
+                )
+            # TODO: Update counters and sensor with chunk.saturation
+            self.future = asyncio.gather(*send_futures)
```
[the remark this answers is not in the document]
amishatishpatel: Let me know if I should update the docstring/add a note to discuss this point (i.e. why we gather *all* async_send_heap Futures).

src/katgpucbf/xbgpu/bsend.py
amishatishpatel: I added a small update to the docstring to mention this.

test/xbgpu/test_bsend.py
amishatishpatel: In case this gets lost in the updates, I updated to `HEAPS_PER_CHUNK`.

test/xbgpu/test_bsend.py
```diff
@@ -197,32 +212,58 @@ async def test_send_simple(
...
         """
+        # We need the "total number of engines required" to process this array
+        # configuration.
+        n_engines = 1
+        while n_engines < num_ants * 4:
+            n_engines *= 2
+
+        # The test still needs to have some idea of "which engine" this is in a
+        # sequence of B-engines.
+        engine_id = 4  # Arbitrarily chosen
```
bmerry, line 223: I don't think it's necessary to repeat this test with different numbers of antennas, since it's just a very indirect way to control the number of engines. My preference would be just to test with a fixed number of engines (say 8), but you could also put n_engines into the pytest.mark.combinations with a few values.

Setting engine_id to 4 when n_engines might be <= 4 is also invalid.

test/xbgpu/test_bsend.py
```diff
@@ -89,82 +94,101 @@ async def _send_data(
...
+        for i, queue in enumerate(queues):
             stream = spead2.recv.asyncio.Stream(out_tp, out_config)
             stream.add_inproc_reader(queue)
 
             # Wait for the first packet to arrive - it is expected to be the
             # SPEAD descriptor.
             ig = spead2.ItemGroup()
             heap = await stream.get()
-            # TODO: No checks on the count sequence yet
+            assert heap.cnt % n_engines == engine_id % n_engines, "The heap IDs are not correctly strided"
```
bmerry, line 164:
```suggestion
            assert heap.cnt % n_engines == engine_id, "The heap IDs are not correctly strided"
```
Since engine_id should be in the range [0, n_engines).

bmerry (changes requested):
Looks like mostly nitpicks at this point. Once they're addressed I'll look over the whole thing again just to check that everything fits together.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -299,8 +299,10 @@ def __init__(
             outputs=outputs,
             heaps_per_fengine_per_chunk=engine.heaps_per_fengine_per_chunk,
             n_tx_items=self.n_tx_items,
+            n_channels=engine.n_channels_total,
             n_channels_per_substream=engine.n_channels_per_substream,
             spectra_per_heap=engine.src_layout.n_spectra_per_heap,
+            adc_sample_rate=engine.adc_sample_rate_hz,
```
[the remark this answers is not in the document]
amishatishpatel: Too right, I've created [NGC-1158](https://skaafrica.atlassian.net/browse/NGC-1158).

test/xbgpu/test_bsend.py
```diff
@@ -89,82 +94,101 @@ async def _send_data(
...
-            Fixture
+        data
+            Random data generated during data transmission, of shape
+            - (N_TX_ITEMS, HEAPS_PER_CHUNK, len(outputs), n_channels_per_substream, n_spectra_per_heap, COMPLEX)
         queues
             List of :class:`spead2.InprocQueue` used to transmit heaps
             in :meth:`_send_data`.
-        channel_offset, n_channels_per_substream, n_spectra_per_heap, heap_timestamp_step
-            Variables declared by the calling unit test to verify
-            transmitted data.
```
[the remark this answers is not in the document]
amishatishpatel: I looked at `_send_data` above and how I hadn't documented any of it, and drew some arbitrary line about what does and doesn't require discussion.

Silly me. I'll update to include all of the above, at least here in `_recv_data`.

test/xbgpu/test_bsend.py
amishatishpatel: Hmmmm, I naively updated the line to read `engine_id = 4 if n_engines > 4 else 3`, but that seems ~unnecessary.

Easier to set it to 3 which satisfies all variant of `num_ants`, and update the comment to explain it's not completely arbitrarily chosen.

test/xbgpu/test_bsend.py
```diff
@@ -0,0 +1,278 @@
...
+                assert items["frequency"].id == FREQUENCY_ID
+                assert items["frequency"].value == channel_offset
+                assert items["bf_raw"].id == BF_RAW_ID
+                assert items["bf_raw"].value.shape == (n_channels_per_substream, n_spectra_per_heap, COMPLEX)
+                assert items["bf_raw"].value.dtype == np.int8
+                np.testing.assert_equal(items["bf_raw"].value, data[j, i, ...])
+
+    @pytest.mark.combinations(
+        "num_ants, num_channels, num_spectra_per_heap",
+        [1, 19, 80],
```
bmerry, line 192: It still seems very roundabout to parametrise the test by num_ants and then use a completely arbitrary formula to determine an n_engines value from that. Why not just parametrise n_engines directly e.g.
```suggestion
        "n_engines, num_channels, num_spectra_per_heap",
        [4, 128, 512],
```
Apart from anything else, it would make the test identifier actually contain the number of engines being tested.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,429 @@
...
+
+    To allow this class to be used with multiple transports, the constructor
+    takes a factory function to create the stream.
+
+    Parameters
+    ----------
+    outputs
+        Sequence of :class:`.output.BOutput`.
+    heaps_per_fengine_per_chunk
+        Number of SPEAD heaps from one F-engine in a single received Chunk.
```
bmerry, line 215: Let's rename this: this class is about transmission and the interface shouldn't be tied to our working assumption that we're have output chunks the same time-length as the input.

Maybe "frames_per_chunk"?

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,429 @@
...
+    n_tx_items
+        Number of :class:`Chunk` to create.
+    adc_sample_rate, n_channels, n_channels_per_substream, spectra_per_heap, channel_offset
+        See :class:`.XBEngine` for further information.
+    timestamp_step
+        The timestamp step between successive heaps, as dictated by the XBEngine.
```
bmerry, line 221:
```suggestion
        The timestamp step between successive heaps.
```
It's dictated by the user of this class, which could be XBEngine, or it could be test_bsend.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,429 @@
...
+    send_rate_factor
+        Factor dictating how fast the send-stream should transmit data.
+    context
+        Device context to create buffers.
+    stream_factory
+        Callback function to create the spead2 send stream. It is passed the
+        stream configuration and memory buffers.
+    packet_payload
+        Size, in bytes, for the output packets (tied array channelised voltage
+        payload only, headers and padding are added to this).
```
bmerry, line 231:
```suggestion
        payload only; headers and padding are added to this).
```

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,429 @@
...
+                    dtype=data.dtype,
+                    value=self.data[i, ...],
+                )
+            )
+            self.heaps.append(spead2.send.HeapReference(heap, substream_index=i))
+
+
+class Chunk:
+    """
+    An array of :class:`Frame`.
```
bmerry, line 94:
```suggestion
    An array of :class:`Frame`\ s.
```

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -0,0 +1,429 @@
...
+        Storage for tied-array-channelised-voltage data, with shape (n_frames,
+        n_beams, n_channels_per_substream, n_spectra_per_heap, COMPLEX) and
+        dtype :const:`SEND_DTYPE`.
+    saturated
+        Storage for saturation counts, with shape (n_frames, n_beams) and dtype
+        uint32.
+    channel_offset
+        The first frequency channel processed.
+    timestamp_step
+        Timestamp step between successive :class:`Frame` in a chunk.
```
bmerry, line 108:
```suggestion
        Timestamp step between successive :class:`Frame`\ s in a chunk.
```

bmerry (changes requested):
We're into bike-shedding territory now. I've made a few comments for suggested changes, but I wouldn't be offended if you decided to just merge.

test/xbgpu/test_bsend.py
amishatishpatel: Ah, I see what you were referring to in the previous comment. Shall do, shall do.

PR #660, opened 2023-11-01
Adjust thread count for benchmark_fgpu

bmerry:
Using 4 threads for n=1 and 2 threads for n=2 seems to improve performance.

- [x] (n/a) If modules are added/removed: use sphinx-apidoc to update files in `doc/`

Relates to NGC-1120.

### DIFF
--- a/scratch/benchmarks/benchmark_fgpu.py
+++ b/scratch/benchmarks/benchmark_fgpu.py
@@ -129,12 +129,12 @@ def fgpu_factory(
         dst_chunk_jones = src_chunk_samples // 4
     if n == 1:
         interface = ",".join(server.interfaces[:2])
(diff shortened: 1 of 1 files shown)


amishatishpatel (approved):

bmerry:
@amishatishpatel raised the point that previously I switched this in the opposite direction - why is it different now? The previous commits were 40e62c6 and be72db7. The former indicates that at the time, 4 threads were causing unreliability (chance of dropping packets was neither 0% nor 100% across a range of rates). There doesn't seem to be evidence for this now.

bmerry:
Actually I think the trade-off is that more threads are needed if the digitiser packets get very small. If one assumes that digitiser packets are always as big as possible (for a power-of-two sample count) given the MTU, then the lower thread count seems to be better.

PR #661, opened 2023-11-01
Bring run-fgpu-hbw.sh up to date

bmerry:
- Set number of destination multicast groups to match run-fgpu.sh
- Set --src-buffer and --src-chunk-samples to match benchmark_fgpu.py

### DIFF
--- a/scratch/fgpu/run-fgpu-hbw.sh
+++ b/scratch/fgpu/run-fgpu-hbw.sh
@@ -11,8 +11,8 @@ dst_affinity="8"
 dst_comp=$dst_affinity
 other_affinity="12"
 src="239.102.0.64+15:7148"
-dst="239.102.200.0+15:7148"
-nb_dst="239.102.216.0+15:7148"
+dst="239.102.200.0+63:7148"
+nb_dst="239.102.216.0+7:7148"
 katcp_port="7140"
 prom_port="7150"
 feng_id="0"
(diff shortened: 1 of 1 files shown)


scratch/fgpu/run-fgpu-hbw.sh
```diff
@@ -21,8 +21,9 @@ export CUDA_VISIBLE_DEVICES="$cuda1"
 
 set -x
 exec spead2_net_raw taskset -c $other_affinity fgpu \
-    --src-chunk-samples 67108864 \
+    --src-chunk-samples 134217728 \
     --dst-chunk-jones 33554432 \
+    --src-buffer=268435456 \
```
amishatishpatel, line 26: It's a bit late for me to be asking what the significance of these numbers are (are they special in some way? Some multiple/power?).

I'll approve so long. Feel free to reply here or Mattermost or chat during the check-in.

amishatishpatel (approved):

scratch/fgpu/run-fgpu-hbw.sh
bmerry: They're powers of 2.

amishatishpatel: My brain wasn't up to temperature yet. I naively did `math.sqrt` on the new values, instead of `math.log2`. Of course they're powers of 2 🤦

PR #668, opened 2023-11-21
Add pol= parameter to --beam command-line option

bmerry:
This determines which polarisation of each antennas goes into the calculation for the beam.

Relates to NGC-1115.

### DIFF
--- a/src/katgpucbf/xbgpu/main.py
+++ b/src/katgpucbf/xbgpu/main.py
@@ -74,7 +74,7 @@ class BOutputDict(OutputDict, total=False):
     See :class:`OutputDict` for further information.
     """
 
-    pass
+    pol: int
 
 
 class XOutputDict(OutputDict, total=False):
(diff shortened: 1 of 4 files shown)


src/katgpucbf/xbgpu/main.py
```diff
@@ -123,14 +123,23 @@ def parse_beam(value: str) -> BOutput:
...
+    - pol
     """
 
     def _field_callback(kws: BOutputDict, key: str, data: str) -> None:
-        raise ValueError(f"unknown key {key}")
+        match key:
+            case "pol":
+                kws[key] = int(data)
+                if kws[key] not in {0, 1}:
+                    raise ValueError("pol must be either 0 or 1")
```
amishatishpatel, line 134: Worth asking, would it be equally acceptable to do `if kws[key] not in {0, 1}` *before* making the assignment?
* i.e. Why check after we've stored it?

amishatishpatel (approved):
One minor comment, otherwise good to go.

src/katgpucbf/xbgpu/main.py
bmerry: Sounds like you're suggesting
```python
                if kws[key] not in {0, 1}:
                    raise ValueError("pol must be either 0 or 1")
                kws[key] = int(data)
```
That won't work because `kws[key]` won't exist yet.

I guess one could do
```python
                value = int(data)
                if value not in {0, 1}:
                    raise ValueError("pol must be either 0 or 1")
                kws[key] = value
```
but that's more code.

amishatishpatel: Argh, too right. Or even
```python
if int(data) not in {0, 1}:
    raise ValueError("pol must be either 0 or 1")
kws[key] = int(data)
```

But I gather there is some unease around using a naked `int(data)`.

PR #671, opened 2023-11-23
Optimise and document beamformer kernel

bmerry:
Speed up the beamformer kernel by pre-computing steering weights once and re-using many times. Also write a high-level design doc and plenty of comments in the code.

Closes NGC-1171.


doc/xbgpu.design.b.rst
```diff
@@ -0,0 +1,57 @@
+Beamforming
+===========
+
+GPU kernel
+----------
+For the low number of beams required for MeerKAT (8 single-pol beams), the
+beamforming operation is entirely bandwidth-limited. There is thus no benefit
```
amishatishpatel, line 7: Minor gripe/request to be more explicit about what bandwidth we/you're referring to here. 
* I know context etc should explain it, but 
* I did Ctrl+F through the docs for *bandwidth* and 
* almost all usages have some explicit context mentioned near/next to the word usage (e.g. 'DRAM bandwidth', or 'bandwidth of the network interface').

src/katgpucbf/xbgpu/kernels/beamform.mako
```diff
@@ -50,65 +52,96 @@ DEVICE_FN float2 cmul(float2 a, float2 b)
 // Compute a * b + c in complex numbers
 DEVICE_FN float2 cmad(float2 a, float2 b, float2 c)
 {
-    return make_float2(a.x * b.x - a.y * b.y + c.x, a.x * b.y + a.y * b.x + c.y);
+    return make_float2(c.x + a.x * b.x - a.y * b.y, c.y + a.x * b.y + a.y * b.x);
```
amishatishpatel, line 55: I see this change belongs to an optimisation commit with the blurb
> Reorder operations in cmad to use **fused operations throughout** [emphasis added]

My (admittedly few minutes of) Google'ing didn't return anything overly useful, at least on first read. Basically,
* What is a fused operation? And/or
* Why is this update better than it was before?

amishatishpatel (commented):
One comment for consideration, the other a request for explanation/clarification.

Must admit, I was alarmed by lack of `beamform-wiring` in `xbgpu/engine.py` in this review. Realised you were optimising in parallel to the actual `processing_loop` update. One day I'll be able to play such 4D chess.

src/katgpucbf/xbgpu/kernels/beamform.mako
```diff
@@ -50,65 +52,96 @@ DEVICE_FN float2 cmul(float2 a, float2 b)
...
+    int n_spectra
 )
 {
-    int time = get_global_id(0);
+    /* Local storage for computed weights. The logical shape is
+     * [BATCH_ANTENNAS][N_BEAMS] but some parts of the code index
+     * it as a linear array.
+     */
+    LOCAL_DECL float2 l_weights[BATCH_ANTENNAS * N_BEAMS];
+    const int beam_pols[N_BEAMS] = { ${ ", ".join(str(p) for p in beam_pols) } };
```
amishatishpatel, line 79: Also, I should really learn `mako`.

src/katgpucbf/xbgpu/kernels/beamform.mako
bmerry: Fused multiply-add (FMA): a single instruction that computes x*y+z (with correct rounding, rather than rounding x*y first). NVIDIA GPUs execute that twice as fast as using separate multiple and add instructions. The compiler was previously generating a MUL, and FMA and an ADD (I forget in what order). With the new version it is two FMA instructions.

amishatishpatel: Two points
First, Re:
> The compiler was previously generating a MUL, an FMA and an ADD (I forget in what order)

Not sure how else to word "How did you know that?".  I guess, now that you've pointed it out, I perhaps see 
* `a.x * b.x` as the single MUL instruction,
* `a.x * b.y + c.x` as an FMA,
* and the leftover being the ADD.

Whereas now, with your update
* The first FMA == `c.x + a.x * b.x` , and
* ... I'm actually not sure how the second FMA comes about? Is there still a final ADD?

Apologies if this is ~incoherent, happy to Zoom for a few mins.

bmerry: > How did you know that?

Nsight showed me the instruction mix (which had lots of FMUL and FADD) and I think maybe even warned me about the low density of FFMA.

> ... I'm actually not sure how the second FMA comes about? Is there still a final ADD?

Let t = c.x + a.x * b.x (the first FMA). Then the second FMA is t - a.y * b.y (it's a subtract rather than an add, but IIRC NVIDIA lets you negate operands for free).

amishatishpatel: Ah, fantastic. I was guesstimating the result of `a.y * b.y` gets folded into another FMA, but it's actually/likely the former that gets computed first (maybe due to left-to-right execution?).

Either way, explanation is much appreciated.

amishatishpatel (approved):

PR #677, opened 2024-01-22
Qualification report fixes

bmerry:
Several modifications to the qualification report:

- Generate figures as PDFs at runtime, instead of as TeX code (faster test execution, much faster report generation, much smaller JSON files, slightly larger final report PDFs, fonts might not perfectly match between figures and text).
- When processing existing JSON files, write figure TeX code to temporary files which are included with `\input` (possibly fixes NGC-1182, and generally should be more robust to catcode changes in the figure TeX)
- Fix reporting of test_check failures (after a previous upgrade of pytest-check)
- Jenkins: allow the image to test to be overridden by an environment variable (in turn set by a parametrized build)

- [x] If qualification tests are changed: attach a sample qualification report: [report-new-opt.pdf](https://github.com/ska-sa/katgpucbf/files/14011452/report-new-opt.pdf) (note, has been run through an optimiser to reduce the file size to fit within Github's limit)

Closes NGC-1196.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,10 +106,11 @@ RUN pip install --no-deps "$(grep '^pycuda==' /tmp/katgpucbf/requirements.txt)" 
 FROM build-base as jenkins
 
 # docker so that Jenkins can build a Docker image
-# All the TeX stuff for building the documentation and qualification report
+# All the TeX and font stuff for building the docs and qualification report
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
     docker.io \
     docker-buildx \
+    fonts-liberation2 \
     latexmk \
     lmodern \
     pdf2svg \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1
 
 ################################################################################
-# Copyright (c) 2021-2023, National Research Foundation (SARAO)
+# Copyright (c) 2021-2024, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1
 
 ################################################################################
-# Copyright (c) 2021-2023, National Research Foundation (SARAO)
+# Copyright (c) 2021-2024, National Research Foundation (SARAO)
```
james-smith-za, line 4: Aha. Going to need to start remembering that...

qualification/Jenkinsfile
```diff
@@ -86,7 +86,7 @@ pipeline {
          sh '''
            spead2_net_raw pytest -v -c qualification/pytest-jenkins.ini qualification\
            --suppress-tests-failed-exit-code \
-           --image-override katgpucbf:harbor.sdp.kat.ac.za/cbf/katgpucbf:latest \
+           --image-override katgpucbf:${katgpucbf_image:-harbor.sdp.kat.ac.za/cbf/katgpucbf:latest} \
```
james-smith-za, line 89: This ... looks very weird. I can't find a reference document anywhere that lets me understand what's going on here. The colon and hyphen - is that Jenkinsfile syntax or something else?

bmerry: It's bash syntax. The bit after the `:-` is used if the variable is not set. Once everything is merged we can probably remove the default, since it's really just there to avoid breaking anything during the transition (since the Jenkins job also has a default).

james-smith-za: Ah, that makes sense. I hadn't known that about bash.

james-smith-za (approved):

PR #678, opened 2024-01-29
Bump pillow to 10.2.0

bmerry:
To keep dependabot happy (the security issue is extremely unlikely to affect us).

Checklist is all N/A.


amishatishpatel (approved):
When you have a second, could you point me to some info on why
> the security issue is extremely unlikely to affect us

Please and thank you.

bmerry:
> the security issue is extremely unlikely to affect us

An attacker has to be able to control the filenames of images being processed with the affected function. https://duartecsantos.github.io/2023-01-02-CVE-2023-50447/

PR #679, opened 2024-01-29
Change sample_tone_response to return float64

bmerry:
It previously returned int32, which is probably left over from a previous implementation (where self-correlations were used and so it could directly return values computed by the X engine). The result is only used to two places:

- sample_tone_response_hdr, which immediately explicitly cast it to float64.
- test_linearity, which normalised it by dividing by the largest element, hence implicitly casting to float.

It has more recently started causing warnings (NGC-1182) when the result of np.hypot was greater than 2^31, which is expected to happen during sample_tone_response_hdr as the power is cranked up.

- [x] If qualification tests are changed: attach a sample qualification report: see http://10.8.80.22:8080/job/qualification_katgpucbf/540/

Closes NGC-1182.


amishatishpatel (approved):
Nice one!

PR #688, opened 2024-02-14
Initial work on qualification testing for the beamformer

bmerry:
This is mostly about putting the infrastructure in place, and enabling 4 beams for all the existing tests just to ensure that it keeps up. There is one beamformer-specific test, but it doesn't yet do what it ought to be doing because the functionality isn't in place yet (it's nevertheless useful to check that data is being received correctly).

There are some checklist items that still need to be attended to before this can be merged (in particular, ensure that it doesn't break the existing tests), but the code can be reviewed so long.

- [x] If qualification tests are changed: attach a sample qualification report: http://10.8.80.22:8080/job/qualification_katgpucbf/564/Qualification_20Test_20Report
- [ ] Merge ska-sa/katsdpcontroller#714

Closes NGC-1118.


bmerry:
It seems like there are still some qualification tests failing. So definitely don't merge until that's fixed.

bmerry:
> It seems like there are still some qualification tests failing. So definitely don't merge until that's fixed.

4 beams comes to 110 Gb/s, so it's fully expected that it will lose packets on the receive side as it's only a 100 Gb/s interface. For now I'll just reduce it to 3 beams, and also make it configurable. We can use 4 beams for qualification testing on site.

bmerry:
I've updated the tests to consume as many beams as the network interface will allow for; and to skip tests if there are less than 2 full beams. That will allow tests to assume they can configure the first two beams and run comparisons between them.

For on-site qualification we'll have a 200 Gb interface, which should always have the capacity to absorb 4 beams.

qualification/conftest.py
```diff
@@ -281,9 +285,20 @@ def matplotlib_report_style() -> Generator[None, None, None]:
         yield
 
 
+@pytest.fixture(autouse=True)
+def quiet_spead2(caplog: pytest.LogCaptureFixture) -> None:
+    """Ensure spead2 only logs warnings and bug.
```
james-smith-za, line 290: Warnings and bug?
Maybe "warnings and above"?

qualification/report/generate_pdf.py
```diff
@@ -75,14 +75,15 @@ def make_cbf_mode_str(config: dict, *, expand: bool = False) -> str:
...
+        "beams": "4",
     }
 
     If `expand` is True, it will return a 'long description' as a sentence:
-    - 4 antennas, 8192 channels, L-band, 0.5s integrations, 4 dsims
+    - 4 antennas, 8192 channels, 4 beams, L-band, 0.5s integrations, 4 dsims
 
     If `expand` is False, it will return a MeerKAT config mode string of:
-    - 8n856M8k
+    - c8n856M8k
```
james-smith-za, line 86:
```suggestion
    - bc8n856M8k
```
?

james-smith-za (approved):
Overall happy, just a couple of possible typos that I have picked up?

james-smith-za:
I see that this gain test comes all in a big chunk at the end of the report.

Do we have some sort of logic to arrange them? Is this expected?

bmerry:
> I see that this gain test comes all in a big chunk at the end of the report.
> 
> Do we have some sort of logic to arrange them? Is this expected?

My guess is that the tests will be sorted by directory (antenna_channelised_voltage, baseline_correlation_products and tied_array_channelised_voltage). There is already an open ticket to add more levels of hierarchy to the sectioning of the report.

qualification/conftest.py
bmerry: Yes, that's exactly what was in my brain but my traitor fingers typed it wrong. Will fix.

PR #691, opened 2024-02-20
Add requirements text for some more requirements.

bmerry:


qualification/report/requirements_text/CBF-REQ-0095.tex
```diff
@@ -0,0 +1,2 @@
+The CBF, when configured to produce the Pulsar Timing data product set and the
```
james-smith-za, line 1: I am still grumpy about the fact that this "Pulsar Timing data product set" is not actually defined anywhere in our requirements so I still have no idea what it really means.

james-smith-za (approved):

qualification/report/requirements_text/CBF-REQ-0095.tex
bmerry: I'm assuming we're going to redo all these files once we get the shiny new set of requirements.

PR #693, opened 2024-02-22
Tweak xbgpu's usage of {X,B}TxQueueItems

amishatishpatel:
As per [this discussion](https://github.com/ska-sa/katgpucbf/pull/650/files#r1365222902) on PR #650, the update proposed here
> [...] ensures the GPU waits, without holding up the Python code.

Apologies for tagging you both to review, I (still) don't know how to remove a reviewer once added.

Whoever gets to it first, please and thank you.

I see there is also potential to update the XPipeline's usage of the `tx_item` in [flush_accumulation](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/xbgpu/engine.py#L696)
* Let me know if I can/should update that line as well.

Closes NGC-1145.

### DIFF
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -431,7 +431,7 @@ class BPipeline(Pipeline[BOutput, BTxQueueItem]):
             await rx_item.async_wait_for_events()
 
             tx_item = await self._tx_free_item_queue.get()
-            await tx_item.async_wait_for_events()
+            tx_item.enqueue_wait_for_events(self._proc_command_queue)
             tx_item.reset(rx_item.timestamp)
 
             # After this point it's too late for set_weights etc to update
@@ -719,7 +719,7 @@ class XPipeline(Pipeline[XOutput, XTxQueueItem]):
                 self.correlation.first_batch = last_batch
 
         tx_item = await self._tx_free_item_queue.get()
-        await tx_item.async_wait_for_events()
+        tx_item.enqueue_wait_for_events(self._proc_command_queue)
 
         # Indicate that the timestamp still needs to be filled in.
         tx_item.timestamp = -1


src/katgpucbf/xbgpu/engine.py
```diff
@@ -431,7 +431,7 @@ async def gpu_proc_loop(self) -> None:  # noqa: D102
             await rx_item.async_wait_for_events()
 
             tx_item = await self._tx_free_item_queue.get()
-            await tx_item.async_wait_for_events()
+            tx_item.enqueue_wait_for_events(self._proc_command_queue)
```
bmerry, line 434: I don't believe this is safe. At line 458 you're modifying tx_item's GPU data from the host (via vkgdr), so the host needs to wait for the GPU to finish the previous work. In practice I think it happens to be safe because the sequencing in sender loop doesn't put the item onto the `_tx_free_item_queue` until it is complete, but no point in planting a land-mine.

Apart from correctness, I think it'll also cause a memory leak because `async_wait_for_events` is the point at which old events get cleaned out of the event list.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -719,7 +719,7 @@ def do_correlation() -> None:
                 self.correlation.first_batch = last_batch
 
         tx_item = await self._tx_free_item_queue.get()
-        await tx_item.async_wait_for_events()
+        tx_item.enqueue_wait_for_events(self._proc_command_queue)
```
bmerry, line 722: I'm pretty sure this is safe, for the trivial reason that this is in the initialisation code and hence the tx_item you get will be freshly minted without prior events anyway. To make a real difference you'd probably need to do the same thing at line 696. I think doing that will be free of race conditions, but will again cause a memory leak as `tx_item.events` will just keep growing (should be easy to check by printing it out each time).

bmerry (changes requested):
I'd recommend just abandoning this. What we have is working, the changes are problematic, and I don't believe they'll improve concurrency at all there won't actually be anything to wait for at these points (due to the way the sender loops happen to be written).

james-smith-za:
I support Bruce's notion.
I guess if it ain't broke, don't fix it. We can revisit if we do run into performance or scaling issues at some stage.

amishatishpatel:
> I support Bruce's notion. I guess if it ain't broke, don't fix it. We can revisit if we do run into performance or scaling issues at some stage.

Noted. Closing this PR and ticket.

PR #694, opened 2024-02-22
Refine xbgpu usage of sample-bits

amishatishpatel:
Specifically use the cmd-line arg to instantiate the CorrelationTemplate. 

Also tweaked the comment accordingly and updated the corresponding unit test.

Closes NGC-1000.

### DIFF
--- a/src/katgpucbf/xbgpu/correlation.py
+++ b/src/katgpucbf/xbgpu/correlation.py
@@ -83,9 +83,6 @@ class CorrelationTemplate:
         n_spectra_per_heap: int,
         input_sample_bits: int,
     ) -> None:
-        if input_sample_bits != 8:
-            raise ValueError(f"input_sample_bits of 8 is only supported, given {input_sample_bits}.")
-
         self.n_ants = n_ants
         self.n_channels = n_channels
         self.n_spectra_per_heap = n_spectra_per_heap
--- a/src/katgpucbf/xbgpu/correlation.py
+++ b/src/katgpucbf/xbgpu/correlation.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2020-2023, National Research Foundation (SARAO)
+# Copyright (c) 2020-2024, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
--- a/src/katgpucbf/xbgpu/correlation.py
+++ b/src/katgpucbf/xbgpu/correlation.py
@@ -73,13 +73,20 @@ class CorrelationTemplate:
         The number of time samples to be processed per frequency channel.
     """
 
-    def __init__(self, context: AbstractContext, n_ants: int, n_channels: int, n_spectra_per_heap: int) -> None:
+    def __init__(
+        self,
+        context: AbstractContext,
+        n_ants: int,
+        n_channels: int,
+        n_spectra_per_heap: int,
+        sample_bitwidth: int,
+    ) -> None:
         self.n_ants = n_ants
         self.n_channels = n_channels
         self.n_spectra_per_heap = n_spectra_per_heap
         self.n_baselines = self.n_ants * (self.n_ants + 1) // 2
 
-        self._sample_bitwidth = 8  # hardcoded to 8 for now, but 4 and 16 bits are also supported
+        self._sample_bitwidth = sample_bitwidth  # hardcoded to 8 upstream, but 4 and 16 bits are also supported
         self._n_ants_per_block = 32  # Hardcoded to 32 for now, but can be set to 32/48/64.
 
         # This 128 is hardcoded in the original Tensor-Core kernel. It loads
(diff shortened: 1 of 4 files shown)


src/katgpucbf/xbgpu/correlation.py
```diff
@@ -73,13 +73,20 @@ class CorrelationTemplate:
...
+        n_spectra_per_heap: int,
+        sample_bitwidth: int,
+    ) -> None:
         self.n_ants = n_ants
         self.n_channels = n_channels
         self.n_spectra_per_heap = n_spectra_per_heap
         self.n_baselines = self.n_ants * (self.n_ants + 1) // 2
 
-        self._sample_bitwidth = 8  # hardcoded to 8 for now, but 4 and 16 bits are also supported
+        self._sample_bitwidth = sample_bitwidth  # hardcoded to 8 upstream, but 4 and 16 bits are also supported
```
bmerry, line 89: It's probably me that wrote the comment, but I think claiming they're "supported" without having any corresponding tests is rather optimistic (I was probably thinking that TCC supports it, rather than this module).

For now I'd suggest raising ValueError if `sample_bitwidth != 8`. There is currently a check at engine.py:1024 that I think can be deleted from there and moved here.

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -73,13 +73,20 @@ class CorrelationTemplate:
         The number of time samples to be processed per frequency channel.
```
bmerry, line 73: Missing documentation of the new parameter.

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -73,13 +73,20 @@ class CorrelationTemplate:
...
     """
 
-    def __init__(self, context: AbstractContext, n_ants: int, n_channels: int, n_spectra_per_heap: int) -> None:
+    def __init__(
+        self,
+        context: AbstractContext,
+        n_ants: int,
+        n_channels: int,
+        n_spectra_per_heap: int,
+        sample_bitwidth: int,
```
bmerry, line 82: I'd rather call this `input_sample_bits` for consistency with fgpu. I'm not too fussed about the `input_` prefix (although it does make it clearer, because the output is 32-bit), but definitely `bits` rather than `bitwidth` for consistency.

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -195,7 +202,10 @@ def __init__(
         )
 
         # TODO: dtypes must depend on input bitwidth
```
bmerry, line 204: I think the point of this PR was to remove this TODO?

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -195,7 +202,10 @@ def __init__(
...
-        self.slots["in_samples"] = accel.IOSlot(dimensions=input_data_dimensions, dtype=np.int8)
+        self.slots["in_samples"] = accel.IOSlot(
+            dimensions=input_data_dimensions,
+            dtype=np.dtype(f"int{self.template._sample_bitwidth}"),
```
bmerry, line 207: It's unfortunately not that simple. numpy doesn't have an int4 type; to support int4 you're going to need to restructure things a lot to remove the N_POLS dimension and instead use a dtype that represents a complex value (see katgpucbf.utils.gaussian_dtype). And the 16-bit support in TCC is (if I recall) float16, not int16, with floating-point output. So I don't think there is a lot of value in this change since it's only correct in one case and in that case the original version was correct anyway. Rather just add an `assert sample_bitwidth == 8` and then use np.int8.

src/katgpucbf/xbgpu/correlation.py
bmerry: I'd also suggest turning it into a public member (no leading `_`) since you're now accessing it from outside the class.

bmerry (changes requested):
Changing the interface to CorrelationTemplate seems reasonable; it allows the command-line argument to actually flow through to where it could later be used. But this change doesn't actually make correlation.py work with anything other than 8. My suggestions would be:

- Move the check that sample_bits == 8 from XBEngine down into CorrelationTemplate
- For any code that clearly isn't going to work with sample_bits != 8 (I've highlighted one in my review), add an assertion that it equals 8 (with a message explaining why), so that if/when someone implements NGC-1106, they get fair warning.
- Not really related to this PR, but I'd suggest updating the command-line argument to specify `choices=[8]` so that if a user tries to invoke it with anything else, they get an error early from the command-line parser.

There is also a TODO referencing NGC-1000 in engine.py that will need to be looked at. It might make sense for the CorrelationTemplate to advertise the dtype it expects.

amishatishpatel:
> Not really related to this PR, but I'd suggest updating the command-line argument to specify choices=[8] so that if a user tries to invoke it with anything else, they get an error early from the command-line parser.

I think this PR has inadvertently become about being more explicit that only 8-bit mode is supported. I'm happy to add that `choices=[8]` in
* I was actually wondering if there was something I could have added in at the top-most level to fail early and fail hard.

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -79,30 +81,33 @@ def __init__(
         n_ants: int,
         n_channels: int,
         n_spectra_per_heap: int,
-        sample_bitwidth: int,
+        input_sample_bits: int,
     ) -> None:
+        if input_sample_bits != 8:
+            raise ValueError(f"input_sample_bits of 8 is only supported, given {input_sample_bits}.")
+
```
bmerry, line 88: I know I'd suggested moving this in here, but it's actually redundant with the exceptions at lines 104-111. Delete one of them (maybe this one, since the existing one is more explicit about whether the value is unimplemented or just wrong.

src/katgpucbf/xbgpu/main.py
```diff
@@ -352,8 +353,6 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
 
     args = parser.parse_args(arglist)
 
-    if args.sample_bits != 8:
-        parser.error("Only 8-bit values are currently supported.")
```
bmerry, line 356: Ah, I hadn't realised this check already existed. Good catch.

bmerry (changes requested):

src/katgpucbf/xbgpu/correlation.py
amishatishpatel: Too right.

PR #699, opened 2024-02-29
Explicitly indicate that AnyStream is a type alias

bmerry:
Mypy complained about this when I ran it by hand - not sure why it doesn't when run under pre-commit.

See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases


amishatishpatel (approved):
Interesting.

PR #701, opened 2024-03-04
Add qualification tests for beamformer delay

bmerry:
There are two tests:
- One uses small delays and cancels them out with the F-engine delays. This is largely to test that the sign conventions, scale etc are consistently implemented in the beamformer compared to the F-engine. It doesn't work on larger delays because the time-domain delay can't be exactly cancelled out in the frequency domain.
- The other uses a wider range of delays, and does not try to cancel out delays in the F-engine. It does show some quantisation effects in the average phase applied, particularly for small delays. I plan to separately experiment with dithering in the beamformer which will hopefully smooth these out. They're probably also exacerbated by having all inputs the same. 

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/14482804/report.pdf)

Contributes to NGC-653.

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -204,6 +204,7 @@ class XBReceiver:
             self.decimation_factor = acv_config["narrowband"]["decimation_factor"]
         else:
             self.decimation_factor = 1
+        self.adc_sample_rate = cbf.config["outputs"][acv_config["src_streams"][0]]["adc_sample_rate"]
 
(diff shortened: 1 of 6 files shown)


amishatishpatel (approved):
Just the one comment that certainly isn't overly important to address.

I don't want to say "The tests seem simple enough" because everything seems simple enough once it's done.

amishatishpatel (approved):

PR #703, opened 2024-03-06
Improvements to layering in the docker build

bmerry:
The main improvement is that the package itself is managed in a separate layer to the dependencies, so that changes to the source code will only make small new layers to push instead of a fresh copy of all the dependencies. That's achieved by having a build layer generate a wheel, which is copied into the final FROM and installed. That does waste some space in that final FROM (the wheel is deleted afterwards but still requires a layer).

Also simplified the way the right version of pycuda is installed, and use COPY --link because the docs say that's what one generally wants.

Closes NGC-1255.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,7 @@ RUN SPEAD2_VERSION=$(grep ^spead2== katgpucbf/requirements.txt | cut -d= -f3) &&
 # Install and immediately uninstall pycuda. This causes pip to cache the
 # wheel it built, making it fast to install later (we uninstall so that the
 # Jenkins image has a clean environment to start from).
-RUN pip install --no-deps "$(grep '^pycuda==' /tmp/katgpucbf/requirements.txt)" && \
+RUN pip install --no-deps -c /tmp/katgpucbf/requirements.txt pycuda && \
     pip uninstall -y pycuda
 
 #######################################################################
@@ -124,7 +124,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
 
 # The above image is independent of the contents of this package (except
 # for requirements.txt), and is used to form the image for Jenkins
-# testing. We now install the package in a new build stage.
+# testing. We now install the requirements in a new build stage.
 
 FROM build-base as build-py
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -124,19 +124,24 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
...
+#######################################################################
+
+FROM build-base as build-wheel
+
+# Build a wheel for the package. Note that this happens independently
+# of the build-py image.
+WORKDIR /tmp/katgpucbf
 COPY . .
-RUN pip install --no-deps . && pip check
+RUN pip install -c requirements-dev.txt build
```
james-smith-za, line 143: Since you've added build to `requirements-dev.in`/`.txt`, I don't see the reason for it to be on this line here.

Dockerfile
```diff
@@ -124,19 +124,24 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
...
+RUN python -m build --wheel
```
james-smith-za, line 144: I'm on board with separating the dependencies from the actual install, but I'm not quite clear on why a wheel needs to be built.

Can't you just move the old line 139 (`pip install --no-deps .`) to 201 where you've currently got pip installing the wheel?

What advantages does it bring over the old way?

Dockerfile
```diff
@@ -126,22 +126,21 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
...
 # Install requirements (already copied to build-base image).
 WORKDIR /tmp/katgpucbf
 RUN pip install -r requirements.txt
 
 #######################################################################
 
-FROM build-base as build-wheel
+FROM build-base as build-py
 
 # Build a wheel for the package. Note that this happens independently
```
james-smith-za, line 139: Not building a wheel anymore.

james-smith-za (changes requested):
One nitpick.

bmerry:
> I guess my overall question though, and this is more from lack of understanding than anything else, is why we need the `build-py` stage is needed at all. What do we gain by doing it this way rather than just moving this operation to the very bottom of the file?

We avoid placing a bunch of things into the final image:
- a copy of the entire source tree (including any non-git files that happened to be lying around e.g. pytest output, mypy cache etc).
- cache files that pip would download as part of running `pip install` (e.g. the packages in build-requires)
- any tools needed at build time (e.g. if we ever wrote modules in C++ again, we'd need compilers for those)

james-smith-za (approved):
That does make sense. Thanks for clarifying.

PR #706, opened 2024-03-06
Add test that beamformer quantisation gives 8-bit output

bmerry:
The dsim_gaussian function from test_delay.py is generalised and moved up to `__init__.py` so that it can be reused.

Closes NGC-1260

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/14510649/report.pdf)

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -48,9 +48,11 @@ from spead2.numba import intp_to_voidptr
 from spead2.recv.numba import chunk_place_data
 
 import katgpucbf.recv
-from katgpucbf import COMPLEX
+from katgpucbf import COMPLEX, DIG_SAMPLE_BITS
 from katgpucbf.spead import DEFAULT_PORT, FREQUENCY_ID, TIMESTAMP_ID
 from katgpucbf.utils import TimeConverter
+
+from .reporter import Reporter as _Reporter
 
 logger = logging.getLogger(__name__)
 DEFAULT_MAX_DELAY = 1000000  # Around 0.5-1ms, depending on band. Increase if necessary
@@ -182,6 +184,28 @@ class CBFRemoteControl:
         """
         reply, _ = await self.dsim_clients[dsim_idx].request("time")
         return aiokatcp.decode(float, reply[0])
+
+    async def dsim_gaussian(self, amplitude: float, pdf_report: _Reporter | None = None, *, dsim_idx: int = 0) -> None:
+        """Configure a dsim with Gaussian noise.
+
+        The identical signal is produced on both polarisations.
+
+        Parameters
+        ----------
+        amplitude
+            Standard deviation, in units of the LSB of the digitiser output
+        pdf_report
+            Reporter to which this process will be reported
+        dsim_idx
+            Index of the dsim to set
+        """
+        if pdf_report is not None:
+            pdf_report.step("Configure the D-sim with Gaussian noise.")
+        dig_max = 2 ** (DIG_SAMPLE_BITS - 1) - 1
+        amplitude /= dig_max  # Convert to be relative to full-scale
+        await self.dsim_clients[0].request("signals", f"common=nodither(wgn({amplitude}));common;common;")
+        if pdf_report is not None:
+            pdf_report.detail(f"Set D-sim with wgn amplitude={amplitude}.")
 
 
 class XBReceiver:
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -52,7 +52,7 @@ from katgpucbf import COMPLEX, DIG_SAMPLE_BITS
 from katgpucbf.spead import DEFAULT_PORT, FREQUENCY_ID, TIMESTAMP_ID
 from katgpucbf.utils import TimeConverter
 
-from .reporter import Reporter as _Reporter
+from .reporter import Reporter
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


qualification/tied_array_channelised_voltage/test_delay.py
```diff
@@ -21,24 +21,12 @@
...
 from .. import CBFRemoteControl, TiedArrayChannelisedVoltageReceiver
 from ..reporter import POTLocator, Reporter
 
 
-async def dsim_gaussian(cbf: CBFRemoteControl, pdf_report: Reporter, amplitude: float) -> None:
-    """Configure the dsim with Gaussian noise."""
-    pdf_report.step("Configure the D-sim with Gaussian noise.")
-    dig_max = 2 ** (DIG_SAMPLE_BITS - 1) - 1
-    # Small so that we don't saturate, as delaying and undelaying a saturated
-    # value won't round-trip properly
```
amishatishpatel, line 36: Did you purposefully remove this comment/note *during the move? I thought it seemed quite useful - maybe your reasoning changed?

EDIT: Although looking at the usages later on, I see the same or similar comment above each call, so 🤷

qualification/tied_array_channelised_voltage/test_precision.py
```diff
@@ -0,0 +1,59 @@
...
+    pdf_report.step("Set weights to select a single input.")
+    weights = [0.0] * len(receiver.source_indices[0])
+    weights[0] = 1.0
+    await client.request("beam-weights", receiver.stream_names[0], *weights)
+    pdf_report.detail(f"Weights for {receiver.stream_names[0]} set to {weights}.")
+
+    pdf_report.step("Collect and validate a chunk.")
+    timestamp, data = await receiver.next_complete_chunk()
+    pdf_report.detail(f"Received chunk with timestamp {timestamp}.")
+    unique_values = np.unique(data)
```
amishatishpatel, line 55: Like London buses, I'd never known of/used the `unique` command, and now I've seen two in succession - here and last week during a check-in.

amishatishpatel (approved):
What a neat test. I can't see anything overtly questionable. Just the one comment on your renovations of `dsim_gaussian` - otherwise good to go.

qualification/__init__.py
```diff
@@ -48,10 +48,12 @@
 from spead2.recv.numba import chunk_place_data
 
 import katgpucbf.recv
-from katgpucbf import COMPLEX
+from katgpucbf import COMPLEX, DIG_SAMPLE_BITS
 from katgpucbf.spead import DEFAULT_PORT, FREQUENCY_ID, TIMESTAMP_ID
 from katgpucbf.utils import TimeConverter
 
+from .reporter import Reporter as _Reporter
```
amishatishpatel, line 55: I should've also asked, why underscore-Reporter?

qualification/tied_array_channelised_voltage/test_delay.py
bmerry: Yes, it shouldn't have been in the function in the first place, because within the function the amplitude could be small or large. It's the call sites where a small amplitude is chosen.

qualification/__init__.py
bmerry: I didn't want code to start writing `from . import Reporter` in place of `from .reporter import Reporter` i.e. I didn't want it to become part of the public API of this file. It's not really consistent though because e.g. TimeConverter ends up being part of the public API. I guess the difference is that TimeConverter is not something you'd likely assume was part of the public API (because it's defined in a completely different place), whereas Reporter is part of the qualification tests.

I'm not committed to it. @james-smith-za do you have an opinion?

amishatishpatel: > I guess the difference is that TimeConverter is not something you'd likely assume was part of the public API (because it's defined in a completely different place), whereas Reporter is part of the qualification tests.

My thoughts exactly. With this overall reasoning (especially `from .reporter` instead of `from . import`), I'm happy for it to stay as `_Reporter` (underscore-Reporter). However, 
* does that mean we need to update other usages under `qualification/` (or at this level in the dir)? Or
* Can everything stay the way it is, and we'll use this going forward?

qualification/tied_array_channelised_voltage/test_precision.py
james-smith-za: That one was `uniq` though which is part of GNU coreutils ;-)

qualification/__init__.py
james-smith-za: I have mixed feelings about this.

- I don't think it's particularly important for now. There is no one really using this repo except us, so "Public API" isn't something that we need to be overly concerned about.
- That being said though, there is a TODO comment from yours truly at the top of the file which talks about perhaps there's enough in this file that it warrants its own module.

It looks like the stuff declared in this file gets imported in the qualification tests here:
https://github.com/ska-sa/katgpucbf/blob/f2da9667cf3e46b5a63f1d85cf810c8b26b55f56/qualification/conftest.py#L41
So it may be worth a technical debt ticket to move the bulk of this code into their own modules, and then have the `__init__.py` mostly blank and explicitly controlling what we want to be in the public API which I think will likely be very little. It should probably just exist so that we can do relative imports (though I'm a bit hazy on how that works, it's been a while).

bmerry: True, "public" users are really just the individual qualification tests, rather than code in other git repositories, so we're in control of the usage. I'll change it to get rid of the underscore.

amishatishpatel (approved):

PR #707, opened 2024-03-07
Add a test that the beam can be steered fast enough

bmerry:
Closes NGC-1259.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/14521338/report.pdf)

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_delay.py
+++ b/qualification/tied_array_channelised_voltage/test_delay.py
@@ -211,7 +211,7 @@ async def test_delay(
         pdf_report.figure(fig)
 
 
-@pytest.mark.requirements("CBF-REQ-0077")
+@pytest.mark.requirements("CBF-REQ-0076")
 async def test_delay_update_time(
     cbf: CBFRemoteControl,
     receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver,
--- a/qualification/tied_array_channelised_voltage/test_delay.py
+++ b/qualification/tied_array_channelised_voltage/test_delay.py
@@ -16,6 +16,8 @@
 
 """Delay test."""
 
+import time
+
 import numpy as np
 import pytest
 from matplotlib.figure import Figure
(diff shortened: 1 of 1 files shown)


qualification/tied_array_channelised_voltage/test_delay.py
```diff
@@ -207,3 +209,40 @@ async def test_delay(
...
+    Verification method
+    -------------------
+    Verified by means of test. Measure the time taken to issue
+    ``?beam-delays`` requests for all beams.
+    """
+    receiver = receive_tied_array_channelised_voltage
+    client = cbf.product_controller_client
+
+    # Use random values just because they'll have lots of significant digits
+    # and hence take longer to format/parse.
```
amishatishpatel, line 231: Noted.

qualification/tied_array_channelised_voltage/test_delay.py
```diff
@@ -207,3 +209,40 @@ async def test_delay(
...
+    start = time.monotonic()
+    pdf_report.detail(f"Start time is {start:.6f}.")
+    for stream_name, stream_params in zip(receiver.stream_names, params):
+        await client.request("beam-delays", stream_name, *stream_params)
+        pdf_report.detail(f"Set delays on {stream_name} to {stream_params}")
+    stop = time.monotonic()
+    pdf_report.detail(f"Stop time is {stop:.6f}.")
+    elapsed = stop - start
+    pdf_report.detail(f"Elapsed time is {elapsed:.6f}.")
+    assert elapsed < 5.0
```
amishatishpatel, line 248: From the requirement text
> [...] and the execution time delay is no more than 2 seconds.

Not to be so blunt, but you've got 5.0 here? Let me know if there's a separate doc/ECP/something I should look at that overrides the 2s mentioned in the REQ.

amishatishpatel (commented):
As neat as this test is, just the one (slightly major) comment for clarification before I can explicitly approve.

qualification/tied_array_channelised_voltage/test_delay.py
bmerry: I see I linked to the wrong requirement. It should have been CBF-REQ-0076:

> The CBF shall execute all of the commands required to re-point a TA beam
utilising H and V polarisation of all available antennae in less than 5
seconds.

amishatishpatel: Aha, yeah the whole context behind REQ-0077 seemed a bit odd here (with the tracking of the request 200ms _before_ execution time).

Either way, please and thank you.

amishatishpatel (approved):

PR #708, opened 2024-03-08
Add support for Beam Prometheus counters

amishatishpatel:
Also add verification in the `test_engine_end_to_end` unit test. I'm not entirely happy with the
workarounds I've had to put in place in `test_engine.py::verify_beam_sensors`. I've tried to 
explain my through/out of it, but feel free to be loudly unhappy with it (with a suggestion or
two, please).

I placed the counter increments in a callback because
1. I figured the `chunk.send` needs to be as uninterrupted as possible,
  a. And the updating of sensors isn't as time-critical as sending of data
2. Of course, I took inspiration from fgpu.

Closes NGC-1154.

### DIFF
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -174,7 +174,7 @@ class Chunk:
 
     @staticmethod
     def _inc_counters(
-        data_shape: tuple[int, int, int], data_type: np.dtype, output_names: Sequence[str], future: asyncio.Future
+        data_shape: tuple[int, int, int], data_dtype: np.dtype, output_names: Sequence[str], future: asyncio.Future
     ) -> None:
         """Increment beam stream Prometheus counters.
 
@@ -183,7 +183,7 @@ class Chunk:
         data_shape
             The shape of the beam data being transmitted. Expected in the
             format of (n_channels_per_substream, samples_per_spectra, COMPLEX).
-        data_type
+        data_dtype
             The `np.dtype` of the beam data transmitted.
         output_names
             List of beam stream names that are enabled for transmission for
@@ -199,7 +199,7 @@ class Chunk:
             for output_name in output_names:
                 output_heaps_counter.labels(output_name).inc(1)
                 # Multiply across dimensions to get total bytes
-                output_bytes_counter.labels(output_name).inc(np.prod(data_shape) * data_type.itemsize)
+                output_bytes_counter.labels(output_name).inc(np.prod(data_shape) * data_dtype.itemsize)
                 # Multiply across the first two dimensions to get complex sample count
                 output_samples_counter.labels(output_name).inc(np.prod(data_shape[:-1]))
 
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -193,7 +193,6 @@ class Chunk:
         """
         if not future.cancelled() and future.exception() is None:
             for beam_name in beam_names:
-                logger.info(f"\nCallback updating for {beam_name}")
                 output_heaps_counter.labels(beam_name).inc(1)
                 # Each beam data sample is 8-bit
                 # - Multiply across dimensions to get total bytes
@@ -232,12 +231,6 @@ class Chunk:
                 send_futures.append(
                     send_stream.stream.async_send_heaps(heaps_to_send, mode=spead2.send.GroupMode.ROUND_ROBIN)
                 )
-                enabled_streams = [
-                    output_name
-                    for output_name, enabled in zip(send_stream.beam_names, send_stream.tx_enabled)
-                    if enabled
-                ]
-                logger.info(f"Adding callback for {enabled_streams} with {frame.data.shape}")
                 send_futures[-1].add_done_callback(
                     functools.partial(
                         self._inc_counters,
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -174,7 +174,7 @@ class Chunk:
 
     @staticmethod
     def _inc_counters(
-        data_shape: tuple[int, int, int], beam_dtype: np.dtype, beam_names: Sequence[str], future: asyncio.Future
+        data_shape: tuple[int, int, int], data_type: np.dtype, output_names: Sequence[str], future: asyncio.Future
     ) -> None:
         """Increment beam stream Prometheus counters.
 
@@ -183,7 +183,9 @@ class Chunk:
         data_shape
             The shape of the beam data being transmitted. Expected in the
             format of (n_channels_per_substream, samples_per_spectra, COMPLEX).
-        beam_names
+        data_type
+            The `np.dtype` of the beam data transmitted.
+        output_names
             List of beam stream names that are enabled for transmission for
             this `future`.
         future
@@ -194,13 +196,12 @@ class Chunk:
             (NGC-1172) Handle missing data transmission for Prometheus counters.
         """
         if not future.cancelled() and future.exception() is None:
-            for beam_name in beam_names:
-                output_heaps_counter.labels(beam_name).inc(1)
-                # Each beam data sample is 8-bit
-                # - Multiply across dimensions to get total bytes
-                output_bytes_counter.labels(beam_name).inc(np.prod(data_shape) * beam_dtype.itemsize)
-                # - Multiply across the first two dimensions to get complex sample count
-                output_samples_counter.labels(beam_name).inc(np.prod(data_shape[:-1]))
+            for output_name in output_names:
+                output_heaps_counter.labels(output_name).inc(1)
+                # Multiply across dimensions to get total bytes
+                output_bytes_counter.labels(output_name).inc(np.prod(data_shape) * data_type.itemsize)
+                # Multiply across the first two dimensions to get complex sample count
+                output_samples_counter.labels(output_name).inc(np.prod(data_shape[:-1]))
 
     def send(
         self,
@@ -223,7 +224,7 @@ class Chunk:
         if n_enabled > 0:
             send_futures: list[asyncio.Future] = []
             enabled_stream_names = [
-                output_name for output_name, enabled in zip(send_stream.beam_names, send_stream.tx_enabled) if enabled
+                output_name for output_name, enabled in zip(send_stream.output_names, send_stream.tx_enabled) if enabled
             ]
             for frame in self._frames:
                 # TODO (NGC-1232): building this list every time may be too expensive.
@@ -325,7 +326,7 @@ class BSend:
 
         self.tx_enabled = [tx_enabled] * len(outputs)
         n_beams = len(outputs)
-        self.beam_names = [output.name for output in outputs]
+        self.output_names = [output.name for output in outputs]
 
         self._chunks_queue: asyncio.Queue[Chunk] = asyncio.Queue()
         buffers: list[np.ndarray] = []
@@ -446,13 +447,13 @@ class BSend:
         # It's a heavy-handed approach, but we don't care about performance
         # during shutdown.
         await self.stream.async_flush()
-        for i in range(len(self.beam_names)):
+        for i in range(len(self.output_names)):
             await self.stream.async_send_heap(stop_heap, substream_index=i)
 
 
 def make_stream(
     *,
-    beam_names: list[str],
+    output_names: list[str],
     endpoints: list[Endpoint],
     interface: str,
     ttl: int,
@@ -493,10 +494,10 @@ def make_stream(
         )
     # Referencing the labels causes them to be created, in advance of data
     # actually being transmitted.
-    for beam_name in beam_names:
-        output_heaps_counter.labels(beam_name)
-        output_bytes_counter.labels(beam_name)
-        output_samples_counter.labels(beam_name)
-        output_clip_counter.labels(beam_name)
+    for output_name in output_names:
+        output_heaps_counter.labels(output_name)
+        output_bytes_counter.labels(output_name)
+        output_samples_counter.labels(output_name)
+        output_clip_counter.labels(output_name)
 
     return stream
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -220,6 +220,9 @@ class Chunk:
         rate = send_stream.bytes_per_second_per_beam * n_enabled
         if n_enabled > 0:
             send_futures: list[asyncio.Future] = []
+            enabled_stream_names = [
+                output_name for output_name, enabled in zip(send_stream.beam_names, send_stream.tx_enabled) if enabled
+            ]
             for frame in self._frames:
                 # TODO (NGC-1232): building this list every time may be too expensive.
                 # Consider caching it and invalidating when streams are enabled/disabled.
@@ -235,11 +238,7 @@ class Chunk:
                     functools.partial(
                         self._inc_counters,
                         frame.data.shape[1:],  # Get rid of 'beam' dimension
-                        [
-                            output_name
-                            for output_name, enabled in zip(send_stream.beam_names, send_stream.tx_enabled)
-                            if enabled
-                        ],
+                        enabled_stream_names,
                     )
                 )
 
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -17,6 +17,7 @@
 """Module for sending tied array channelised voltage products onto the network."""
 
 import asyncio
+import functools
 import logging
 from math import ceil
 from typing import Callable, Final, Sequence
@@ -171,6 +172,35 @@ class Chunk:
         self._timestamps += delta
         self._timestamp = value
 
+    @staticmethod
+    def _inc_counters(beam_data_shape: tuple[int, int, int], beam_names: Sequence[str], future: asyncio.Future) -> None:
+        """Increment beam stream Prometheus counters.
+
+        Parameters
+        ----------
+        beam_data_shape
+            The shape of the beam data being transmitted. Expected in the
+            format of (n_channels_per_substream, samples_per_spectra, COMPLEX).
+        beam_names
+            List of beam stream names that are enabled for transmission for
+            this `future`.
+        future
+            Future returned by the spead2 stream's `async_send_heaps`.
+
+        .. todo::
+
+            (NGC-1172) Handle missing data transmission for Prometheus counters.
+        """
+        if not future.cancelled() and future.exception() is None:
+            for beam_name in beam_names:
+                logger.info(f"\nCallback updating for {beam_name}")
+                output_heaps_counter.labels(beam_name).inc(1)
+                # Each beam data sample is 8-bit
+                # - Multiply across dimensions to get total bytes
+                output_bytes_counter.labels(beam_name).inc(np.prod(beam_data_shape))
+                # - Multiple across the first two dimensions to get complex sample count
+                output_samples_counter.labels(beam_name).inc(np.prod(beam_data_shape[:-1]))
+
     def send(
         self,
         send_stream: "BSend",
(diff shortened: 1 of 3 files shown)


src/katgpucbf/xbgpu/bsend.py
```diff
@@ -267,17 +322,18 @@ def __init__(
             raise ValueError("channel_offset must be an integer multiple of n_channels_per_substream")
 
         self.tx_enabled = [tx_enabled] * len(outputs)
-        self.n_beams = len(outputs)
+        n_beams = len(outputs)
+        self.beam_names = [output.name for output in outputs]
```
amishatishpatel, line 326: I made the beam names an attribute of the BSend as `Chunk.send` doesn't get called directly from the BPipeline
* It gets called from `BSend.send_chunk`

test/xbgpu/test_engine.py
```diff
@@ -271,12 +271,12 @@ def valid_end_to_end_combination(combo: dict) -> bool:
...
     corrprod_results: list[np.ndarray],
-    missing_antenna: int | None,
-    incomplete_accumulation_counters: list[int],
     prom_diff: PromDiff,
+    actual_sensor_updates: dict[str, list[tuple[Any, aiokatcp.Sensor.Status]]],
+    incomplete_accumulation_counters: list[int],
     n_channels_per_substream: int,
     n_baselines: int,
-    actual_sensor_updates: dict[str, list[tuple[Any, aiokatcp.Sensor.Status]]],
+    missing_antenna: int | None,
```
amishatishpatel, line 279: Yeah apologies for the seemingly unnecessary reshuffle of the arglist. Something just wasn't sitting right about the order.

james-smith-za (commented):
I am happy with the overall direction.

It's been a while since I've looked at this code in anger though so I'd be more confident once @bmerry casts an eye.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
 ) -> None:
-    """Verify katcp sensors for BPipeline data.
+    """Verify katcp sensors and Prometheus counters for BPipeline data.
+
+    NOTE: There is currently logic in place to compensate for the BPipeline's
+    lack of tracking missing data. More specifically, it compensates for the
+    difference in expected versus measured counter values by seemingly
+    'forcing' them to match. In actuality, it's counteracting the BPipeline's
+    BSend transmitting `HEAPS_PER_FENGINE_PER_CHUNK` heaps regardless of how
+    empty/full the incoming Chunk is.
```
bmerry, line 366: I didn't find that explanation entirely clear. It doesn't specifically refer to the final chunk, which is the one causing the problems.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
     ----------
     beam_outputs
         Output beam configurations parsed into BOutput objects.
-    first_timestamp, last_timestamp
-        Two timestamps indicating the start and end of data processing
-        by the :class:`BPipeline`.
+    beam_results_shape
+        The shape of the verified beam data for all beams with the
+        following shape
+        - (len(beam_outputs), n_beam_heaps_sent, n_channels_per_substream,
```
bmerry, line 375: Is the `-` intended to be a list bullet? I don't think it'll render that way without more whitespace.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
-        is implemented.
     """
-    # Generate expected sensor updates from BPipeline and compare
+    # Get the number of total heaps transmitted by each beam output
+    # NOTE: We round up to the nearest multiple of `HEAPS_PER_FENGINE_PER_CHUNK`
+    # as the BSend currently transmits all Frames of data, regardless of how
+    # empty/full an input (received Chunk) is. Also, the number of Frames in a
+    # BSend Chunk is the same as `HEAPS_PER_FENGINE_PER_CHUNK`.
+    n_beam_heaps_sent = beam_results_shape[1]
+    n_beam_heaps_sent += n_beam_heaps_sent % HEAPS_PER_FENGINE_PER_CHUNK
```
bmerry, line 404: That doesn't look right for rounding up to a multiple of `HEAPS_PER_FENGINE_PER_CHUNK` - it just happens to work for the particular values used in this test. The easiest option is to use `katsdpsigproc.accel.roundup`.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
-
-        Add verification of Prometheus counters once NGC-1154
-        is implemented.
     """
-    # Generate expected sensor updates from BPipeline and compare
+    # Get the number of total heaps transmitted by each beam output
+    # NOTE: We round up to the nearest multiple of `HEAPS_PER_FENGINE_PER_CHUNK`
+    # as the BSend currently transmits all Frames of data, regardless of how
+    # empty/full an input (received Chunk) is. Also, the number of Frames in a
+    # BSend Chunk is the same as `HEAPS_PER_FENGINE_PER_CHUNK`.
```
bmerry, line 402: You've got one comment here explaining this and another in the docstring. Pick just one place to explain it. I'd suggest a comment here, since the docstring should document how to call the function, and this is more of an implementation detail.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
+    n_beam_heaps_sent = beam_results_shape[1]
+    n_beam_heaps_sent += n_beam_heaps_sent % HEAPS_PER_FENGINE_PER_CHUNK
+    beam_data_shape = beam_results_shape[2:]
```
bmerry, line 405: Maybe call this `heap_shape`? The `beam` prefix doesn't really convey any information.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
+    # NOTE: Explicitly cast np.prod returns to int as the default is np.int64
+    heap_bytes = int(np.prod(beam_data_shape))
+    # We get rid of the final (COMPLEX) dimension in the beam data as each
+    # sample has a bitwidth of eight (8).
+    heap_samples = int(np.prod(beam_data_shape[:-1]))
```
bmerry, line 410: The bitwidth has nothing to do with it. We get rid of it because a "sample" is a complex value rather than the individual real and imaginary parts.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
+    # Get the number of total heaps transmitted by each beam output
+    # NOTE: We round up to the nearest multiple of `HEAPS_PER_FENGINE_PER_CHUNK`
+    # as the BSend currently transmits all Frames of data, regardless of how
+    # empty/full an input (received Chunk) is. Also, the number of Frames in a
+    # BSend Chunk is the same as `HEAPS_PER_FENGINE_PER_CHUNK`.
+    n_beam_heaps_sent = beam_results_shape[1]
+    n_beam_heaps_sent += n_beam_heaps_sent % HEAPS_PER_FENGINE_PER_CHUNK
+    beam_data_shape = beam_results_shape[2:]
+    # NOTE: Explicitly cast np.prod returns to int as the default is np.int64
+    heap_bytes = int(np.prod(beam_data_shape))
```
bmerry, line 407: This assumes each element is 1 byte. Maybe pass in the dtype and multiple through by its itemsize.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
+    # We get rid of the final (COMPLEX) dimension in the beam data as each
+    # sample has a bitwidth of eight (8).
+    heap_samples = int(np.prod(beam_data_shape[:-1]))
     for i, beam_output in enumerate(beam_outputs):
+        # `prom_diff.get_sample_diff` can return a range of data types. This
```
bmerry, line 412: Really? It should only ever return float or None.

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
+        if prom_output_b_heaps_total != n_beam_heaps_sent:
+            # NOTE: The output_b_{heaps, bytes, samples} counters still think
+            # `n-1` heaps were sent, so we need to increment those over by one
+            # heap's worth of data respectively.
+            heap_count_diff = n_beam_heaps_sent - prom_output_b_heaps_total
+            # Add the mod difference to the heaps counter so it seems less
+            # 'hacked' than simply adding the difference between the two values.
+            prom_output_b_heaps_total += prom_output_b_heaps_total % HEAPS_PER_FENGINE_PER_CHUNK
+            prom_output_b_bytes_total += heap_count_diff * heap_bytes
+            prom_output_b_samples_total += heap_count_diff * heap_samples
```
bmerry, line 435: This block has a big smell to it. Is the number of heaps sent non-deterministic (varies from run to run)? If not, then rather start by computing n_beam_heaps_sent correctly e.g. maybe it needs to be rounded down rather than rounded up at the start?

test/xbgpu/test_engine.py
```diff
@@ -971,8 +1028,19 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
             for j in range(beam_results.shape[1]):
                 np.testing.assert_allclose(beam_results[i, j], expected_beams[i, j], atol=1)
 
+        # `beam_results` hold results for each heap transmitted by a `beam_output`
```
bmerry, line 1031:
```suggestion
        # `beam_results` holds results for each heap transmitted by a `beam_output`
```

test/xbgpu/test_engine.py
```diff
@@ -347,41 +347,98 @@ def verify_corrprod_sensors(
...
         names, the values are a list of tuples for each sensor update captured
         via the callback attached to :class:`BPipeline` sensors. Accommodating
         for three value types as there are three different types of sensors in
         the XBEngine.
+    first_timestamp, last_timestamp
+        Two timestamps indicating the start and end of data processing
+        by the :class:`BPipeline`.
+    n_heaps_total
+        Total number of heaps transmitted by the beam data stream - including
+        descriptors.
```
bmerry, line 391: This doesn't appear in the function signature.

test/xbgpu/test_engine.py
amishatishpatel: Of course now when I try to recreate the `mypy` error it doesn't show.

Maybe I misread a plain old python error regarding casting to int? e.g. Removing a `type: ignore`
* `Argument 1 to "int" has incompatible type "float | None"; expected "str | Buffer | SupportsInt | SupportsIndex | SupportsTrunc"  [arg-type]`, and
* The arithmetic operations complain about `Unsupported operand types for {+, -} ("int" and "None")`

I'll fiddle with my syntax to see if I can solve it. Of course, open to suggestions.

bmerry: If you just assert that it is not None I would imagine that should do it.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -156,6 +172,34 @@ def timestamp(self, value: int) -> None:
...
+        .. todo::
+
+            (NGC-1172) Handle missing data transmission for Prometheus counters.
+        """
+        if not future.cancelled() and future.exception() is None:
+            for beam_name in beam_names:
+                output_heaps_counter.labels(beam_name).inc(1)
+                # Each beam data sample is 8-bit
+                # - Multiply across dimensions to get total bytes
+                output_bytes_counter.labels(beam_name).inc(np.prod(beam_data_shape))
```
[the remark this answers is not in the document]
amishatishpatel: Apologies, I seem to have missed this one. Getting on it now.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -173,7 +173,9 @@ def timestamp(self, value: int) -> None:
         self._timestamp = value
 
     @staticmethod
-    def _inc_counters(beam_data_shape: tuple[int, int, int], beam_names: Sequence[str], future: asyncio.Future) -> None:
+    def _inc_counters(
+        beam_data_shape: tuple[int, int, int], beam_dtype: np.dtype, beam_names: Sequence[str], future: asyncio.Future
```
bmerry, line 177: Bikeshedding: I'm not sure it's really necessary to prefix everything with "beam_". This is bsend.py, so everything is implicitly beam-related.

test/xbgpu/test_engine.py
```diff
@@ -358,22 +362,17 @@ def verify_beam_sensors(
 ) -> None:
     """Verify katcp sensors and Prometheus counters for BPipeline data.
 
-    NOTE: There is currently logic in place to compensate for the BPipeline's
-    lack of tracking missing data. More specifically, it compensates for the
-    difference in expected versus measured counter values by seemingly
-    'forcing' them to match. In actuality, it's counteracting the BPipeline's
-    BSend transmitting `HEAPS_PER_FENGINE_PER_CHUNK` heaps regardless of how
-    empty/full the incoming Chunk is.
-
```
bmerry, line 358: This function is looking much better now that the hackery is gone, thanks.

test/xbgpu/test_engine.py
```diff
@@ -695,8 +668,15 @@ async def _send_data(
 
                 corrprod_results[i][j] = ig_recv["xeng_raw"].value
 
+        # NOTE: Update `batch_indices` to be end on a multiple of
+        # `HEAPS_PER_FENGINE_PER_CHUNK`, but only for the beam_outputs because
+        # they currently send `HEAPS_PER_FENGINE_PER_CHUNK` heaps all the time.
+        # This does not mean the final heap (for each beam_output) has sane
+        # data in it. In fact, ensure you verify data up to `n_beam_heaps - 1`.
```
bmerry, line 675: Why `n_beam_heaps - 1`? Shouldn't one just verify the values in `batch_indices`?

test/xbgpu/test_engine.py
```diff
@@ -695,8 +668,15 @@ async def _send_data(
...
+        n_beam_heaps = roundup(len(batch_indices), HEAPS_PER_FENGINE_PER_CHUNK)
+        beam_batch_indices = range(batch_indices[0], batch_indices[0] + n_beam_heaps)
```
bmerry, line 677: I think this logic only works if `batch_indices` is contiguous AND starts on a multiple of HEAPS_PER_FENGINE_PER_CHUNK. I don't think it's worth trying to handle the general case in this PR, although it should does need to be tested under NGC-1172. Just add a comment to document the assumption, and maybe an assertion too so that the computer will remind you when it is violated.

test/xbgpu/test_engine.py
```diff
@@ -1025,15 +1005,20 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
         # assert_allclose converts to float, which bloats memory usage.
         # To keep it manageable, compare a batch at a time.
         for i in range(beam_results.shape[0]):
-            for j in range(beam_results.shape[1]):
+            # NOTE: As per the explanation at the end of `_send_data`, we
+            # only verify up the the penultimate heap's data for each `beam_result`
+            # because the final heap is sent coincidentally - not because it is
+            # expected to have sane data in it.
+            for j in range(beam_results.shape[1] - 1):
```
bmerry, line 1012: Again, you're hardcoding the assumption about how must extra is sent. How about just iterating up to the length of `expected_beams`?

bmerry (changes requested):

src/katgpucbf/xbgpu/bsend.py
amishatishpatel: Too right. I think I was still in the zone of working in `test_engine`, so I felt like I needed to keep the convention between these unrelated files.

test/xbgpu/test_engine.py
amishatishpatel: Argh, too right. Both here and in your last comment, I'd made a mental `TODO` to update this explanation for something less 'hardcoded'.

Apologies, will update now-now.

test/xbgpu/test_engine.py
amishatishpatel: Once again, too right. I completely glossed over the mention in the parameter's docstring.

Will update accordingly.

test/xbgpu/test_engine.py
```diff
@@ -638,8 +668,23 @@ async def _send_data(
...
                 corrprod_results[i][j] = ig_recv["xeng_raw"].value
 
+        # TODO: NGC-1172 The tweaks to process beam data below rely on
+        # `batch_indices` to be contiguous. The check below is temporary until
+        # the BPipeline is able to handle missing data.
+        assert sorted(batch_indices) == (
+            list(range(min(batch_indices), max(batch_indices) + 1))
+        ), "Batch indices need to be contiguous for testing beam data"
+
+        # NOTE: Update `batch_indices` to be end on a multiple of
```
bmerry, line 678:
```suggestion
        # NOTE: Update `batch_indices` to end on a multiple of
```

test/xbgpu/test_engine.py
```diff
@@ -638,8 +668,23 @@ async def _send_data(
 
                 corrprod_results[i][j] = ig_recv["xeng_raw"].value
 
+        # TODO: NGC-1172 The tweaks to process beam data below rely on
+        # `batch_indices` to be contiguous. The check below is temporary until
+        # the BPipeline is able to handle missing data.
+        assert sorted(batch_indices) == (
```
bmerry, line 674: Remove the `sorted`. You've got assumptions below that e.g. `batch_indices[0]` is the smallest index.

test/xbgpu/test_engine.py
```diff
@@ -638,8 +668,23 @@ async def _send_data(
 
                 corrprod_results[i][j] = ig_recv["xeng_raw"].value
 
+        # TODO: NGC-1172 The tweaks to process beam data below rely on
+        # `batch_indices` to be contiguous. The check below is temporary until
```
bmerry, line 672: If I'm not mistaken, it also depends on `batch_indices[0]` being aligned to a multiple of HEAPS_PER_FENGINE_PER_CHUNK.

test/xbgpu/test_engine.py
```diff
@@ -967,12 +1012,28 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
         )
         # assert_allclose converts to float, which bloats memory usage.
         # To keep it manageable, compare a batch at a time.
-        for i in range(beam_results.shape[0]):
-            for j in range(beam_results.shape[1]):
-                np.testing.assert_allclose(beam_results[i, j], expected_beams[i, j], atol=1)
-
+        for i in range(expected_beams.shape[0]):
+            # NOTE: As per the explanation at the end of `_send_data`, we
+            # only verify up the the penultimate heap's data for each `beam_result`
```
bmerry, line 1017: You fixed the code below, but "penultimate" is still only coincidentally true for the particular test parameters.

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -346,42 +348,70 @@ def verify_corrprod_sensors(
...
+    weights, quant_gains, delays
+        The beam weights, quantiser-gains and delays applied to each input of
+        the beam data product. These are real floating-point values generated
+        for the unit test.
     """
-    # Generate expected sensor updates from BPipeline and compare
+    # Get the number of total heaps transmitted by each beam output
+    n_beam_heaps_sent = beam_results_shape[1]
+    heap_shape = beam_results_shape[2:]
+    # NOTE: Explicitly cast np.prod returns to int as the default is np.int64
```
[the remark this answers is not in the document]
amishatishpatel: I could've sworn I was getting complaints from someone in the `pre-commit` around the comparison I was doing in the `assert` statement(s) later on.
* Maybe that was before I had the `assert` check on each Prometheus counter to confirm they aren't None-types?

Either way, removing the cast seems to go through just fine now. Thanks for flagging it.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -174,13 +174,13 @@ def timestamp(self, value: int) -> None:
 
     @staticmethod
     def _inc_counters(
-        beam_data_shape: tuple[int, int, int], beam_dtype: np.dtype, beam_names: Sequence[str], future: asyncio.Future
+        data_shape: tuple[int, int, int], beam_dtype: np.dtype, beam_names: Sequence[str], future: asyncio.Future
```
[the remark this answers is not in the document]
amishatishpatel: Argh, too right once again. Not sure where my head was at yesterday, I definitely should've noted the `beam_` prefix throughout (just nextdoor).

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
amishatishpatel: Ah, I see it's a bit clearer later on (in the test itself) around the usage of `n_heaps`. When I remove the explicit cast there I get
> test/xbgpu/test_engine.py:948: error: Incompatible types in assignment (expression has type "signedinteger[_64Bit | Any]", variable has type "int")  [assignment]

(Also, the variables using `n_heaps` don't need to be a signed 64-bit integer, but 🤷 )

test/xbgpu/test_engine.py
amishatishpatel: Too right, although I would say that's nothing overly concerning
* We've just done that as it's a convenient (reliable/repeatable) way of forcing a test of the XPipeline (Re: that old ~resync logic comment).

I'm not sure it requires a tweak to the comment here
* It's a relatively unrelated point that shouldn't cause an issue for this check (?)
* But perhaps worth a mention, regardless

amishatishpatel:
@bmerry I've got unit test failures at that batch-indices contiguous check. I'll pass the ticket over to you once it's fixed.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -174,7 +174,7 @@ def timestamp(self, value: int) -> None:
 
     @staticmethod
     def _inc_counters(
-        data_shape: tuple[int, int, int], beam_dtype: np.dtype, beam_names: Sequence[str], future: asyncio.Future
+        data_shape: tuple[int, int, int], data_type: np.dtype, output_names: Sequence[str], future: asyncio.Future
```
bmerry, line 177: I'd rather call it `data_dtype` (or just `dtype`): "dtype" has a specific meaning in numpy that's more specific than just "type".

test/xbgpu/test_engine.py
```diff
@@ -669,13 +668,15 @@ async def _send_data(
                 corrprod_results[i][j] = ig_recv["xeng_raw"].value
 
         # TODO: NGC-1172 The tweaks to process beam data below rely on
-        # `batch_indices` to be contiguous. The check below is temporary until
-        # the BPipeline is able to handle missing data.
-        assert sorted(batch_indices) == (
+        # `batch_indices` to be contiguous and for the zeroth (and minimum)
+        # value to be a multiple of `HEAPS_PER_FENGINE_PER_CHUNK`. The check
+        # below is temporary until the BPipeline is able to handle missing
+        # data.
```
bmerry, line 674: You've documented the extra assumption (zeroth element is a multiple) but haven't asserted it.

test/xbgpu/test_engine.py
```diff
@@ -1014,9 +1015,9 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
         # To keep it manageable, compare a batch at a time.
         for i in range(expected_beams.shape[0]):
             # NOTE: As per the explanation at the end of `_send_data`, we
-            # only verify up the the penultimate heap's data for each `beam_result`
-            # because the final heap is sent coincidentally - not because it is
-            # expected to have sane data in it.
+            # only verify data in the range of `batch_indices` for each
+            # `beam_result` as any heaps sent afterwards are sent by default
+            # - not because it is expected to have sane data in it.
```
bmerry, line 1020: You've switched from plural ("heaps .. are") to singular ("it is") mid-sentence.

bmerry (changes requested):

PR #710, opened 2024-03-08
Add test that beam-weights are using the right inputs.

bmerry:
Closes NGC-1258.

- [x] If qualification tests are changed: attach a sample qualification report: [report-test.pdf](https://github.com/ska-sa/katgpucbf/files/14536590/report-test.pdf)

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_weights.py
+++ b/qualification/tied_array_channelised_voltage/test_weights.py
@@ -1,0 +1,99 @@
+################################################################################
+# Copyright (c) 2024, National Research Foundation (SARAO)
+#
+# Licensed under the BSD 3-Clause License (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+"""Weights test."""
+
+import numpy as np
+import pytest
+from pytest_check import check
+
+from .. import CBFRemoteControl, TiedArrayChannelisedVoltageReceiver
+from ..reporter import Reporter
+
+
+@pytest.mark.requirements("CBF-REQ-0123")
+async def test_weight_mapping(
+    cbf: CBFRemoteControl,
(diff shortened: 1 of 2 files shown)


qualification/tied_array_channelised_voltage/test_weights.py
```diff
@@ -0,0 +1,99 @@
...
+        assert candidate_beams, "No beam includes this input"
+        test_beam = rng.choice(candidate_beams)
+        test_beam_name = receiver.stream_names[test_beam]
+        pdf_report.detail(f"Using beam {test_beam_name} as the test beam.")
+
+        source_indices = receiver.source_indices[test_beam]
+        input_pos = source_indices.index(input_idx)
+        weights = [0.0] * len(source_indices)
+        weights[input_pos] = 1.0
+        inv_weights = [1.0 - w for w in weights]
```
amishatishpatel, line 78: Must admit, this is neat. Although, why this approach specifically? 
* Are you accommodating for testing weights other than `1.0` in the future?
* And/or, why didn't you just do it in the for-loop below with an `else` case?

amishatishpatel (approved):
Again, seems so simple now, but I don't think I would've come up with this test so easily.

One or two questions, but otherwise good to go.

qualification/tied_array_channelised_voltage/test_weights.py
bmerry: > Are you accommodating for testing weights other than 1.0 in the future?

That's what NGC-1257 is for.

> And/or, why didn't you just do it in the for-loop below with an else case?

Either way works; but I figured then the special one will get lost in amongst all the rest, whereas with this approach it's clearer when reading the report how the one beam is being treated specially.

qualification/tied_array_channelised_voltage/test_weights.py
```diff
@@ -0,0 +1,99 @@
...
+        pdf_report.detail(f"Set gain on {input_label} to 1.0, all others to 0.0.")
+        candidate_beams = [i for i, source_indices in enumerate(receiver.source_indices) if input_idx in source_indices]
+        assert candidate_beams, "No beam includes this input"
+        test_beam = rng.choice(candidate_beams)
+        test_beam_name = receiver.stream_names[test_beam]
+        pdf_report.detail(f"Using beam {test_beam_name} as the test beam.")
+
+        source_indices = receiver.source_indices[test_beam]
+        input_pos = source_indices.index(input_idx)
+        weights = [0.0] * len(source_indices)
```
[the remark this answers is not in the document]
bmerry: receiver.source_indices is a list of list of ints. I'll add an annotation to that effect, since it'll probably strengthen mypy's checks.

amishatishpatel (approved):
Please and thank you.

PR #713, opened 2024-03-12
Set auto_strategy on rx.missing-unixtime sensors

bmerry:
When data was partially missing, these were getting updated at much higher frequency than desired.

### DIFF
--- a/src/katgpucbf/fgpu/recv.py
+++ b/src/katgpucbf/fgpu/recv.py
@@ -262,6 +262,8 @@ def make_sensors(sensor_timeout: float) -> aiokatcp.SensorSet:
                 "The timestamp (in UNIX time) when missing data was last detected",
                 default=aiokatcp.core.Timestamp(-1.0),
                 initial_status=aiokatcp.Sensor.Status.NOMINAL,
+                auto_strategy=aiokatcp.SensorSampler.Strategy.EVENT_RATE,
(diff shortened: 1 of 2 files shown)


src/katgpucbf/fgpu/recv.py
```diff
@@ -262,6 +262,8 @@ def make_sensors(sensor_timeout: float) -> aiokatcp.SensorSet:
                 "The timestamp (in UNIX time) when missing data was last detected",
                 default=aiokatcp.core.Timestamp(-1.0),
                 initial_status=aiokatcp.Sensor.Status.NOMINAL,
+                auto_strategy=aiokatcp.SensorSampler.Strategy.EVENT_RATE,
+                auto_strategy_parameters=(MIN_SENSOR_UPDATE_PERIOD, math.inf),
```
james-smith-za, line 266: I'm battling to find documentation which will help me understand why the second parameter is `math.inf` - what does that mean?

bmerry: It's in the katcp spec: it controls the maximum time between transmitting updates if the value hasn't changed. math.inf means that if the value doesn't change it just won't send updates.

james-smith-za: Ah. I was looking in the aiokatcp docs, I couldn't find it there. I didn't think to look in the spec.

Thanks.

james-smith-za (approved):

PR #715, opened 2024-03-18
Bump some copyright years

bmerry:
This change was meant to go into #713 but I forgot to push it at the time.


amishatishpatel (approved):
I think you should've sent this James's way as he reviewed #713.

I know you were kind enough to point me there in the PR blurb, but he likely has more recent knowledge of why these files should get/have now gotten the bump.

PR #720, opened 2024-04-05
Set fixed number of Jones vectors per F-engine output batch

bmerry:
Previously spectra-per-heap was fixed (at 256), but this leads to problems because the heap size scales with the channel count. Small heaps don't amortise the per-heap overheads well (leading to reduced performance with 1K channels) while large heaps can cause problems with memory usage.

Replace it by fixing "jones-per-batch": the number of Jones vectors (dual-pol, complex sample) in each F-engine output batch aka spectra-per-heap * channels. The default is set to 2^20, so that 4K-channel correlators have the same integration time resolution as MeerKAT (and >4K-channel correlators have more resolution). I also took the opportunity to make `jones_per_batch` a per-output argument of fgpu, rather than a global argument (spectra-per-heap should always have been per-output, but I never got around to it).

The reduction in spectra-per-heap for 32K-channel correlation showed some performance limitations in correlate.py, because it always parallelised over batches rather than summing in registers across batches; as a result, it used far too much global memory bandwidth. The TCC kernel code has been modified so that the parallelisation granularity is independent of the batch size. This not only allows parallelism to be reduced for 32K (where the number of channels provides ample parallelism); it also allows it to be increased when channel count is low (parallel within a batch).

The kernel code change also decouples the number of intermediate accumulators from the number of batches per chunk. That allows the `--heaps-per-fengine-per-batch` default to be increased to 32, since it no longer causes excessive memory usage from the intermediate accumulators.

- [ ] Merge together with matching katsdpcontroller PR: https://github.com/ska-sa/katsdpcontroller/pull/725

Closes NGC-1121.

### DIFF
--- a/doc/control.rst
+++ b/doc/control.rst
@@ -146,7 +146,6 @@ instance on ``host2``::
     user@host1:~/katgpucbf/src/tools$ spead2_net_raw ./fsim --interface <100GbE NIC IP> --ibv \
                                       --array-size 4 --channels 4096 \
                                       --channels-per-substream 512 \
-                                      --spectra-per-heap 256 \
                                       239.10.10.10+1:7148
     .
     .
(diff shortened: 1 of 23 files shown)


amishatishpatel:
You had my curiosity during the check-ins this week, but now you have my attention with this PR.

I will admit, it's quite a meaty one. I'm likely going to review it commit-by-commit unless you have a preference/better suggestion.

bmerry:
> I'm likely going to review it commit-by-commit unless you have a preference/better suggestion.

I don't recommend that. I went down some blind alleys and had to rewrite stuff. There is an intermediate point though, which I can point out on Monday.

bmerry:
Closing in favour of #723.

PR #723, opened 2024-04-07
Set fixed number of Jones vectors per F-engine output batch

bmerry:
Previously spectra-per-heap was fixed (at 256), but this leads to problems because the heap size scales with the channel count. Small heaps don't amortise the per-heap overheads well (leading to reduced performance with 1K channels) while large heaps can cause problems with memory usage.

Replace it by fixing "jones-per-batch": the number of Jones vectors (dual-pol, complex sample) in each F-engine output batch aka spectra-per-heap * channels. The default is set to 2^20, so that 4K-channel correlators have the same integration time resolution as MeerKAT (and >4K-channel correlators have more resolution). I also took the opportunity to make `jones_per_batch` a per-output argument of fgpu, rather than a global argument (spectra-per-heap should always have been per-output, but I never got around to it).

The reduction in spectra-per-heap for 32K-channel correlation showed some performance limitations in correlate.py, because it always parallelised over batches rather than summing in registers across batches; as a result, it used far too much global memory bandwidth. The TCC kernel code has been modified so that the parallelisation granularity is independent of the batch size. This not only allows parallelism to be reduced for 32K (where the number of channels provides ample parallelism); it also allows it to be increased when channel count is low (parallel within a batch).

The kernel code change also decouples the number of intermediate accumulators from the number of batches per chunk. That allows the `--heaps-per-fengine-per-batch` default to be increased to 32, since it no longer causes excessive memory usage from the intermediate accumulators.

- [ ] Merge together with matching katsdpcontroller PR: https://github.com/ska-sa/katsdpcontroller/pull/725

Closes NGC-1121.

### DIFF
--- a/doc/control.rst
+++ b/doc/control.rst
@@ -146,7 +146,6 @@ instance on ``host2``::
     user@host1:~/katgpucbf/src/tools$ spead2_net_raw ./fsim --interface <100GbE NIC IP> --ibv \
                                       --array-size 4 --channels 4096 \
                                       --channels-per-substream 512 \
-                                      --spectra-per-heap 256 \
                                       239.10.10.10+1:7148
     .
     .
@@ -159,7 +158,7 @@ instance on ``host2``::
                                         --dst-interface <100GbE interface name or IP address> \
                                         --src-ibv --dst-ibv \
                                         --adc-sample-rate 1712e6 --array-size 4 \
-                                        --channels 4096 --spectra-per-heap 256 \
+                                        --channels 4096 \
                                         --channels-per-substream 512 \
                                         --samples-between-spectra 8192 \
                                         --katcp-port 7150 \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ known_local_folder = ["noisy_search", "remote", "sighandler"]
 [tool.pytest.ini_options]
 testpaths = "test"
 addopts = "--cov-context=test --cov-report html --import-mode=prepend"
-markers = ["mask_timestamp", "use_vkgdr", "cmdline_args"]
+markers = ["mask_timestamp", "use_vkgdr", "cmdline_args", "spectra_per_heap"]
 asyncio_mode = "auto"
 
 [tool.coverage.run]
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2020-2023, National Research Foundation (SARAO)
+# Copyright (c) 2020-2024, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -38,6 +38,7 @@ from katsdpsigproc.abc import AbstractContext
 from katsdptelstate.endpoint import Endpoint, endpoint_list_parser
 
 from .. import (
+    DEFAULT_JONES_PER_BATCH,
     DEFAULT_KATCP_HOST,
     DEFAULT_KATCP_PORT,
     DEFAULT_PACKET_PAYLOAD_BYTES,
@@ -109,6 +110,7 @@ class OutputDict(TypedDict, total=False):
 
     name: str
     channels: int
+    jones_per_batch: int
     dst: list[Endpoint]
     taps: int
     w_cutoff: float
@@ -152,7 +154,7 @@ def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str]
                         raise ValueError(f"{key} specified twice")
                     case "name":
                         kws[key] = data
-                    case "channels" | "taps":
+                    case "channels" | "taps" | "jones_per_batch":
                         kws[key] = int(data)
                     case "w_cutoff":
                         kws[key] = float(data)
@@ -185,10 +187,15 @@ def parse_wideband(value: str) -> WidebandOutput:
     try:
         kws: WidebandOutputDict = {}
         _parse_stream(value, kws, field_callback)
-        kws = {"taps": DEFAULT_TAPS, "w_cutoff": DEFAULT_W_CUTOFF, **kws}  # type: ignore
+        kws = {
+            "taps": DEFAULT_TAPS,
+            "w_cutoff": DEFAULT_W_CUTOFF,
+            "jones_per_batch": DEFAULT_JONES_PER_BATCH,
+            **kws,
+        }
+        return WidebandOutput(**kws)
     except ValueError as exc:
         raise ValueError(f"--wideband: {exc}") from exc
-    return WidebandOutput(**kws)
 
 
 def parse_narrowband(value: str) -> NarrowbandOutput:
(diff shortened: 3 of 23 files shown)


bmerry:
@amishatishpatel I edited the history to collapse away the blind alleys, so this PR should be more suitable for reviewing commit-at-a-time.

amishatishpatel:
> @amishatishpatel I edited the history to collapse away the blind alleys, so this PR should be more suitable for reviewing commit-at-a-time.

Please and thank you!

pyproject.toml
```diff
@@ -33,7 +33,7 @@ known_local_folder = ["noisy_search", "remote", "sighandler"]
 [tool.pytest.ini_options]
 testpaths = "test"
 addopts = "--cov-context=test --cov-report html --import-mode=prepend"
-markers = ["mask_timestamp", "use_vkgdr", "cmdline_args"]
+markers = ["mask_timestamp", "use_vkgdr", "cmdline_args", "spectra_per_heap"]
```
amishatishpatel, line 36: Admittedly not 100% sure on how these markers fit in with running pytest (I know I could go off and look it up).

Does this allow you to override the args from your own `pytest.ini` file?

src/katgpucbf/fgpu/main.py
```diff
@@ -226,13 +233,14 @@ def field_callback(kws: NarrowbandOutputDict, key: str, data: str) -> None:
         kws = {
             "taps": DEFAULT_TAPS,
             "w_cutoff": DEFAULT_W_CUTOFF,
+            "jones_per_batch": DEFAULT_JONES_PER_BATCH,
             "weight_pass": DEFAULT_WEIGHT_PASS,
             "ddc_taps": DEFAULT_DDC_TAPS_RATIO * kws["decimation"],
-            **kws,  # type: ignore[misc]
```
amishatishpatel, line 231: No need for the `# type: ignore` anymore?

src/katgpucbf/fgpu/output.py
```diff
@@ -29,10 +29,22 @@ class Output(ABC):
...
+    jones_per_batch: int
     taps: int
     w_cutoff: float
     dst: list[Endpoint]
 
+    def __post_init__(self) -> None:
+        if self.channels % len(self.dst) != 0:
+            raise ValueError("channels must be a multiple of the number of destinations")
+        if self.jones_per_batch % self.channels != 0:
+            raise ValueError("jones_per_batch must be a multiple of channels")
```
amishatishpatel, line 41: TIL about `__post_init__`!

pyproject.toml
amishatishpatel: Ah, I see you can `@pytest.mark` any of these args at the top of a unit test. Nice one.

doc/control.rst
amishatishpatel, line 1: I know it might have just been a `grep` or Ctrl+F through the directory, but this is a good catch.

src/katgpucbf/xbgpu/kernels/tensor_core_correlation_kernel.cu
```diff
@@ -246,12 +247,14 @@ template <typename T> struct FetchData
...
-  __device__ void load(const Samples samples, unsigned channel, unsigned time, unsigned firstReceiver, bool skipLoadCheck = NR_RECEIVERS % NR_RECEIVERS_PER_BLOCK == 0)
+  __device__ void load(const Samples *samples, unsigned channel, unsigned time, unsigned firstReceiver, bool skipLoadCheck = NR_RECEIVERS % NR_RECEIVERS_PER_BLOCK == 0)
   {
     if (skipLoadCheck || firstReceiver + loadRecv < NR_RECEIVERS)
     {
-      data = * (T *) &samples[firstReceiver + loadRecv][channel][time][loadTime][0];
-      //memcpy(&data, &samples[firstReceiver + loadRecv][channel][time][loadTime][0], sizeof(T));
+      unsigned outerTime = time / NR_BLOCKS_PER_BATCH;
+      unsigned innerTime = time % NR_BLOCKS_PER_BATCH;
+      data = * (T *) &samples[outerTime][firstReceiver + loadRecv][channel][innerTime][loadTime][0];
```
amishatishpatel, line 256: Once more, I know I could quickly hack away and/or look this up for my own clarity, but this review has been quite a journey already (and a good one) so I'm just going to ask here:
* Re: The update to require `*samples` instead of `samples` (in the arglist), what is the reasoning there?
  * I gather it's a practice in C++ that is ~better?
* Again, showing my naivety on C++, the `Samples` type doesn't seem to have explicitly _gained_ an extra dimension. Your documentation update pretty neatly says 
  * "The `FetchData` class splits the time index into a batch index and an offset within the batch [`outerTime` and `innerTime`], so that the rest of the code can be oblivious to batches, and just work in time "blocks" (`NR_TIMES_PER_BLOCK` spectra).
  * Obviously, brilliant. But how does the indexing "actually work" now? I don't expect anything like a Python IndexError; it's the addition of a new indexing dimension (`[outerTime]`) that's throwing me a bit.
  * You're still combing across the same amount of data, just in a different way.

test/xbgpu/test_correlation.py
```diff
@@ -144,9 +146,12 @@ def test_correlator(
     fill_random(rng, buf_visibilities_device, command_queue)
 
     # Calculate expected values
-    calculated_visibilities_host = np.zeros(buf_visibilities_host.shape, buf_visibilities_host.dtype)
+    calculated_visibilities_host = np.zeros(buf_visibilities_host.shape, np.int64)
```
amishatishpatel, line 149: Why the move to `np.int64`? (from `np.int32`)

test/xbgpu/test_parameters.py
```diff
@@ -21,8 +21,7 @@
 # run.
 array_size = [1, 3, 4, 19, 33, 64, 80]
 
-# This is always set to 256 for the MeerKAT case.
-num_spectra_per_heap = [256]
+num_jones_per_batch = [2**20]
```
amishatishpatel, line 24: Aha, thank you.

amishatishpatel (commented):
As always, a couple comments for clarification before explicitly approving. That was quite an adventure.

The logic around the kernel update is quite a bit to hold in my head all at once (the mental image, anyway). 
* My lack of specific comments there is not due to lack of attention, I promise. 
* I think I've just about cracked it; might need to look at the use of `accel.divup` once more at the end of today just to be sure
  * That did feel like the sting in the logical tail when reviewing this. That, and the unit test updates 😅

pyproject.toml
bmerry: Yes. Putting them here just tells pytest "yes this really is a mark that has some meaning." Otherwise it'll warn you that you might be missing a plugin that would interpret it.

src/katgpucbf/fgpu/main.py
bmerry: No, mypy fixed the bug.

src/katgpucbf/xbgpu/kernels/tensor_core_correlation_kernel.cu
bmerry: C is rather weird about arrays and particularly about passing arrays to functions. In various contexts (including parameter passing), an array "decays" to a pointer to the first element of the array. So now instead of passing a single instance of SampleData, I'm passing an array of SampleData by passing a pointer to the first element of the array.

test/xbgpu/test_correlation.py
bmerry: Something I changed (probably just passing more spectra per heap in some circumstances) was causing int32 to overflow, so I upgraded the test to correctly saturate the same way the GPU code does (by first calculating in int64 then clamping).

amishatishpatel: Noted and noted.

src/katgpucbf/xbgpu/kernels/tensor_core_correlation_kernel.cu
amishatishpatel: Noted and appreciated.

Still, a bit of a question mark around the updated indexing of `samples`.
* Are my points above valid?
* Slash, have I understood it mostly/~correctly around "the number of axes over which you index doesn't matter as long as you're within the bounds of the desired data"?

bmerry:
Amish has verbally approved.

PR #724, opened 2024-04-07
Improve beamformer performance with >32 dual-pol beams

bmerry:

### DIFF
--- a/doc/xbgpu.design.b.rst
+++ b/doc/xbgpu.design.b.rst
@@ -52,9 +52,19 @@ sizes have two advantages:
    small, and there is not enough work to keep all the work items busy, making
    the coefficient computation less efficient.
 
-The design above works well for small numbers of beams, but the register usage
-scales with the number of beams. Supporting significantly more beams would
-require rework to process beams in smaller batches.
+Higher beam counts
+^^^^^^^^^^^^^^^^^^
+The design above works well for small numbers of beams (up to about 64
+single-pol beams), but the register usage scales with the number of beams and
+eventually the registers spill to memory, causing very poor performance.
+
+To handle more beams, the kernel batches over beams, just as it does over
+antennas. The beam batch loop becomes an outer loop, with the rest of the
+kernel operating as before but only on a single batch.
+
+This does mean that the inputs are loaded multiple times, but caches help
+significantly here, and the kernel tends to be be more compute-bound in this
+domain.
 
 Dithering
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


doc/xbgpu.design.b.rst
```diff
@@ -52,9 +52,19 @@ sizes have two advantages:
...
    the coefficient computation less efficient.
 
-The design above works well for small numbers of beams, but the register usage
-scales with the number of beams. Supporting significantly more beams would
-require rework to process beams in smaller batches.
+Higher beam counts
+^^^^^^^^^^^^^^^^^^
+The design above works well for small numbers of beams (up to about 64
+single-pol beams), but the register usage scales with the number of beams and
+eventually the registers spill to memory, causing very poor performance.
```
amishatishpatel, line 59: I have to clarify, Re:
> and eventually the registers spill to memory

What/which memory are you referring to here? It's "all GPU memory", so I would think it spills over to System memory? (Admittedly not entirely familiar with the term/phrase)

src/katgpucbf/xbgpu/kernels/beamform.mako
```diff
@@ -88,64 +94,78 @@ KERNEL REQD_WORK_GROUP_SIZE(BLOCK_SPECTRA, 1, 1) void beamform(
...
+                    }
                 }
             }
+            BARRIER();  // protect against next loop iteration overwriting l_weights
         }
-        BARRIER();  // protect against next loop iteration overwriting l_weights
-    }
-
-    if (!valid)
-        return;
```
amishatishpatel, line 140: The removal of this branch of code makes me wonder what it was used for in the first place - seeing as it was checked so late in the day/kernel?

amishatishpatel (commented):
Really neat, really cool update. Just a couple questions before I can explicitly Approve.

Also, I'm still unsure whether this requires an update to `beamform.py`
* The writing to `out` over the batches still happens for all `b_batch`es unrolled,
* I don't think the way the kernel is invoked needs to change either (I'm not too familiar with mako syntax anyway to really critique)
  * I gather it will stil acknowledge your passing of `"beam_pols": self.beam_pols` as a recognised parameter - just for a different use now.

doc/xbgpu.design.b.rst
james-smith-za: Registers are the little boxes very close to the CPU which some working data is stored in.

If you have too many variables then data needs to get stored in global memory which has quite a long round-trip if you need to read it again.

amishatishpatel (approved):

PR #725, opened 2024-04-08
Misc mypy fixes

bmerry:
Mostly adding more packages to the pre-commit config and dealing with the errors that reveals.

Closes NGC-1247.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_channel_shape.py
+++ b/qualification/antenna_channelised_voltage/test_channel_shape.py
@@ -119,7 +119,7 @@ async def test_channel_shape(
         (np.arange(-0.5, 0.55, 0.1), -1.5, "Channel response (zoomed)"),
     ]:
         fig = Figure()
-        ax = fig.subplots()
+        ax = fig.add_subplot()
         ax.plot(x, db_plot)
         ax.set_title(title)
         ax.set_xlabel("Channel")
--- a/qualification/tied_array_channelised_voltage/test_delay.py
+++ b/qualification/tied_array_channelised_voltage/test_delay.py
@@ -17,9 +17,11 @@
 """Delay test."""
 
 import time
+from typing import Sequence, cast
 
 import numpy as np
 import pytest
+from matplotlib.axes import Axes
 from matplotlib.figure import Figure
 from pytest_check import check
 
@@ -186,7 +188,8 @@ async def test_delay(
 
         corr_phase = np.angle(corr)
         fig = Figure(tight_layout=True)
-        ax, ax_err = fig.subplots(2)
+        # matplotlib's typing doesn't specialise for Nx1 case
+        ax, ax_err = cast(Sequence[Axes], fig.subplots(2))
         x = range(receiver.n_chans)
         delta = wrap_angle(corr_phase - expected_phase)
         max_error_deg = np.max(np.abs(np.rad2deg(delta)))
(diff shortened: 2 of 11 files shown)


qualification/antenna_channelised_voltage/test_channel_shape.py
```diff
@@ -119,7 +119,7 @@ async def sample(offsets: ArrayLike) -> np.ndarray:
         (np.arange(-0.5, 0.55, 0.1), -1.5, "Channel response (zoomed)"),
     ]:
         fig = Figure()
-        ax = fig.subplots()
+        ax = fig.add_subplot()
```
james-smith-za, line 122: What was wrong with the way it was?

qualification/tied_array_channelised_voltage/test_delay.py
```diff
@@ -186,7 +188,8 @@ async def test_delay(
 
         corr_phase = np.angle(corr)
         fig = Figure(tight_layout=True)
-        ax, ax_err = fig.subplots(2)
+        # matplotlib's typing doesn't specialise for Nx1 case
+        ax, ax_err = cast(Sequence[Axes], fig.subplots(2))
```
james-smith-za, line 192: If this is a common pattern it may be worth filing a bug / feature request for this particular issue?

I can forsee this code pattern getting boring pretty quickly.

scratch/benchmarks/sighandler.py
```diff
@@ -36,4 +36,6 @@ def handler(task: asyncio.Task) -> None:
         task.cancel()
 
     loop = asyncio.get_running_loop()
-    loop.add_signal_handler(signal.SIGINT, handler, asyncio.current_task())
+    task = asyncio.current_task()
+    assert task is not None
+    loop.add_signal_handler(signal.SIGINT, handler, task)
```
james-smith-za, line 41: Is this just to keep mypy happy or did it actually reveal a subtle bug?
(This is just a benchmark script I realise so it's not something we'd encounter often anyway but out of curiosity.)

james-smith-za (approved):
Looks good to me, a couple of requests for clarification mostly for my own understanding.

qualification/antenna_channelised_voltage/test_channel_shape.py
bmerry: `subplots` can return a range of different things, and that was causing mypy to complain because some of them aren't usable the way `ax` is used. `add_subplot` has a more restricted return type.

qualification/tied_array_channelised_voltage/test_delay.py
bmerry: You mean like https://github.com/matplotlib/matplotlib/issues/28045 ?

scratch/benchmarks/sighandler.py
bmerry: To keep mypy happy, although it could be None if this function was used incorrectly.

qualification/tied_array_channelised_voltage/test_delay.py
james-smith-za: Already on top of things. I like it.

qualification/antenna_channelised_voltage/test_channel_shape.py
james-smith-za: Right, makes sense. Matplotlib gives me the general feeling of being from an older time before type checking was even a thing, so trying to fit it on after the fact feels weird.

PR #727, opened 2024-04-10
Cache HeapReferenceList in bsend

bmerry:
This avoids having to recompute the list of heaps on every transmission. Instead, it is recomputed (on each frame) whenever it changes, which is much rarer.

This makes a noticeable (~20%) improvement to CPU usage when there is a large number (50+) beams, although at present the beamformer kernel is the bottleneck to supporting more beams.

- [x] (N/A)If design has changed: ensure documentation is up to date

Relates to NGC-1232.

### DIFF
--- a/src/katgpucbf/xbgpu/bsend.py
+++ b/src/katgpucbf/xbgpu/bsend.py
@@ -59,7 +59,9 @@ SEND_DTYPE = np.dtype(np.int8)
 class Frame:
     """Hold all data for heaps with a single timestamp.
 
-    It does not own its memory - the backing store is in :class:`Chunk`.
+    It does not own its memory - the backing store is in :class:`Chunk`. It keeps
+    a cached :class:`spead2.send.HeapReferenceList` with the heaps of the enabled
+    beams, along with a version counter that is used to invalidate it.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


james-smith-za:
Just to clarify what the 20% means.

So is it going e.g. from 80% to 60%? Or by 20% of whatever it was?

bmerry:
Goes from about 60% to 40%, IIRC.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -238,15 +242,17 @@ def send(
...
-                    if enabled
-                ]
+                if frame.tx_enabled_version != send_stream.tx_enabled_version:
+                    frame.tx_heaps = spead2.send.HeapReferenceList(
+                        [
+                            spead2.send.HeapReference(heap, substream_index=i, rate=rate)
+                            for i, (heap, enabled) in enumerate(zip(frame.heaps, send_stream.tx_enabled))
+                            if enabled
+                        ]
+                    )
```
amishatishpatel, line 252: Ah, I see what you mean when you say
> building this list every time may be too expensive.

BSend itself honours the `enable_substream` request and ticks over when it's time to update any substreams. The `Frame`s then follow.

Very cool.

PR #728, opened 2024-04-11
Port fsim to Python

bmerry:
There are a few differences from the C++ versions

- The interface is given by name instead of IP address.
- I removed the --max-heaps argument, because I've never used it, and
  used a constant (QUEUE_DEPTH) instead to keep the CLI simpler. It
  could easily be reintroduced against if needed, but YAGNI.
- The default --array-size is increased to 80.
- It has an explicit --samples-between-spectra for simulating
  narrowband.
- It has --affinity and --main-affinity to control thread affinity.
- It sends descriptors.
- It exposes Prometheus metrics.
- It has unit tests.

Partially closes NGC-1087.


src/katgpucbf/fsim/__init__.py
james-smith-za, line 1: Vaguely amusing that git thinks this file is moved rather than two unrelated files.

IMO the entire `tools` directory can probably be deleted, at the moment there's just an unnecessary `.gitignore` there.

src/katgpucbf/fsim/main.py
```diff
@@ -0,0 +1,330 @@
...
+from . import METRIC_NAMESPACE
+
+DTYPE = np.dtype(np.int8)
+#: Number of heaps in time to keep in flight
+QUEUE_DEPTH = 8
+output_heaps_counter = Counter("output_heaps", "number of heaps transmitted", namespace=METRIC_NAMESPACE)
+output_bytes_counter = Counter("output_bytes", "number of payload bytes transmitted", namespace=METRIC_NAMESPACE)
+time_error_gauge = Gauge(
+    "time_error_s", "elapsed time minus expected elapsed time", ["stream"], namespace=METRIC_NAMESPACE
+)
```
james-smith-za, line 61: Do these things work if you run the fsim not via the master controller?
I was under the impression that Prometheus only knows how to find things that the MC has launched.

Edit: Looking through the arguments list now clearly I had misunderstood things previously. Prometheus gets data from correlators that the MC has launched because the MC passes `--prometheus-port` which I guess you could do from a scratch script.

Though how does Prometheus know to look there?

src/katgpucbf/fsim/main.py
```diff
@@ -0,0 +1,330 @@
...
+    """Parse the command-line arguments."""
+    parser = argparse.ArgumentParser(prog="fsim")
+
+    parser.add_argument(
+        "--adc-sample-rate", type=float, default=1712e6, help="Digitiser sampling rate (Hz) [%(default)s]"
+    )
+    parser.add_argument("--interface", default="lo", help="Network interface on which to send packets [%(default)s]")
+    parser.add_argument("--array-size", type=int, default=80, help="Number of antennas in the array")
+    parser.add_argument(
+        "--channels", type=int, default=32768, help="Total number of channels in the array [%(default)s]"
```
james-smith-za, line 74: bike-shedding:

```suggestion
        "--channels", type=int, default=32768, help="Total number of channels in the (simulated) array [%(default)s]"
```

james-smith-za (changes requested):

Dockerfile
```diff
@@ -144,19 +130,12 @@ RUN pip install --no-deps --root=/install-root .
...
+FROM build-base as build-c
 
-# Build utilities.
-# We use make clean to ensure that an existing build from the build context
-# won't accidentally get used instead.
 WORKDIR /tmp/tools
-COPY src/tools .
-RUN make clean && make -j fsim
-
 RUN wget https://raw.githubusercontent.com/ska-sa/katsdpdockerbase/master/docker-base-runtime/schedrr.c && \
```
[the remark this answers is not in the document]
bmerry: It runs a child process with real-time scheduling requested, provided that schedrr itself has the required capabilities. So a bit like `spead2_net_raw` but for real-time scheduling instead of raw network device access.

`chrt` would do the same thing, but it seems to launch the target as a child process and keeps running rather than using `exec` to replace itself, which causes problems when you want to send signals and the like.

src/katgpucbf/fsim/main.py
bmerry: Done (but without parentheses) and also applied to `--array-size` for consistency.

doc/control.rst
```diff
@@ -143,23 +143,23 @@ single :ref:`feng-packet-sim` on ``host1`` and a single :program:`xbgpu`
 instance on ``host2``::
 
     [Connect to host1, navigate to the location of the fsim and build it using 'make']
```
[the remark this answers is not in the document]
bmerry: Thanks for catching that - fixed.

src/katgpucbf/fsim/__init__.py
bmerry: Thanks, thought I'd done that but missed `.gitignore`. Done.

src/katgpucbf/fsim/main.py
bmerry: > Though how does Prometheus know to look there?

Do you mean how does it know for things launched by katsdpcontroller, or how would it know for fsim? For the former, katsdpcontroller registers all the tasks with consul, and Prometheus queries consul to discover them. For the latter, I expect it wouldn't (unless we add a stream type to katsdpcontroller to let us launch fsims through that interface). However, Prometheus uses a plain-text http protocol (by default anyway) so even if we don't connect it to the Prometheus server it can be handy to use `curl` to just fetch metrics.

james-smith-za (approved):

PR #731, opened 2024-04-15
Reduce memory usage for beamformer

bmerry:
For some reason I previously made an array of random states per BTxItem, but there is no reason these need to be scoped that way since they're never transferred on or off the GPU. Change Beamform to allocate the random states internally, which also logically simplifies the structure since the caller doesn't need to worry about building the random states.

Closes NGC-1303.

### DIFF
--- a/src/katgpucbf/xbgpu/beamform.py
+++ b/src/katgpucbf/xbgpu/beamform.py
@@ -62,9 +62,14 @@ class BeamformTemplate:
         n_ants: int,
         n_channels: int,
         n_spectra_per_frame: int,
+        seed: int,
+        sequence_first: int,
+        sequence_step: int = 1,
     ) -> "Beamform":
         """Generate a :class:`Beamform` object based on the template."""
-        return Beamform(self, command_queue, n_frames, n_ants, n_channels, n_spectra_per_frame)
+        return Beamform(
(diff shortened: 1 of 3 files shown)


src/katgpucbf/xbgpu/beamform.py
```diff
@@ -105,6 +111,8 @@ class Beamform(accel.Operation):
         Number of frequency channels
     n_spectra_per_frame
         Number of samples in time axis for each frame (fine time dimension)
+    seed, sequence_first, sequence_step
+        See :class:`.RandomStateBuilder`.
```
amishatishpatel, line 115: Don't mean to be that guy
* This doesn't render as an actual hyperlink to `src/katgpucbf/curand_helpers.py::RandomStateBuilders`.

Admittedly, I don't have a suggestion to fix this
* It would be nice if this was updated to link to the correct location, but
* I wouldn't hold this PR up over that.

amishatishpatel (approved):
One small nit-pick; otherwise, good to go.

src/katgpucbf/xbgpu/beamform.py
bmerry: You should totally be that guy. I'm that guy all the time, and it improves quality.

In this case it's because I failed to do the checklist properly when I added the curand_helpers.py module, so it's not linked into the documentation. The good news is I already fixed that in another PR in the meantime, and when I merge main into this branch, it fixes the rendering.

PR #735, opened 2024-04-18
Update Jenkins qualification test config for KDRA

bmerry:
See NGC-1043.


bmerry:
> Though I find myself wondering whether we should make this a parameter that the server itself could fill in.

It's possible, but probably not worth it at this stage. Probably the way to do it would be to have separate pytest.ini files for the different sites, and then use an environment variable to select between them. But I don't want to be debugging that at the same time as trying to debug the initial on-site qualification testing.

PR #738, opened 2024-04-19
Update BPipeline to handle missing data

amishatishpatel:
What initially started as an update for the BPipeline crept a bit into a feature tweak
on the XPipeline too. This was all to accommodate more robust data testing/handling
in the end-to-end unit test; specifically to test non-contiguous data (e.g. missed Chunks
in the receiver).

The unit test update did become quite an adventure as mypy eventually wasn't happy
with how complex it had become - hence, some abstractions. I did try to explain the 
unit test updates as much, and as often, as possible. As always, let me know if it's just
more confusing/needs tweaking.

Just realised I have tweaked an existing design of the XPipeline. Should I lump that under
the "update XBEngine documentation" ticket?

Closes NGC-1172.


bmerry:
@amishatishpatel we'll also need to update https://docs.google.com/document/d/14irzt0m5SOtKDWFzZG6xYbwhWuLQK52QIHFDaCqrCk0/edit. Do you want to do that under NGC-1172 or a new ticket?

amishatishpatel:
> @amishatishpatel we'll also need to update https://docs.google.com/document/d/14irzt0m5SOtKDWFzZG6xYbwhWuLQK52QIHFDaCqrCk0/edit. Do you want to do that under NGC-1172 or a new ticket?

Argh, too right. Admittedly the first I'm seeing/hearing of such an officially unofficial document. Still
* It should definitely be part of NGC-1172 (so I/we don't forget). 
* Maybe I'll create a child task.

bmerry (changes requested):
Some changes you can start with on the implementation. I'll look at the test updates next.

test/xbgpu/test_engine.py
```diff
@@ -947,87 +1006,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
-                    n_spectra_per_heap,
-                    missing_antenna,
-                )
-                base_batch_index += num_batches_in_current_accumulation
-                np.testing.assert_equal(expected_output, corrprod_result)
-            incomplete_accums_counters.append(incomplete_accums_counter)
+            last_timestamp = batch_end_index2 * timestamp_step
+
+        # The first warnings in the list should be from each of the XPipelines,
+        # the rest are warnings related to timestamps between output heaps.
```
bmerry, line 1021: Can you be sure that they come in this order?

If not, I think caplog has facilities to ignore particular warnings for the duration of the test.

test/xbgpu/test_engine.py
```diff
@@ -947,87 +1006,75 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
-        # To keep it manageable, compare a batch at a time.
-        for i in range(expected_beams.shape[0]):
-            # NOTE: As per the explanation at the end of `_send_data`, we
-            # only verify data in the range of `batch_indices` for each
-            # `beam_result` as any heaps sent afterwards are sent by default
-            # - not because they are expected to have sane data in them.
-            for j in range(expected_beams.shape[1]):
-                np.testing.assert_allclose(expected_beams[i, j], beam_results[i, j], atol=1)
+        for batch_start_index, batch_end_index in batch_index_tuples:
+            expected_beams = generate_expected_beams(
```
bmerry, line 1053: I would suggest modifying generate_expected_beams to take a list of batch indices to generate, instead of a range, so that you don't need to do this nested looping. It's an `njit` function and those don't cope that well with Python lists, so you should convert the list to an ndarray when passing it in.

bmerry (changes requested):
Here are some initial review comments on the tests. I'm still trying to get my head around all the test changes. One concern I have is that the choice of batch indices to send still revolves around the accumulation interval, and doesn't consider heaps_per_fengine_per_chunk. Ideally I'd like to ensure that we have full, partial and skipped chunks of input to ensure that those cases are all handled. Maybe that's already the case though? I haven't worked through the numbers.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -149,23 +150,37 @@ class BTxQueueItem(QueueItem):
 
     The `buffer_device`, `weights` and `delays` must have the same shape and
     type as the ``in``, ``weights`` and ``delays`` slots in
-    :class:`.Beamform`.
+    :class:`.Beamform`. The TxQueueitem between the gpu-proc and sender loops
+    needs to carry a record of which antennas have missed data in the current
+    :class:`recv.Chunk` of data. This is used to determine whether any beam
```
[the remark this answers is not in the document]
amishatishpatel: Admittedly in a prison of my own making. Would you be ok if I updated this to refer to the Chunk in a general sense, e.g.
- "[...] of which antennas have missed data in the **current received Chunk** being processed."
  - Similar to the docstring for `RxQueueItem`

There's probably a reason why I can't find references to files/classes from the `src/katgpucbf/` level...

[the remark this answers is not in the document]
bmerry: Sphinx lets you easily reference things from the current module/class, but it doesn't do relative paths in general. If you want a link, fully qualify it e.g. ``:class:`~katgpucbf.fgpu.recv.Chunk` `` (the leading `~` makes it render just as `Chunk`).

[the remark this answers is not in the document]
amishatishpatel: Yeah I figured I could fully spell it out for Sphinx - for some reason that felt a bit dirty to me. Still, can't argue if it works; will update accordingly.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -489,6 +505,8 @@ async def gpu_proc_loop(self) -> None:  # noqa: D102
             )
             self._beamform()
 
+            tx_item.present[:] = rx_item.present.copy()
```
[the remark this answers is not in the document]
amishatishpatel: This must be a relic of me debugging why the `beam_ants` SPEAD item wasn't updating during dev/testing. I threw all combination of methods at it to ensure I wasn't unintentionally linking arrays with incorrect assignment.

Noted and noted Re:
> That's going to make a temporary copy, which you then copy again into `tx_item.present`.

test/xbgpu/test_bsend.py
```diff
@@ -119,6 +119,11 @@ async def _send_data(
             # Populate the buffer with dummy data.
             chunk.data[:] = data[i, ...]
 
+            # Fill chunk.present_ants counts otherwise no heaps will be sent
+            # Any value > 0 will do, we aren't too concerned about handling
+            # missing data in this test.
```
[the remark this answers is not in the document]
amishatishpatel: I appreciate your diplomacy 😂

test/xbgpu/test_engine.py
```diff
@@ -269,17 +262,62 @@ def valid_end_to_end_combination(combo: dict) -> bool:
...
+    corrprod_results
+        List of arrays of all GPU-generated data from
+        :class:`TestEngine._send_data`.
+    batch_index_tuples
+        A list of batch start and end indices used to generate stimulus
+        data for the XBEngine. See :class:`TestEngine.test_engine_end_to_end`
+        for more details.
+    """
+    for i, corrprod_output in enumerate(corrprod_outputs):
+        # Or assert if incomplete_accs_total == incomplete_accums_counter * len(xbengine._pipelines)
```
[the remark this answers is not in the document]
amishatishpatel: Yeah genuinely unsure, if I'm honest. Going back through the git blame, can't really find a corresponding check that used to exist before/above it.

I'll remove it now.

test/xbgpu/test_engine.py
```diff
@@ -658,7 +713,6 @@ async def _send_data(
                     f"{hex(accumulation_index * timestamp_step * corrprod_output.heap_accumulation_threshold)}, "
                     f"actual: {hex(ig_recv['timestamp'].value)}."
                 )
-
```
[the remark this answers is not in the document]
amishatishpatel: My apologies, most certainly leftover from reshuffling things.

test/xbgpu/test_engine.py
```diff
@@ -606,6 +654,13 @@ async def _send_data(
             heaps = heap_factory(batch_index)
             await feng_stream.async_send_heaps(heaps, spead2.send.GroupMode.ROUND_ROBIN)
 
+        # NOTE: The first accumulations are expected to be dropped entirely as
+        # they are incomplete. Remove their corresponding accumulation indices
+        # from the list as these indices are used when verifying data output by
+        # the XPipelines - i.e. data for corresponding timestamps will not exist.
+        for i, corrprod_output in enumerate(corrprod_outputs):
+            acc_indices[i].remove(batch_indices[0] // corrprod_output.heap_accumulation_threshold)
+
```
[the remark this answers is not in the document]
amishatishpatel: This looks a lot like an iteration of the test I had a while back - tracking heap counts per accumulation.

Appreciate the suggestion, will definitely update it to this sort of MO.

test/xbgpu/test_engine.py
amishatishpatel: Did some digging this morning and found this ~recent PR on pytest that adds the ability to filter within the unit test
- https://github.com/pytest-dev/pytest/pull/11625/files

I see it's part of a newer version of pytest, which we might not necessarily want to upgrade to yet.

I'll use a plain old `filter` for now, maybe create a technical-debt task for updating version and usage.

bmerry: I'd say try updating pytest to the latest and check what happens, but be sure to also test the qualification tests. I vaguely recall that pytest-asyncio 0.23 has some updates to make it compatible with pytest 8, and we're stuck on pytest-asyncio 0.21 until they fix an issue with how scoped fixtures work that breaks the qualification tests.

amishatishpatel: I kicked off a qualification run (on the site Jenkins) around 13:15 today with katgpucbf/upgrade-pytest.

Will likely have to revisit this tomorrow morning. I've done a normal `filter` for now.

amishatishpatel: The qualification run seems to have passed with no complaints/failures. I'll send you the build URL via Mattermost.

Guess I have a bit more work to do before I can re-request a review...

test/xbgpu/test_engine.py
```diff
@@ -448,6 +496,13 @@ def prom_get(name: str) -> float:
         ]
 
 
+class AccumWarningFilter(Filter):
+    def filter(self, record: LogRecord) -> bool:
+        if record.message == "All Antennas had a break in data during this accumulation":
+            return True
+        return False
```
bmerry, line 503:
```suggestion
        return record.message == "All Antennas had a break in data during this accumulation":
```

test/xbgpu/test_engine.py
```diff
@@ -346,18 +379,33 @@ def verify_corrprod_sensors(
         # Verify sensor updates while we're here
         xsync_sensor_name = f"{xpipeline.output.name}.rx.synchronised"
         # As per the explanation in :func:`~send_data`, the first accumulation
-        # is expected to be incomplete.
-        expected_xsensor_updates.append((False, aiokatcp.Sensor.Status.ERROR))
+        # is expected to be incomplete. We also accommodate for any other
```
bmerry, line 382:
```suggestion
        # is expected to be incomplete. We also accommodate any other
```

test/xbgpu/test_engine.py
```diff
@@ -959,72 +1021,58 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
...
+                    n_spectra_per_heap=n_spectra_per_heap,
+                    missing_antennas=missing_antennas,
                 )
-                base_batch_index += num_batches_in_current_accumulation
-                np.testing.assert_equal(expected_output, corrprod_result)
-            incomplete_accums_counters.append(incomplete_accums_counter)
+            last_timestamp = batch_end_index2 * timestamp_step
+
+        # TODO: NGC-1308 Update this check to not be a subset of the warnings filtered
+        assert caplog.record_tuples[: len(corrprod_outputs)] == [
```
bmerry, line 1042: It looks like you've put in the filtering now, and could remove the `: len(corrprod_outputs)`?

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -219,32 +217,36 @@ def generate_expected_beams(
...
-    accum = np.zeros((len(beam_pols), num_batches, channels), np.complex64)
+    accum = np.zeros((len(beam_pols), len(batch_indices), channels), np.complex64)
     sample = np.empty((N_POLS, COMPLEX), np.int8)
     sample_fp = np.empty(N_POLS, np.complex64)
-    for batch in range(num_batches):
+    for batch_index in batch_indices:
+        # np.where returns a tuple with matches in row and column indices.
+        # `batch_index` is a flat list, so we only expect matches for row
+        # indices - and a single match, at that.
+        batch_id = np.where(batch_indices == batch_index)[0][0]
```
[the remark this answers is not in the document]
amishatishpatel: Too right. I wasn't sure what is/isn't supported by `@njit`, so defaulted to a (slightly) more convoluted approach.
* That is, from my reading it seems `numpy`-related operations are more likely supported than native Python operations (?)

test/xbgpu/test_engine.py
```diff
@@ -1047,7 +1050,9 @@ def heap_factory(batch_index: int) -> list[spead2.send.HeapReference]:
         verify_corrprod_data(
             corrprod_outputs=corrprod_outputs,
             corrprod_results=corrprod_results,
-            batch_index_tuples=batch_index_tuples,
+            # Use the batch indices that start *after* the first
+            # purposefully-empty accumulation
+            batch_indices=test_batch_indices[HEAPS_PER_FENGINE_PER_CHUNK:],
```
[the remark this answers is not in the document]
amishatishpatel: Glad you suggested this. I had something vastly similar in a previous iteration of the test (maybe even the previous `HEAD`).

I think I was getting a bit iffy with the amount of things being returned and passed around (also because I'm on two self-imposed strikes with `mypy`'s "test_engine_end_to_end is too complex").

Still, will update accordingly now.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
bmerry: Oh, hadn't spotted that it was an `@njit`. But I guess try it and see if it works. If it doesn't work, I'd suggest just having a counter you increment each time around the loop.

test/xbgpu/test_engine.py
amishatishpatel: Unfortunately not. Some combinations of `heap_accumulation_thresholds` - `(3,7), (5, 9)` - still report the "All Antennas [...]" warning during that jump in data in the middle of the test (the artificial break in data we introduced).
* Those warnings may technically be true, but not exactly because of our decision to start 'in the middle of a Chunk' at the start of the test.

Let me know if that does/doesn't make sense.

test/xbgpu/test_engine.py
```diff
@@ -292,28 +292,24 @@ def verify_corrprod_data(
     corrprod_results
         List of arrays of all GPU-generated data from
         :class:`TestEngine._send_data`.
-    batch_indices
-        Batch indices used to generate stimulus data for the XBEngine. See
-        :class:`TestEngine.test_engine_end_to_end` for more details.
+    acc_indices
+        Accumulation indices used to generate stimulus data for the XBEngine.
```
amishatishpatel, line 296: Hmmmm, maybe this description requires an update. Let me know if you're of the same opinion.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
amishatishpatel: Seems to work fine. I'll chalk this up to another happy accident!

src/katgpucbf/xbgpu/engine.py
```diff
@@ -157,9 +157,12 @@ class needs to carry a record of which antennas have missed data in the
 
     Parameters
     ----------
-    buffer_device
+    out
         An int8 type :class:`~katsdpsigproc.accel.DeviceArray` with shape
         (frames, ants, channels, spectra_per_frame, N_POLS, COMPLEX).
+    saturated
+        An uint32 type :class:`~aktsdpsigproc.accel.DeviceArray` with shape
```
bmerry, line 164:
```suggestion
        An uint32 type :class:`~katsdpsigproc.accel.DeviceArray` with shape
```

test/xbgpu/test_engine.py
bmerry: I agree - my first reaction on seeing the type annotation was "why a list of lists"? Maybe add something like "This is a list of lists, with the outer index matching `corrprod_outputs`."

test/xbgpu/test_engine.py
```diff
@@ -666,6 +656,8 @@ async def _send_data(
         beam_results
             Beamformer output, with shape (n_beams, n_frames,
             n_channels_per_substream, n_spectra_per_heap, COMPLEX).
+        accumulation_indices
```
bmerry, line 659:
```suggestion
        acc_indices
```
for consistency with the code.

bmerry (changes requested):
Looking good I think. Before merging I want to check what's going on with the qualification test code I'm developing, which seems to be getting garbage in the SPEAD item. It's probably on my side though (since you are verifying the item in test_bsend).

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Looks like a failed phishing attempt atm. Thanks for spotting it.

amishatishpatel:
> Looking good I think. Before merging I want to check what's going on with the qualification test code I'm developing, which seems to be getting garbage in the SPEAD item. It's probably on my side though (since you are verifying the item in test_bsend).

Fair enough. I've pushed changes to address your last round of comments. I'll wait on your go-ahead Re: Merging.

test/xbgpu/test_engine.py
```diff
@@ -293,8 +293,10 @@ def verify_corrprod_data(
         List of arrays of all GPU-generated data from
         :class:`TestEngine._send_data`.
     acc_indices
-        Accumulation indices used to generate stimulus data for the XBEngine.
-        See :class:`TestEngine.test_engine_end_to_end` for more details.
+        Accumulation indices used to generate stimulus data for each
+        corrprod_output. This is a list of lists, with the outer index matching
+        `corrprod_outputs`. See :class:`TestEngine.test_engine_end_to_end` for
```
bmerry, line 298:
```suggestion
        `corrprod_outputs`. See :meth:`TestEngine.test_engine_end_to_end` for
```

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -282,17 +273,57 @@ def valid_end_to_end_combination(combo: dict) -> bool:
...
+) -> None:
+    """Verify XPipeline data.
+
+    Parameters
+    ----------
+    corrprod_outputs, n_ants, n_channels_per_substream, n_spectra_per_heap
+        Unit test fixtures in :class:`TestEngine`.
+    corrprod_results
+        List of arrays of all GPU-generated data from
+        :class:`TestEngine._send_data`.
```
bmerry, line 294:
```suggestion
        :meth:`TestEngine._send_data`.
```

test/xbgpu/test_engine.py
amishatishpatel: How strange, I committed an updated `test_engine.py` in the previous commit, but this specific change didn't make it through (?). Committing now.

bmerry (approved):

PR #744, opened 2024-04-25
Add qual test for beam-weights/beam-quant-gains linearity

bmerry:
Closes NGC-1256 and NGC-1257.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/15109094/report.pdf)

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_linearity.py
+++ b/qualification/tied_array_channelised_voltage/test_linearity.py
@@ -1,0 +1,137 @@
+################################################################################
+# Copyright (c) 2024, National Research Foundation (SARAO)
+#
+# Licensed under the BSD 3-Clause License (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+"""Test linearity of gains and weights."""
+
+from typing import Awaitable, Callable
+
+import aiokatcp
+import numpy as np
+import pytest
+from matplotlib.figure import Figure
+
+from .. import CBFRemoteControl, TiedArrayChannelisedVoltageReceiver
+from ..reporter import Reporter
+
+
+async def test_linearity(
+    cbf: CBFRemoteControl,
+    receiver: TiedArrayChannelisedVoltageReceiver,
+    pdf_report: Reporter,
+    fixed: str,
+    variable: str,
+    set_fixed: Callable[[aiokatcp.Client, str, float], Awaitable],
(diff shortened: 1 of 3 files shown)


bmerry:
I'm going to close this, because NGC-1256 is going to involve very similar code and I want to rework things to avoid duplication.

bmerry:
@amishatishpatel I've reopened with the more general version that also addresses NGC-1256.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -0,0 +1,137 @@
...
+
+    pdf_report.step("Measure total power responses")
+    scales = np.logspace(-2.0, 2.0, 41)  # Note: n is chosen to ensure 1.0 is included
+    middle = np.searchsorted(scales, 1.0)
+    assert scales[middle] == pytest.approx(1.0)
+    powers = np.zeros_like(scales)
+    for i, scale in enumerate(scales):
+        await set_variable(client, beam_name, scale)
+        _, data = await receiver.next_complete_chunk()
+        powers[i] = np.sum(np.square(data[0].astype(np.float64)))
```
amishatishpatel, line 59: (Not entirely important)

It did take me a few seconds to realise/remember why you only grab the first element in `data`. Looked around because you definitely made a note of setting the dsim period to one heap. If I could really be a pain, could you either
* Add a reminder comment here that we do only send/expect one heap of data to be processed, or
* Add the same/similar comment above `period = ...` 

Again, not a hill I'm willing to die on, but maybe for future readers. Please and thank you!

amishatishpatel (commented):
Man, so neat. Just a couple requests for one or two more comments.

Otherwise, good to go.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -0,0 +1,137 @@
...
+    set_fixed: Callable[[aiokatcp.Client, str, float], Awaitable],
+    set_variable: Callable[[aiokatcp.Client, str, float], Awaitable],
+) -> None:
+    client = cbf.product_controller_client
+    beam_name = receiver.stream_names[0]
+    n_sources = len(receiver.source_indices[0])
+
+    # Small amplitude so that we don't saturate in the time domain.
+    period = receiver.n_spectra_per_heap * receiver.n_samples_between_spectra
+    await cbf.dsim_gaussian(16.0, pdf_report, period=period)
```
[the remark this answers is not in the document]
bmerry: Good point. I like the latter.

qualification/tied_array_channelised_voltage/test_linearity.py
bmerry: Actually the index `[0]` is to take the first beam, since we don't need multiple beams for the test. I'll try to make that clearer.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -27,7 +27,7 @@
 from ..reporter import Reporter
 
 
-async def test_linearity(
+async def _test_linearity(
```
amishatishpatel, line 30: Aha, I was thinking that when I did the initial review, just forgot to comment/ask. Thanks for changing it.

amishatishpatel (approved):

PR #746, opened 2024-04-25
Add a narrowband frequency resolution test

bmerry:
It's done by analysis. The actual resolution is so much smaller than the channel width that I think it would be tough to measure, although possibly one could measure the long-term phase drift of a tone.

The requirement text is taken from the original MK requirements spec. MK actually ships with much coarser resolution, so we should be fine when the requirements are finalised.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/15110115/report.pdf)

Closes NGC-638.


james-smith-za (approved):
Significantly better than the 26 kHz that seems to have been the informal requirement.

PR #748, opened 2024-04-29
Qual: skip beam chunks with missing data

bmerry:
Check the beam_ants SPEAD item to determine whether any input heaps went missing.

Closes NGC-1172.

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -49,7 +49,7 @@ from spead2.recv.numba import chunk_place_data
 
 import katgpucbf.recv
 from katgpucbf import COMPLEX, DIG_SAMPLE_BITS
-from katgpucbf.spead import DEFAULT_PORT, FREQUENCY_ID, TIMESTAMP_ID
+from katgpucbf.spead import BEAM_ANTS_ID, DEFAULT_PORT, FREQUENCY_ID, TIMESTAMP_ID
 from katgpucbf.utils import TimeConverter
 
 from .reporter import Reporter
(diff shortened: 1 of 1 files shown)


qualification/__init__.py
```diff
@@ -576,21 +581,27 @@ def create_tied_array_channelised_voltage_receive_stream(
...
+        payload_size = items[3]
         beam = user_data[0]
         # If the payload size doesn't match, discard the heap (could be descriptors etc).
         if payload_size == expected_payload_size and timestamp >= 0:
             data[0].chunk_id = timestamp // timestamp_step
             data[0].heap_index = channel_offset // n_chans_per_substream + beam * n_substreams
             data[0].heap_offset = data[0].heap_index * payload_size
+            extra[0] = items[2]  # beam_ants
+            data[0].extra_size = beam_ants_itemsize
+            data[0].extra_offset = data[0].heap_index * beam_ants_itemsize
```
amishatishpatel, line 604: I see this follows the format in `src/.../fgpu/recv.py`, which is great. However, entirely unrelated, regarding "accommodating for `extra` item(s)"
* Would you be open to e.g. me writing some kind of "How to write a `chunk_place` callback"?
* Or is there already some note on this in your recent tutorial update?

qualification/__init__.py
```diff
@@ -604,9 +615,11 @@ def chunk_place(data_ptr, data_size, user_data_ptr):
         use_ibv,
         stream_config,
         max_chunks,
+        beam_ants_dtype.itemsize,
         chunk_place.ctypes,
         lambda stream: katgpucbf.recv.Chunk(
             present=np.empty((n_beams, n_substreams), np.uint8),
+            extra=np.zeros((n_beams, n_substreams), beam_ants_dtype),
```
amishatishpatel, line 622: Admittedly haven't been in the mindset of qualification (and katsdpcontroller) for some time; not entirely reconciling the idea that this Receiver is receiving data **output by the XBEngine**. More specifically,
* Why is it that the shape is `(n_beams, n_substreams)`?
* e.g. In the BSend, each beam data product is referred to as a substream
* I'm likely missing something in the change in naming convention/environment from `katgpucbf/src/` to here.

amishatishpatel (approved):
As usual, the odd comment for clarification. We can chat separately from this PR if you'd prefer.

qualification/__init__.py
bmerry: It's documented at https://spead2.readthedocs.io/en/latest/recv-chunk.html#extra-data. I don't think it's in the tutorials yet, but it might go into a later round once the current set have been reviewed.

qualification/__init__.py
bmerry: There is typically a corner turn involved: each XBEngine sends part of the band for every beam, while a typical receiver will consume the entire band for a single beam. So the sender's substreams are not the same as the receiver's substreams. In this case we're consuming the whole band for multiple beams, so I guess more properly one should call this `n_substreams_per_beam` or something, but that's quite a mouthful.

PR #750, opened 2024-04-30
Add test that beamformer gets channels right

bmerry:
Closes NGC-1272.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/ska-sa/katgpucbf/files/15160723/report.pdf)

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_channels.py
+++ b/qualification/tied_array_channelised_voltage/test_channels.py
@@ -1,0 +1,77 @@
+################################################################################
+# Copyright (c) 2024, National Research Foundation (SARAO)
+#
+# Licensed under the BSD 3-Clause License (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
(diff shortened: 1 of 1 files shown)


qualification/tied_array_channelised_voltage/test_channels.py
```diff
@@ -0,0 +1,77 @@
...
+    receiver = receive_tied_array_channelised_voltage
+    client = cbf.product_controller_client
+
+    pdf_report.step("Select channels")
+    rng = random.Random(1)
+    channels = sorted(rng.sample(range(receiver.n_chans), 30))
+    pdf_report.detail(f"Using channels {channels}")
+
+    pdf_report.step("Configure dsim")
+    signal = " + ".join(f"cw(0.5, {c / receiver.n_chans * receiver.bandwidth})" for c in channels)
```
james-smith-za, line 50: You don't think 0.5 is a bit strong? When 30 CW tones are in phase it'll be very loud.

james-smith-za (approved):
Overall I am happy, just one question-mark.

qualification/tied_array_channelised_voltage/test_channels.py
bmerry: I hadn't thought of that. Let me check that the test still passes if I reduce that.

bmerry: I reduced it to 0.05. That could still theoretically saturate if 20 tones were all in phase, but that seems like an unlikely eventuality, and even more unlikely to spoil the test overall.

james-smith-za: Yes, time-domain clipping is likely to manifest itself all over the band.

Ideally we'd do a more rigorous analysis of what this test should check but since it's an informal one, I think this is fine. 0.05 sounds in line with what I'd expect.

james-smith-za (approved):

PR #752, opened 2024-04-30
Add 'comb' function to dsim

bmerry:
This will be useful for testing impulse response through the PFB.

Relates to NGC-624.

### DIFF
--- a/src/katgpucbf/dsim/signal.py
+++ b/src/katgpucbf/dsim/signal.py
@@ -255,8 +255,7 @@ class CW(Signal):
         return f"cw({self.amplitude}, {self.frequency})"
 
 
-# mypy override is due to https://github.com/python/mypy/issues/5374
-@dataclass  # type: ignore
+@dataclass
 class Random(Signal):
     """Base class for randomly-generated signals.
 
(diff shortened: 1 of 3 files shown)


src/katgpucbf/dsim/signal.py
```diff
@@ -198,28 +200,73 @@ def __str__(self) -> str:
...
+        ----------
+        offset
+            Index of the first element of the chunk (1-element array)
+        amplitude
+            Amplitude of the signal
+        chunk_size
+            Number of samples to produce
+        frequency
+            Rounded frequency, expressed as cycles per sample
         """
```
james-smith-za, line 229: The fact that this is passing pre-commit suggests that this question is pointless, but I was under the impression that an `@abstractmethod` needed at the very least a `...` or `pass`. Or does the docstring serve that purpose?

james-smith-za (approved):

src/katgpucbf/dsim/signal.py
bmerry: Yep, a docstring is enough to keep the parser happy - it counts as a "statement"

PR #753, opened 2024-05-03
Extend timeout of qualification tests to 7 hours

james-smith-za:
See NGC-1318.

I also shifted the scheduled run two hours earlier (i.e. 01:00 SAST rather than the 03:00 that it previously was) to accommodate this.

I still want to get to the bottom of why it's suddenly not finishing inside the timeout, but if an extra two hours doesn't work then it's probably quite egregious.

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 5, unit: 'HOURS')
+    timeout(time: 7, unit: 'HOURS')
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


amishatishpatel:
Is there any way we can trial this before committing it?

james-smith-za:
I guess we could.

amishatishpatel:
Also, both Jenkins branch and pr-merge builds failed. Not sure if that's worth looking into just yet, but let me know what you find.

amishatishpatel:
Closing this as it was the update was implemented more holistically in #763.

PR #756, opened 2024-05-07
Make report generator work with skipped tests

bmerry:
It seems that the `wasxfail` attribute that is written with the reason for an xfail test, does not even exist for skipped tests.

We don't expect any skipped tests in normal operation, but I threw in a pytest.skip just while debugging and it broke the report generation.


amishatishpatel (approved):
It's not that I don't believe that this fixes it. Rather, if a test is decorated with `@pytest.mark.xfail`, the `report.outcome` is `skipped`?
- i.e. the outcome "skipped" applies to both tests marked with `xfail` **and** `skip`
- But we want to confirm the special type of "skipped" outcome, in this case by checking for the `wasxfail` attribute

I know you say
> We don't expect any skipped tests in normal operation

Which still begs the question
* What made you uncover this?

Either way, great spot.

PR #759, opened 2024-05-08
Update numba to the latest version

bmerry:
The version we were using doesn't work on Python 3.12.


james-smith-za (approved):
Looks fine to me.

Are we planning a port to 3.12 imminently? Or is this just to make things smoother when we finally do?

bmerry:
My new laptop comes with 3.12, so I thought I'd give it a try and see what shakes out.

PR #765, opened 2024-05-20
Naming convention cleanup

amishatishpatel:
As discussed on the ticket, this PR addresses consolidation of the following terms
* adc_sample_rate vs adc_sample_rate_hz
* n_channels vs n_channels_total
  * Keep n_channels as the total number of frequency channels for this array configuration across fgpu and xbgpu
  * Update the kernels/submodules in xbgpu to n_channels_per_substream where n_channels is used
* InItem / OutItem vs RxItem / TxItem
  * Move to In / Out prefix within both engines to indicate GPU-related work
  * fgpu already has a send_free_queue utilising Chunks from send.py, so we can’t go with Recv / Send prefixes
  * Also fill out the name completely as InQueueItem and OutQueueItem where applicable
* Batch vs Frame, in both fgpu and xbgpu
  * Simply update the usage of the word/term frame to be batch
* Usage of num_x and n_x for variable names

- [ ] If ICD-defined sensors have been added: update ~~`fake_servers.py`~~ `generator.py` (no mention of affected sensor in `fake_servers.py`) in katsdpcontroller to match - see https://github.com/ska-sa/katsdpcontroller/pull/736

Closes NGC-1158 and NGC-1269.

### DIFF
--- a/doc/fgpu.design.rst
+++ b/doc/fgpu.design.rst
@@ -534,7 +534,7 @@ From there a number of transformations occur:
    computed (per polarisation) to determine whether any are missing, to
    produce per-spectrum presence flags.
 4. When an output chunk is ready to be sent, the per-spectrum flags are
-   reduced to per-frame flags.
+   reduced to per-batch flags.
 
 .. _fgpu.narrow:
 
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -372,6 +372,26 @@ class XBReceiver:
                     return chunks
         raise RuntimeError(f"stream was shut down before we received {n} complete chunk(s)")
 
+    @overload
+    def channel_frequency(self, channel: float) -> float:
+        ...
+
+    @overload
+    def channel_frequency(self, channel: np.ndarray) -> np.ndarray:
+        ...
+
+    def channel_frequency(self, channel):
+        """Compute the frequency (in Hz) for a given channel.
+
+        The channel number may be a real value to select a frequency that is
+        not at the centre of a channel bin. Integral values correspond to bin
+        centres. This is the frequency of the signal received from the
+        digitiser rather than sky frequency.
+
+        Either a scalar or a numpy array may be used.
+        """
+        return (channel - self.n_chans / 2) * self.bandwidth / self.n_chans + self.center_freq
+
     def compute_tone_gain(self, amplitude: float, target_voltage: float) -> float:
         """Compute F-Engine gain.
 
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -233,7 +233,7 @@ class XBReceiver:
         self.sync_time = cbf.sensors[f"{acv_name}.sync-time"].value
         self.scale_factor_timestamp = cbf.sensors[f"{acv_name}.scale-factor-timestamp"].value
         self.bandwidth = cbf.sensors[f"{acv_name}.bandwidth"].value
-        self.center_freq = cbf.sensors[f"{acv_name}.center-freq"].value
+        self.centre_freq = cbf.sensors[f"{acv_name}.centre-freq"].value
         self.multicast_endpoints = [
             [
                 (endpoint.host, endpoint.port)
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -233,7 +233,7 @@ class XBReceiver:
         self.sync_time = cbf.sensors[f"{acv_name}.sync-time"].value
         self.scale_factor_timestamp = cbf.sensors[f"{acv_name}.scale-factor-timestamp"].value
         self.bandwidth = cbf.sensors[f"{acv_name}.bandwidth"].value
-        self.centre_freq = cbf.sensors[f"{acv_name}.centre-freq"].value
+        self.center_freq = cbf.sensors[f"{acv_name}.center-freq"].value
         self.multicast_endpoints = [
             [
                 (endpoint.host, endpoint.port)
@@ -390,7 +390,7 @@ class XBReceiver:
 
         Either a scalar or a numpy array may be used.
         """
-        return (channel - self.n_chans / 2) * self.bandwidth / self.n_chans + self.centre_freq
+        return (channel - self.n_chans / 2) * self.bandwidth / self.n_chans + self.center_freq
 
     def compute_tone_gain(self, amplitude: float, target_voltage: float) -> float:
         """Compute F-Engine gain.
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -390,7 +390,7 @@ class XBReceiver:
 
         Either a scalar or a numpy array may be used.
         """
-        return (channel - self.n_chans / 2) * self.bandwidth / self.n_chans + self.center_freq
+        return (channel - self.n_chans / 2) * self.bandwidth / self.n_chans + self.centre_freq
 
     def compute_tone_gain(self, amplitude: float, target_voltage: float) -> float:
         """Compute F-Engine gain.
--- a/qualification/antenna_channelised_voltage/test_linearity.py
+++ b/qualification/antenna_channelised_voltage/test_linearity.py
@@ -42,11 +42,11 @@ async def test_linearity(
     cw_scales = [0.5**i for i in range(10)]
     pdf_report.detail(f"CW scales: {cw_scales}")
 
-    pdf_report.step("Select a channel and compute the channel center frequency for the D-sim.")
-    sel_chan_center = receiver.n_chans // 3
-    channel_frequency = sel_chan_center * (receiver.bandwidth / receiver.n_chans)
+    pdf_report.step("Select a channel and compute the channel centre frequency for the D-sim.")
+    sel_chan_centre = receiver.n_chans // 3
+    channel_frequency = sel_chan_centre * (receiver.bandwidth / receiver.n_chans)
     pdf_report.detail(
-        f"Channel {sel_chan_center} selected, with center frequency " + f"{channel_frequency/1e6:.2f} MHz."
+        f"Channel {sel_chan_centre} selected, with centre frequency " + f"{channel_frequency/1e6:.2f} MHz."
     )
 
     pdf_report.step("Set EQ gain.")
@@ -56,12 +56,12 @@ async def test_linearity(
     await cbf.product_controller_client.request("gain-all", "antenna-channelised-voltage", gain)
 
     base_corr_prod = await sample_tone_response(
-        rel_freqs=sel_chan_center,
+        rel_freqs=sel_chan_centre,
         amplitude=cw_scales,
         receiver=receiver,
     )
 
-    linear_scale_result = base_corr_prod[:, sel_chan_center]
+    linear_scale_result = base_corr_prod[:, sel_chan_centre]
 
     # Normalise and compute the effective received voltage value (from power) for comparison to the requested value.
     linear_test_result = np.sqrt(linear_scale_result / np.max(linear_scale_result))
--- a/qualification/antenna_channelised_voltage/test_linearity.py
+++ b/qualification/antenna_channelised_voltage/test_linearity.py
@@ -43,10 +43,10 @@ async def test_linearity(
     pdf_report.detail(f"CW scales: {cw_scales}")
 
     pdf_report.step("Select a channel and compute the channel centre frequency for the D-sim.")
-    sel_chan_centre = receiver.n_chans // 3
-    channel_frequency = sel_chan_centre * (receiver.bandwidth / receiver.n_chans)
+    sel_channel_centre = receiver.n_chans // 3
+    channel_frequency = receiver.channel_frequency(sel_channel_centre)
     pdf_report.detail(
-        f"Channel {sel_chan_centre} selected, with centre frequency " + f"{channel_frequency/1e6:.2f} MHz."
+        f"Channel {sel_channel_centre} selected, with centre frequency " + f"{channel_frequency/1e6:.2f} MHz."
     )
 
     pdf_report.step("Set EQ gain.")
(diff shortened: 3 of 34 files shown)


qualification/__init__.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, cbf: CBFRemoteControl, stream_names: Sequence[str]) -> None:
         self.sync_time = cbf.sensors[f"{acv_name}.sync-time"].value
         self.scale_factor_timestamp = cbf.sensors[f"{acv_name}.scale-factor-timestamp"].value
         self.bandwidth = cbf.sensors[f"{acv_name}.bandwidth"].value
-        self.center_freq = cbf.sensors[f"{acv_name}.center-freq"].value
```
ludwigschwardt, line 236: Are we really changing CBF sensor names?

james-smith-za: If you prefer, Ludwig, we can add the change to katdal and you can just approve it.
We just had an inconsistency with how it was used and there's no way to make it consistent without breaking *something*.

ludwigschwardt: Nah, it's fine, I'll do that part.

It's funny. In my world, consistency means CBF says "center" and the rest of MeerKAT says "centre". I'm used to that by now. :-P

bmerry: I'm against updating these sensor names. katdal is just one of many consumers of sensors, and forcing them all to cope with two possible spellings makes the system less maintainable rather than more.

james-smith-za: The product-configure schema uses "centre" so either some inconsistency or some level of breakage is unavoidable at this point.

doc/fgpu.design.rst
james-smith-za, line 1: I approve of this choice.

I forget where the convention to refer to it as a "frame" originated (possibly in SDP somewhere?) but in my mind this can introduce some ambiguity with an Ethernet frame so batch seems better.

qualification/__init__.py
ludwigschwardt: That was kind of my point... We've been living with SPEAD / sensor inconsistency for so long that it's turned into consistency for me. Make sure you understand the breakage before you put consistency first though. From my side it is OK to add a workaround to katdal but I'm sympathetic to Bruce's concerns.

doc/fgpu.design.rst
bmerry: I started using the term in [ingest](https://github.com/ska-sa/katsdpingest/blob/3e365a47920f523124e616bc013b8c5f8f35de85/katsdpingest/receiver.py#L34) at least 8 years ago. I'm not married to it though.

qualification/antenna_channelised_voltage/test_linearity.py
```diff
@@ -42,11 +42,11 @@ async def test_linearity(
     cw_scales = [0.5**i for i in range(10)]
     pdf_report.detail(f"CW scales: {cw_scales}")
 
-    pdf_report.step("Select a channel and compute the channel center frequency for the D-sim.")
-    sel_chan_center = receiver.n_chans // 3
-    channel_frequency = sel_chan_center * (receiver.bandwidth / receiver.n_chans)
+    pdf_report.step("Select a channel and compute the channel centre frequency for the D-sim.")
+    sel_chan_centre = receiver.n_chans // 3
+    channel_frequency = sel_chan_centre * (receiver.bandwidth / receiver.n_chans)
```
bmerry, line 47: Since we're aiming for consistency, I'd suggest renaming to `sel_channel_centXX` so that we're locally consistent about "channel" versus "chan".

src/katgpucbf/fgpu/engine.py
```diff
@@ -290,7 +290,7 @@ class OutItem(QueueItem):
     spectra_samples
         Number of ADC samples between spectra.
     timestamp
-        Timestamp of the first spectrum in the `OutItem`.
+        Timestamp of the first spectrum in the `OutQueueItem`.
```
bmerry, line 293:
```suggestion
        Timestamp of the first spectrum in the :class:`OutQueueItem`.
```

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -344,9 +344,9 @@ class BSend:
     ----------
     outputs
         Sequence of :class:`.output.BOutput`.
-    frames_per_chunk
-        Number of :class:`Frame`\ s in each transmitted :class:`Chunk`.
-    n_tx_items
+    batches_per_chunk
+        Number of :class:`Batch`\ es in each transmitted :class:`Chunk`.
+    n_out_items
```
bmerry, line 349: I think this should maybe get called `n_chunks`, since this is not the number of BOutQueueItem instances.

src/katgpucbf/xbgpu/bsend.py
```diff
@@ -344,9 +344,9 @@ class BSend:
...
         Number of :class:`Chunk` to create.
```
bmerry, line 350:
```suggestion
        Number of :class:`Chunk`\ s to create.
```

src/katgpucbf/xbgpu/correlation.py
```diff
@@ -130,7 +130,7 @@ def __init__(
                 f"-DNR_RECEIVERS={self.n_ants}",
                 f"-DNR_RECEIVERS_PER_BLOCK={self._n_ants_per_block}",
                 f"-DNR_BITS={self.input_sample_bits}",
-                f"-DNR_CHANNELS={self.n_channels}",
+                f"-DNR_CHANNELS_PER_SUBSTREAM={self.n_channels_per_substream}",
```
bmerry, line 133: I'd prefer not to be messing with the variable names used by TCC - it makes it harder to incorporate updates. Several of the other variables names are also different to what we use, but that's reasonable at interface boundaries with 3rd-party code.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -18,7 +18,7 @@
 This module defines the objects that implement an entire GPU XB-Engine pipeline.
 
 The XBEngine comprises multiple Pipeline objects that facilitate output data
-products. Additionally, the RxQueueItem and TxQueueItem objects, used in the
+products. Additionally, the InQueueItem and OutQueueItem objects, used in the
 XBEngine for passing information between different async processing loops,
 are defined here.
```
bmerry, line 23: There is no OutQueueItem class, and it's the *classes* that are defined here. Additionally, "async processing loops" sounds like event loops, which is confusing.

```suggestion
products. Additionally, several subclasses of :class:`.QueueItem` are defined for passing
data between parts of the engine.
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -100,11 +100,11 @@ def reset(self, timestamp: int = 0) -> None:
         self.chunk: recv.Chunk | None = None
 
 
-class XTxQueueItem(QueueItem):
+class XOutQueueItem(QueueItem):
     """
     Extension of the QueueItem to track antennas that have missed data.
 
-    The TxQueueItem between the gpu-proc and sender loops needs to carry a
+    The OutQueueItem between the gpu-proc and sender loops needs to carry a
```
bmerry, line 107:
```suggestion
    The XOutQueueItem between the gpu-proc and sender loops needs to carry a
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -256,55 +256,51 @@ def __init__(self, outputs: Sequence[_O], name: str, engine: "XBEngine", context
...
+        # - The _in_queue receives items from :meth:`.XBEngine._receiver_loop`
         #   to be used by :meth:`_gpu_proc_loop`.
-        # - The _tx_item_queue receives items from :meth:`gpu_proc_loop` to be
+        # - The _out_queue receives items from :meth:`gpu_proc_loop` to be
         #   used by :meth:`sender_loop`.
         # Once the destination function is finished with an item, it will pass
         # it back to the corresponding free-item queue to ensure that all
         # allocated buffers are in continuous circulation.
-        # NOTE: Pipelines must not place :class:`RxQueueItem`s directly back on
+        # NOTE: Pipelines must not place :class:`InQueueItem`s directly back on
```
bmerry, line 266:
```suggestion
        # NOTE: Pipelines must not place :class:`InQueueItem`\ s directly back on
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -256,55 +256,51 @@ def __init__(self, outputs: Sequence[_O], name: str, engine: "XBEngine", context
...
         # the `_rx_free_item_queue` as multiple pipelines will hold
```
bmerry, line 267: `_rx_free_item_queue` still needs to be renamed to `_in_free_queue`.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -969,14 +965,14 @@ class XBEngine(DeviceServer):
...
     katcp_port
         Network port on which to listen for KATCP C&M connections.
-    adc_sample_rate_hz
+    adc_sample_rate
         Sample rate of the digitisers in the current array. This value is
         required to calculate the packet spacing of the output heaps. If it is
         set incorrectly, the packet spacing could be too large causing the
         pipeline to stall as heaps queue at the sender faster than they are
         sent.
     bandwidth_hz
```
bmerry, line 974: Let's take out the `_hz` here too (and anywhere else it turns up in variable names).

src/katgpucbf/xbgpu/kernels/tensor_core_correlation_kernel.cu
bmerry, line 1: Again, would like to leave this file alone as far as possible.

test/xbgpu/test_bsend.py
```diff
@@ -35,9 +35,9 @@
 
 from . import test_parameters
 
-FRAMES_PER_CHUNK: Final[int] = 5
-N_TX_ITEMS: Final[int] = 2
-TX_HEAPS_PER_SUBSTREAM: Final[int] = N_TX_ITEMS * FRAMES_PER_CHUNK
+BATCHES_PER_CHUNK: Final[int] = 5
+N_OUT_ITEMS: Final[int] = 2
```
bmerry, line 39: I'd suggest calling this N_CHUNKS, since it's controlling allocation of Chunks not BOutQueueItems.

bmerry (changes requested):
Generally happy - mostly it's the external interfaces where I think we need to pull back a bit on renaming things.

I would still like to see other 'tx' and 'rx' variables (e.g. tx_enabled) renamed to match the 'send' and 'recv' terms that are already used in module names (and in spead2).

bmerry:
@james-smith-za I'm having ambivalent feeling about whether the correlation.py and beamform.py modules need to change from n_channels to n_channels_per_substream. On the one hand, they're self-contained pieces of code that don't have or need any knowledge of channels other than the ones they're actually processing, so introducing the concept of "substreams" into them feels unnatural. On the other hand, if they take parameters called "n_channels", it would be quite easy to write incorrect code that passes engine.n_channels into them.

src/katgpucbf/xbgpu/correlation.py
james-smith-za: This is a good point that I hadn't thought about.

james-smith-za:
> @james-smith-za I'm having ambivalent feeling about whether the correlation.py and beamform.py modules need to change from n_channels to n_channels_per_substream. On the one hand, they're self-contained pieces of code that don't have or need any knowledge of channels other than the ones they're actually processing, so introducing the concept of "substreams" into them feels unnatural. On the other hand, if they take parameters called "n_channels", it would be quite easy to write incorrect code that passes engine.n_channels into them.

Unless we introduce some kind of a new term for the subset of the bandwidth that is being correlated / beamformed.

I forget - does the XB engine know that it's number N of M?

bmerry:
> I forget - does the XB engine know that it's number N of M?

Indirectly: N = channel_offset_value // channels_per_substream and M = channels // channels_per_substream.

qualification/__init__.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, cbf: CBFRemoteControl, stream_names: Sequence[str]) -> None:
         self.sync_time = cbf.sensors[f"{acv_name}.sync-time"].value
         self.scale_factor_timestamp = cbf.sensors[f"{acv_name}.scale-factor-timestamp"].value
         self.bandwidth = cbf.sensors[f"{acv_name}.bandwidth"].value
-        self.centre_freq = cbf.sensors[f"{acv_name}.centre-freq"].value
+        self.center_freq = cbf.sensors[f"{acv_name}.center-freq"].value
```
james-smith-za: The thought I had in mind was that everything internally would use "centre", except for the actual sensor which would be "center".

It's not a hill I'm willing to die on though.

qualification/antenna_channelised_voltage/test_linearity.py
```diff
@@ -43,10 +43,10 @@ async def test_linearity(
     pdf_report.detail(f"CW scales: {cw_scales}")
 
     pdf_report.step("Select a channel and compute the channel centre frequency for the D-sim.")
-    sel_chan_centre = receiver.n_chans // 3
-    channel_frequency = sel_chan_centre * (receiver.bandwidth / receiver.n_chans)
+    sel_channel_centre = receiver.n_chans // 3
```
james-smith-za, line 46: Particularly since we're using "centre" here.

james-smith-za (approved):
One piece of bike-shedding, change or leave unchanged at your discretion. I don't have a strong opinion either way.

No need to re-review if you do decide to make a change.

qualification/__init__.py
bmerry: These variables are purely mirrors of the sensor values, so I thought it was best if they used exactly the same names as the sensors (modulo dash to underscore conversion).

james-smith-za: Fair point.

PR #770, opened 2024-05-23
Make delay tests fail fast if data is missing

bmerry:
Since it's using complete_chunks() directly instead of next_complete_chunk(), it doesn't have any timeout, and if complete chunks aren't arriving it can hang the test forever.

Also fixes a bug where if non-contiguous chunks are seen, it could hang things by not returning the chunk to the stream. It's possible that this is the real reason for NGC-1341.

Closes NGC-1341.

- [x] If qualification tests are changed: attach a sample qualification report: here's a report where I deliberately decreased the timeout to induce failures: 

Sample qualification test reports: I haven't included the success case, because it looks like the same as before. Induced failures:

- test_group_delay: induce retries by decreasing the timeout: [report.pdf](https://github.com/ska-sa/katgpucbf/files/15416365/report.pdf)
- both: induce failure by issuing a capture-stop: [report2.pdf](https://github.com/ska-sa/katgpucbf/files/15417692/report2.pdf)

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -257,6 +257,7 @@ class XBReceiver:
         *,
         all_timestamps: Literal[False] = False,
         max_delay: int = DEFAULT_MAX_DELAY,
+        time_limit: float | None = None,
     ) -> AsyncGenerator[tuple[int, katgpucbf.recv.Chunk], None]:  # noqa: D102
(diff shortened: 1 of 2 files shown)


james-smith-za (approved):
Elegant.

PR #772, opened 2024-05-24
Set dig-rms-dbfs sensors to FAILURE on missing data

james-smith-za:
If all the data is present, update the sensor with the average power calculated by the GPU.

If any is missing, set it to FAILURE.

Closes NGC-855.


james-smith-za:
I initially tried adding test functionality for this but I couldn't quite work out a good way to do it.

Not on a Friday afternoon anyway. Maybe it's not that important to test?

src/katgpucbf/fgpu/engine.py
```diff
@@ -895,7 +895,7 @@ async def _chunk_send_and_cleanup(
                 chunk.cleanup()
                 chunk.cleanup = None  # Potentially helps break reference cycles
 
-    async def run_transmit(self) -> None:
+    async def run_transmit(self) -> None:  # noqa: C901
```
bmerry, line 898: I think this is a sign that we should lift the whole `if dig_total_power is not None: ...` block into a helper function.

src/katgpucbf/fgpu/engine.py
```diff
@@ -948,18 +948,26 @@ async def run_transmit(self) -> None:
...
-                    # Normalise relative to full scale. The factor of 2 is because we
-                    # want 1.0 to correspond to a sine wave rather than a square wave.
-                    avg_power /= ((1 << (self.engine.src_layout.sample_bits - 1)) - 1) ** 2 / 2
-                    # If for some reason there's zero power, avoid reporting
-                    # -inf dB by assigning the most negative representable value
-                    avg_power_db = 10 * math.log10(avg_power) if avg_power else np.finfo(np.float64).min
-                    self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
-                        avg_power_db, timestamp=self.engine.time_converter.adc_to_unix(out_item.end_timestamp)
-                    )
+                if np.all(out_item.present):
```
bmerry, line 951: You're making this decision on the whole item, which means that if a packet goes missing on one pol, you're setting both pols to failure. It looks like `out_item.present` has a pol axis, so you could make this more fine-grained.

src/katgpucbf/fgpu/engine.py
```diff
@@ -948,18 +948,26 @@ async def run_transmit(self) -> None:
...
+                        # If for some reason there's zero power, avoid reporting
+                        # -inf dB by assigning the most negative representable value
+                        avg_power_db = 10 * math.log10(avg_power) if avg_power else np.finfo(np.float64).min
+                        self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
+                            avg_power_db, timestamp=self.engine.time_converter.adc_to_unix(out_item.end_timestamp)
+                        )
+                else:
+                    for pol in range(N_POLS):
+                        self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
+                            -np.inf,
```
bmerry, line 967:
```suggestion
                            np.finfo(np.float64).min,
```
We ensured we avoid -inf in the success path because CAM has problems with non-finite sensors (I think due to JSON encoding values).

We could also just use 0.0, since the value is supposed to be ignored when the status is FAILURE.

bmerry (changes requested):
Re testing: at a minimum we want to be sure that the code isn't going to throw an exception and break the whole engine.

It might be easier to add something to the existing test_missing_heaps test rather than trying to extend test_dig_rms_dbfs_sensors. If you attach an Observer to the sensor you can get a record of all the sensor updates (see the xbgpu end-to-end test for an example), and verify that at some point the sensor went to FAILURE.

james-smith-za:
> It might be easier to add something to the existing test_missing_heaps test rather than trying to extend test_dig_rms_dbfs_sensors. If you attach an Observer to the sensor you can get a record of all the sensor updates (see the xbgpu end-to-end test for an example), and verify that at some point the sensor went to FAILURE.

I did initially do that (if you look back at previous commits) but the problem that I had was that sensor updates didn't line up with heaps, there were a lot more of the latter than the former.

If I just printed stuff on the traceback then I could definitely see it working but I couldn't figure out how confirm whether it was going to failure at the right time.

src/katgpucbf/fgpu/engine.py
james-smith-za: I thought of making it zero but if you're looking in a dbfs range, then zero implies all the power rather than none of it. I'm not sure what CAM's displays do, if they know to omit a failure, I'd rather make the graph go down than up.

I like your idea though, thanks for the suggestion.

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+                        continue
+
+                    # If a long time has gone by that means that more than a chunk
+                    # was missing on the input and the chunk was just not
+                    # transmitted, therefore the sensor would not be updated. The
+                    # 1.1 is arbitrary, it just needs to be bigger than 1.
+                    if mark < i - 1.1 * timestamps_per_output_chunk:
+                        # Once the input recovers, the output starts a new chunk
+                        # and the sensor won't consider the entire missing range,
+                        # just the most recent complete (or nearly complete chunk).
```
bmerry, line 952:
```suggestion
                        # just the most recent complete (or nearly complete) chunk.
```

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+                        mark = i - timestamps_per_output_chunk + 1
+
+                    # TODO: make the 32 spectra-per-heap into an easier-to-use
+                    # constant. It's a naked 32 in the pytest mark at the top
+                    # of this test.
+                    recent_present = dst_present[mark // 32 : i // 32]
```
bmerry, line 958: It's already made available as a local variable in the function.
```suggestion
                    recent_present = dst_present[mark // spectra_per_heap : i // spectra_per_heap]
```

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
                 # the time difference is a multiple of the mixer wavelength.
                 np.testing.assert_equal(x, y)
 
+        # Information needed to check that the dig-rms-dbfs sensor is correctly
+        # going into FAILURE when there are missing samples.
+        time_converter = TimeConverter(SYNC_TIME, ADC_SAMPLE_RATE)
+        unix_timestamps = [time_converter.adc_to_unix(t) for t in timestamps]
+        # The sensor is updated once per out_item.
+        timestamps_per_output_chunk = engine_server.chunk_jones // output.channels
```
bmerry, line 923: "timestamps" is not a helpful name here - there are many different objects (ADC samples, spectra, heaps) that have associated timestamps and there are different numbers of them per output chunk. If I'm reading the code right, this is `spectra_per_output_chunk`?

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+                # what the last timestamp was that we considered.
+                mark = 0
+                for i, ts in enumerate(unix_timestamps[: len(unix_timestamps) // 2]):
+                    # Advance until we have caught up with the sensor's timestamp.
+                    if ts < sensor_update.timestamp:
+                        continue
+
+                    # If a long time has gone by that means that more than a chunk
+                    # was missing on the input and the chunk was just not
+                    # transmitted, therefore the sensor would not be updated. The
```
bmerry, line 947:
```suggestion
                    # transmitted, and therefore the sensor would not be updated. The
```

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+
+            # Check dig-rms-dbfs sensor. Not present in the narrowband mode.
+            if decimation == 1:
+                sensor_update_iter = iter(sensor_update_dict[f"input{pol}.dig-rms-dbfs"])
+                sensor_update = next(sensor_update_iter)
+                # The sensor is updated once per out_item. But the step between
+                # timestamps of the sensor update is not consistent, it wobbles a
+                # bit (I am currently not sure why). So we need to keep track of
+                # what the last timestamp was that we considered.
+                mark = 0
```
bmerry, line 939: What units is `mark` in? I think it's spectra?

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+                        # and the sensor won't consider the entire missing range,
+                        # just the most recent complete (or nearly complete chunk).
+                        mark = i - timestamps_per_output_chunk + 1
+
+                    # TODO: make the 32 spectra-per-heap into an easier-to-use
+                    # constant. It's a naked 32 in the pytest mark at the top
+                    # of this test.
+                    recent_present = dst_present[mark // 32 : i // 32]
+                    if np.all(recent_present):
+                        assert sensor_update.status != aiokatcp.Sensor.Status.FAILURE
```
bmerry, line 960: Can you check that it more tightly (e.g. I think it should always be NOMINAL or WARN)?

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
             input_missing_heaps = np.sum(~src_present[pol])
             assert prom_diff.get_sample_diff("input_missing_heaps_total", {"pol": str(pol)}) == input_missing_heaps
+
+            # Check dig-rms-dbfs sensor. Not present in the narrowband mode.
+            if decimation == 1:
+                sensor_update_iter = iter(sensor_update_dict[f"input{pol}.dig-rms-dbfs"])
+                sensor_update = next(sensor_update_iter)
+                # The sensor is updated once per out_item. But the step between
+                # timestamps of the sensor update is not consistent, it wobbles a
+                # bit (I am currently not sure why). So we need to keep track of
```
bmerry, line 937: Can you elaborate? Is this just the limitations of floating-point rounding (which limits UNIX timestamps to about 200ns precision these days) or something bigger?

test/fgpu/test_engine.py
```diff
@@ -913,9 +915,60 @@ async def test_missing_heaps(
...
+
+                    # If a long time has gone by that means that more than a chunk
+                    # was missing on the input and the chunk was just not
+                    # transmitted, therefore the sensor would not be updated. The
+                    # 1.1 is arbitrary, it just needs to be bigger than 1.
+                    if mark < i - 1.1 * timestamps_per_output_chunk:
+                        # Once the input recovers, the output starts a new chunk
+                        # and the sensor won't consider the entire missing range,
+                        # just the most recent complete (or nearly complete chunk).
+                        mark = i - timestamps_per_output_chunk + 1
```
bmerry, line 953: I think you're going to have to explain the + 1 to me: why are you excluding a single spectrum?

bmerry (changes requested):
I feel like the testing is made more complicated that necessary because it's working backwards: it takes each sensor update, then tries to figure out which input went into it (which needs a lot of fuzzy logic because the timestamp is a float that's lost some precision). It also means that test failures will probably be hard to debug. I think a cleaner solution would be to build a list of expected sensor updates:

1. For each output chunk:
1.1. If there was no data sent for it, no update is expected.
1.2. Otherwise, if there was some missing data, the status is expected to be FAILURE (and the timestamp can be computed from the ADC timestamp of the chunk).
1.3. Otherwise, the status is expected to be NOMINAL or WARN.

Then after you've built a list of expected sensor updates you can assert that it equals the actual list.

src/katgpucbf/fgpu/engine.py
```diff
@@ -895,7 +895,34 @@ async def _chunk_send_and_cleanup(
                 chunk.cleanup()
                 chunk.cleanup = None  # Potentially helps break reference cycles
 
-    async def run_transmit(self) -> None:  # noqa: C901
+    def _update_dig_power_sensors(self, dig_total_power: accel.HostArray | None, out_item: OutQueueItem) -> None:
+        """Update digitiser power sensors.
+
+        Helper function for keeping the complexity of :meth:`run_transmit` down to manageable levels.
+        """
+        present = out_item.present.reshape((N_POLS, -1), order="F")  # pols are interleaved
```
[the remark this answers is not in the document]
james-smith-za: I did poke around at this and it is doing what I thought it was doing but it's still wrong.

It's a one-dimensional array:
```
out_item.present.shape=(2048,)
out_item.present=array([ True,  True,  True, ...,  True,  True,  True])
```
but what I had forgotten is that each `True` is a heap, and not a subset of a heap. The pols are interleaved within the heap.

So as it turns out, we can't distinguish pols by the `out_item.present`. Unless I change the architecture of this feature, I'd have to just make both pols fail when any data is missing.

test/fgpu/test_engine.py
james-smith-za: You're correct, that's probably a better way to refer to it.

test/fgpu/test_engine.py
james-smith-za: There are two hard things in computer science.

test/fgpu/test_engine.py
james-smith-za: I am not quite sure whether it's just floating-point rounding or something else.

In the first test which I used for debugging, with a `spectra_per_chunk = 2048` I saw mostly diffs of 2047, 2048, a handful of 2050. And then a really large one (34848) where there was the large gap.

I couldn't see any kind of pattern.

james-smith-za:
> Then after you've built a list of expected sensor updates you can assert that it equals the actual list.

The problem that I've been scratching my head with all day today is that the F-engine re-syncs (for want of a better term) when it's missed more than an entire chunk of data. So for the unit test, I can't just do 
```python
chunk_present = np.reshape(dst_present, (-1, heaps_per_chunk))
```
and then based on that calculate whether I expect to see a sensor update for that chunk. Because if the large missing range doesn't fall exactly on chunk boundaries then the sensor updates get misaligned and all the test logic from that point forward just breaks.

test/fgpu/test_engine.py
james-smith-za: So I didn't have any reasoning for this, I added it and then the test started working.

But I think it was masking something that I had misunderstood. See my other comment in the thread.

test/fgpu/test_engine.py
james-smith-za: In retrospect I think it was floating-point rounding. The individual spectra are close enough to each other that often multiple of them get the same output when converting ADC to unix. So which one was being matched was effectively random.

src/katgpucbf/fgpu/engine.py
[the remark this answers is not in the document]
bmerry: I don't think it's worth trying to change the architecture. I'd suggested separating the pols when I thought the `present` array distinguished them anyway, but I was probably confusing the `input` versus `output` present arrays.

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+
+        # Same with dst_present
+        if (remainder := len(dst_present) % heaps_per_output_chunk) != 0:
+            dst_present = np.append(dst_present, [False for _ in range(heaps_per_output_chunk - remainder)])
+        chunk_present = np.reshape(dst_present, (-1, heaps_per_output_chunk))
+
+        # Sensor is not present in the narrowband mode.
+        if output.decimation == 1:
+            # We can't just for-loop because we may not have an update every chunk.
+            sensor_update_iter = iter(sensor_update_dict["input1.dig-rms-dbfs"])  # Both sensors should be the same.
```
bmerry, line 966: Rather than trusting to be the same, why not stick a loop over them.

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+                            )
+                        else:
+                            assert sensor_update.status == aiokatcp.Sensor.Status.FAILURE
+
+                        try:
+                            sensor_update = next(sensor_update_iter)
+                        except StopIteration:
+                            # We are finished with sensor updates. The last
+                            # chunk doesn't finish so we don't expect anymore.
+                            break
```
bmerry, line 992: Simply stopping the test here means that an implementation that gets the first update correct but never updates again is going to pass.

The way I see it, there are 3 cases for a chunk:
- np.all(present): an update MUST be present with NOMINAL or WARN status
- if not, but np.any(present): an update MUST be present with FAILURE status
- not np.any(present): a single update MAY be present, in which case it must have FAILURE status.

I don't think you're verifying all those conditions.

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+        # Sensor is not present in the narrowband mode.
+        if output.decimation == 1:
+            # We can't just for-loop because we may not have an update every chunk.
+            sensor_update_iter = iter(sensor_update_dict["input1.dig-rms-dbfs"])  # Both sensors should be the same.
+            sensor_update = next(sensor_update_iter)
+
+            for chunk_time, present in zip(chunk_timestamps, chunk_present):
+                # Sensor timestamp should be close to the end of the chunk if # it exists.
+                diff = abs(sensor_update.timestamp - np.max(chunk_time))
+                if diff / chunk_time_step > 1:  # Somewhat arbitrary, if a chunk is skipped this will be large.
```
bmerry, line 972: This seems rather fragile: both `diff` and `chunk_time_step` are at the mercy of floating-point rounding, and in the case where you're one chunk behind, the value will be very close to 1. Would it make more sense to test with 0.5?

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+
+        # Sensor is not present in the narrowband mode.
+        if output.decimation == 1:
+            # We can't just for-loop because we may not have an update every chunk.
+            sensor_update_iter = iter(sensor_update_dict["input1.dig-rms-dbfs"])  # Both sensors should be the same.
+            sensor_update = next(sensor_update_iter)
+
+            for chunk_time, present in zip(chunk_timestamps, chunk_present):
+                # Sensor timestamp should be close to the end of the chunk if # it exists.
+                diff = abs(sensor_update.timestamp - np.max(chunk_time))
```
bmerry, line 971: Using `abs` concerns me - if the sensor timestamp is in the past then this condition will continue to be true and we'll never advance to the next sensor reading.

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+                    # don't have anything to compare it to for a while.
+                    assert not np.any(present)
+                else:
+                    if np.any(present) or len(present) == 1:
+                        # There were heaps in this chunk, or the edge case where a chunk is one heap.
+                        if np.all(present):
+                            assert (
+                                sensor_update.status == aiokatcp.Sensor.Status.NOMINAL
+                                or sensor_update.status == aiokatcp.Sensor.Status.WARN
+                            )
```
bmerry, line 983:
```suggestion
                            assert sensor_update.status in {aiokatcp.Sensor.Status.NOMINAL, aiokatcp.Sensor.Status.WARN}
```

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+            for chunk_time, present in zip(chunk_timestamps, chunk_present):
+                # Sensor timestamp should be close to the end of the chunk if # it exists.
+                diff = abs(sensor_update.timestamp - np.max(chunk_time))
+                if diff / chunk_time_step > 1:  # Somewhat arbitrary, if a chunk is skipped this will be large.
+                    # The next sensor update is some time in the future, so we
+                    # don't have anything to compare it to for a while.
+                    assert not np.any(present)
+                else:
+                    if np.any(present) or len(present) == 1:
+                        # There were heaps in this chunk, or the edge case where a chunk is one heap.
```
bmerry, line 978: Why is this an edge case? I suspect it's papering over another bug.

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
...
+        unix_timestamps = [time_converter.adc_to_unix(t) for t in timestamps]
+
+        # The sensor is updated once per out_item.
+        spectra_per_output_chunk = engine_server.chunk_jones // output.channels
+        heaps_per_output_chunk = spectra_per_output_chunk // spectra_per_heap
+
+        # Need to reshape timestamp list so that we can work in batches of
+        # chunk size. The last chunk isn't finished so we need to add a
+        # bit on the end.  Need to test this after the prometheus counters
+        # becaus this affects np arrays which those tests use to count things.
```
bmerry, line 952:
```suggestion
        # because this affects np arrays which those tests use to count things.
```

test/fgpu/test_engine.py
```diff
@@ -933,6 +937,60 @@ async def test_missing_heaps(
             == np.sum(~dst_present) * n_substreams
         )
 
+        # Information needed to check that the dig-rms-dbfs sensor is correctly
+        # going into FAILURE when there are missing samples.
+        time_converter = TimeConverter(SYNC_TIME, ADC_SAMPLE_RATE)
```
bmerry, line 942: It's no longer showing up on this revision, but I previously suggested making this a TIME_CONVERTER constant at the top of the file to avoid repeating this statement multiple times across tests.

src/katgpucbf/fgpu/engine.py
```diff
@@ -895,6 +896,33 @@ async def _chunk_send_and_cleanup(
...
+                    total_power = float(trg)
+                    avg_power = total_power / (out_item.n_spectra * self.output.spectra_samples)
+                    # Normalise relative to full scale. The factor of 2 is because we
+                    # want 1.0 to correspond to a sine wave rather than a square wave.
+                    avg_power /= ((1 << (self.engine.src_layout.sample_bits - 1)) - 1) ** 2 / 2
+                    # If for some reason there's zero power, avoid reporting
+                    # -inf dB by assigning the most negative representable value
+                    avg_power_db = 10 * math.log10(avg_power) if avg_power else np.finfo(np.float64).min
+                    self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
+                        avg_power_db, timestamp=self.engine.time_converter.adc_to_unix(out_item.end_timestamp)
```
bmerry, line 916: The code for updating feng-clip-cnt always uses a timestamp which is the start of the next chunk. When I wrote this code I assumed this would be the same thing, but when n_spectra < capacity I see it is not. We should probably make them consistent with each other.

I think that using the timestamp for the start of the next chunk will also make the testing easier/more reliable since it will be easily predictable within the test.

src/katgpucbf/fgpu/engine.py
```diff
@@ -681,6 +681,7 @@ async def _flush_out(self, new_timestamp: int) -> None:
         # data).
         accs = self._out_item.n_spectra // self.output.spectra_per_heap
         self._out_item.n_spectra = accs * self.output.spectra_per_heap
+        self._out_item.present[self._out_item.n_spectra :] = False
```
bmerry, line 684: I worry that this may be expensive in some corner cases (although it is probably fine). It would be cheaper to modify `_update_dig_power_sensors` to check
```python
if out_item.n_spectra == out_item.capacity and np.all(out_item.present):
```

bmerry (changes requested):

src/katgpucbf/fgpu/engine.py
james-smith-za: True, and that would solve this problem.

But later on we might check for all packets present and get surprised by this again.

Unless we do something like this for the `jjjOutQueueItem`: 

```python
@property 
def all_present(self) -> bool:
    return self.n_spectra == self.capacity and np.all(self.present)
```

bmerry: I like the all_present idea.

src/katgpucbf/fgpu/engine.py
```diff
@@ -387,6 +387,20 @@ def pols(self) -> int:  # noqa: D401
...
+
+        :attr:`OutQueueItem.present` does not get "reset" in-between chunks. If
+        one chunk is complete and it is filled with the value `True`, and the
+        last few heaps / batches of the next chunk processed by this Item go
+        missing, you won't be able to tell directly from _out_item.present.
+
+        The canonical way to check whether this item's data is complete is
+        therefore to use this property.
+        """
+        return self.n_spectra == self.capacity and np.all(self.present)  # type: ignore
```
james-smith-za, line 402: Without the `ignore` I got the following:

```
src/katgpucbf/fgpu/engine.py:402: error: Incompatible return value type (got "Literal[False] | bool_", expected "bool")  [return-value]
```

I was too tired to fight with it in this case.

bmerry: np.all returns a numpy bool, but the signature for this function says it returns a Python bool. So try:
```python
        return self.n_spectra == self.capacity and bool(np.all(self.present))
```

src/katgpucbf/fgpu/engine.py
```diff
@@ -377,11 +377,30 @@ def capacity(self) -> int:  # noqa: D401
         # basically, the number of spectra.
         return self.spectra.shape[0] * self.spectra.shape[2]
 
+    @property
+    def next_timestamp(self) -> int:  # noqa: D401
+        """Timestamp of the next chunk afer this one."""
```
bmerry, line 382:
```suggestion
        """Timestamp of the next :class:`OutQueueItem` after this one."""
```

src/katgpucbf/fgpu/engine.py
```diff
@@ -377,11 +377,30 @@ def capacity(self) -> int:  # noqa: D401
...
         return self.spectra.shape[3]
 
+    @property
+    def all_present(self) -> bool:  # noqa: D410
+        """All batches for this chunk are complete.
+
+        :attr:`OutQueueItem.present` does not get "reset" in-between chunks. If
+        one chunk is complete and it is filled with the value `True`, and the
+        last few heaps / batches of the next chunk processed by this Item go
+        missing, you won't be able to tell directly from _out_item.present.
```
bmerry, line 397: This is talking about implementation details rather than semantics. I'd rather say something like (you'll probably need to reflow it):
```suggestion
        Only the first :attr:`n_spectra` elements of :attr:`OutQueueItem.present` have
        defined values, so it is not safe to use ``np.all(item.present)``.
```

src/katgpucbf/fgpu/engine.py
```diff
@@ -377,11 +377,30 @@ def capacity(self) -> int:  # noqa: D401
...
+    def next_timestamp(self) -> int:  # noqa: D401
+        """Timestamp of the next chunk afer this one."""
+        return self.timestamp + self.capacity * self.spectra_samples
+
     @property
     def pols(self) -> int:  # noqa: D401
         """Number of polarisations."""
         return self.spectra.shape[3]
 
+    @property
```
bmerry, line 390: I feel like this should be a method rather than a property, because it takes O(n) time rather than O(1).

src/katgpucbf/fgpu/engine.py
```diff
@@ -895,6 +914,33 @@ async def _chunk_send_and_cleanup(
...
+                    avg_power_db = 10 * math.log10(avg_power) if avg_power else np.finfo(np.float64).min
+                    self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
+                        avg_power_db, timestamp=self.engine.time_converter.adc_to_unix(out_item.next_timestamp)
+                    )
+            else:
+                for pol in range(N_POLS):
+                    self.engine.sensors[f"input{pol}.dig-rms-dbfs"].set_value(
+                        np.finfo(np.float64).min,
+                        status=aiokatcp.Sensor.Status.FAILURE,
+                        timestamp=self.engine.time_converter.adc_to_unix(out_item.next_timestamp),
```
bmerry, line 941: You could  make this code slightly more DRY by extracting `self.engine.time_converter.adc_to_unix(out_item.next_timestamp)` into a local variable before the if-else.

test/fgpu/test_engine.py
```diff
@@ -933,6 +939,61 @@ async def test_missing_heaps(
...
+            # going into FAILURE when there are missing samples.
+            class Update(Enum):
+                NORMAL = 1
+                FAILURE = 2
+                OPTIONAL_FAILURE = 3
+
+            expected_updates = []
+            # The sensor is updated once per out_item.
+            spectra_per_output_chunk = engine_server.chunk_jones // output.channels
+            heaps_per_output_chunk = spectra_per_output_chunk // spectra_per_heap
```
bmerry, line 954: Should be batches_per_output_chunk.

src/katgpucbf/fgpu/engine.py
```diff
@@ -377,11 +377,30 @@ def capacity(self) -> int:  # noqa: D401
         # basically, the number of spectra.
         return self.spectra.shape[0] * self.spectra.shape[2]
 
+    @property
+    def next_timestamp(self) -> int:  # noqa: D401
```
bmerry, line 381: Take a look for where `max_end_out` is assigned: you can use this property to make this simplification:
```diff
-            max_end_out = self._out_item.timestamp + self._out_item.capacity * self.output.spectra_samples
+            max_end_out = self._out_item.next_timestamp

File: test/fgpu/test_engine.py [hunk 30]
```diff
@@ -933,6 +939,61 @@ async def test_missing_heaps(
             == np.sum(~dst_present) * n_substreams
         )
 
+        # Sensor is not present in the narrowband mode.
+        if output.decimation == 1:
+            # Information needed to check that the dig-rms-dbfs sensor is correctly
+            # going into FAILURE when there are missing samples.
```
bmerry, line 945: This comment seems a bit misplaced now (probably my fault when I hacked this approach together).

bmerry (changes requested):
Looking good - just some nitpicking.

test/fgpu/test_engine.py
james-smith-za: We did discuss this and I started making the change, but I wondered about the consistency.

There are other local variables used in the function such as `total_heaps` and `spectra_per_heap` which should probably also be referred to as batches?

test/fgpu/test_engine.py
james-smith-za: I went through it to try and make sure everything was sensible but this is obviously one that I missed. Removed.

src/katgpucbf/fgpu/engine.py
james-smith-za: If I remove this, flake8 gives me:
```
src/katgpucbf/fgpu/engine.py:921: error: Function "all_present" could always be true in boolean context  [truthy-function]
```
but I don't see how that's the case, I haven't used a literal.

Though I seem to recall mypy thought the left side of the AND was a literal too 🤔

james-smith-za: Duh. My bad.

bmerry (approved):

PR #774, opened 2024-05-28
Extract some repeated code for sending

bmerry:
- Create a common xbgpu.send.Send class
- Add a send_rate helper function to compute the send rate. In the case of fgpu, this now makes a correction to the send rate to account for the packet overheads, which was missing before. It still needs to be tested though.

Closes NGC-1146.


james-smith-za (approved):
Looks good to me.

PR #784, opened 2024-06-05
Accelerate qualification/tacv/test_linearity

bmerry:
Instead of using only one beam, test weights/gains in batches where each beam gets a different weight/gain so that multiple values can be tested in parallel.

Also update pytest-jenkins.ini to allow it to use more than the default of 100 Gb/s, to get more parallelism. I picked 150 Gbps as the limit since that's bigger than we expect the beamformer output to be in practice (worst case is 140 Gb/s for 4 beams of S-band), and I didn't want to go all the way to 200 Gb/s since we probably won't keep up at that rate.

On a small test this saved about 3 minutes over 12 tests.

See NGC-1343.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/user-attachments/files/15592619/report.pdf)


james-smith-za (approved):
Nice idea. I like it.

PR #787, opened 2024-06-07
Qual: make test_group_delay more robust

bmerry:
Instead of retrying 5 times, keep retrying an unlimited number of times but with a 30 second limit. Failures often happen quickly, and this gives scope for a much larger number of retries without allowing things to run for an excessive amount of time.

This still doesn't prevent failures with large numbers of engines (sometimes the system seems to get into a state where it just doesn't keep up), but it does reduce the probability.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/user-attachments/files/15722067/report.pdf)

Relates to NGC-938.


james-smith-za (approved):
Seems a sensible approach.

PR #789, opened 2024-06-10
Qualification: make default_antennas configurable

bmerry:
Ideally one wants to make it large, so that tests that parallelise over antennas will run quickly, but it can't be so large that one runs out of resources. So it makes sense for it to be configurable.

See NGC-938.


james-smith-za (approved):

bmerry:
> I'll add a qualification report once it's run with a different default_antennas value. When not overridden in the config, the report shouldn't look any different.

The report is too big to attach.

PR #790, opened 2024-06-11
Make qualification reporting more robust to failed CBFs

bmerry:
When a CBF fails to configure, the tests don't even get to the point of reporting their blurbs, which was leading to errors about inconsistent names/blurbs. We should probably tweak the conftest to fix that, but for now this makes the reporting more robust so that we can at least get a PDF out the other end.

Relates to NGC-938.

- [ ] If qualification tests are changed: attach a sample qualification report: I have a file but it's too big to attach.

### DIFF
--- a/qualification/report/generate_pdf.py
+++ b/qualification/report/generate_pdf.py
@@ -391,17 +391,23 @@ class ResultSet:
         self.group = self.results[0].group
         for result in self.results:
             assert result.base_nodeid == self.base_nodeid, "base_nodeids do not match"
-            if result.text_name != self.text_name:
-                raise ValueError(
-                    f"Inconsistent text names for tests in ResultSet ({result.text_name} != {self.text_name})"
-                )
-            if result.blurb != self.blurb:
-                raise ValueError(f"Inconsistent blurbs for tests in ResultSet {self.base_nodeid}")
+            self.text_name = self._merge(self.text_name, result.text_name, "text names")
+            self.blurb = self._merge(self.blurb, result.blurb, "blurbs")
             self.duration += result.duration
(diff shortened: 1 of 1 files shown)


qualification/report/generate_pdf.py
```diff
@@ -391,18 +391,24 @@ def __init__(self, results: Sequence[Result]) -> None:
...
 
+    def _merge(self, a: str, b: str, name: str) -> str:
+        """Combine a property from two results.
+
+        If either is blank, the other is used. If both are non-blank and they
+        don't match, raise a :exc:`ValueError`.
+        """
+        if a != b and a and b:
+            raise ValueError(f"Inconsistent {name} for tests in ResultSet {self.base_nodeid}")
+        return a or b
```
james-smith-za, line 410: This is a rather weird syntax choice. When did Python start returning a string in this kind of case?

I'd have expected a boolean return given what's here.

james-smith-za (approved):

qualification/report/generate_pdf.py
bmerry: It's been like that as far as I can recall. See https://docs.python.org/3/reference/expressions.html#boolean-operations. Mostly it's an annoyance (e.g. it's the reason you needed to wrap `bool(...)` around a numpy boolean expression, to force the result of an `and` to be a Python bool), but in this case it's actually pretty handy (the link above describes exactly this use case, so I feel justified in saying it's Pythonic :-).

Fun fact: Python didn't originally have the ternary syntax ("x if cond else y") and the idiomatic way to do it was `cond and x or y` although if `x` was falsy it wouldn't work.

PR #792, opened 2024-06-11
Qualification: increase antenna counts

bmerry:
This is about as big an array as we can run with the hardware currently available.

See NGC-938.

- [ ] If qualification tests are changed: attach a sample qualification report: will run overnight

### DIFF
--- a/qualification/pytest-jenkins.ini
+++ b/qualification/pytest-jenkins.ini
@@ -11,8 +11,8 @@ use_ibv = true
 log_cli = true
 log_cli_level = info
 addopts = --report-log=report.json
-default_antennas = 8
-max_antennas = 10
+default_antennas = 26
+max_antennas = 40
 wideband_channels = 1024 4096 8192 32768
 narrowband_channels = 32768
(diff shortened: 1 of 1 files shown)


qualification/pytest-jenkins.ini
```diff
@@ -11,8 +11,8 @@ use_ibv = true
 log_cli = true
 log_cli_level = info
 addopts = --report-log=report.json
-default_antennas = 8
-max_antennas = 10
+default_antennas = 26
+max_antennas = 40
```
bmerry, line 15: If you're wondering why I haven't used 40 for both: antenna-channelised-voltage tests use a dsim per antenna, so 26 is as much as the system can handle right now. The other tests use only a single dsim.

bmerry:
@james-smith-za @amishatishpatel I requested you both for review, but it only needs one review - so whoever gets to it first. Also feel free to just merge if I've already clocked out, so that it'll get run tonight.

james-smith-za (approved):

PR #793, opened 2024-06-11
Increase qualification timeout to 10 hours

bmerry:
Not sure if this will be needed, but after #792 it might be.

See NGC-938.

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
   }
 
   options {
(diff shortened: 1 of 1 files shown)


bmerry:
I'm going to merge this directly so that it gets used tonight. @james-smith-za @amishatishpatel let's take a look tomorrow to see if the timeout is sensible. 10 hours ensures the test is finished by 9am in the worst case, so that if it does time out it'll do so before the standup.

PR #795, opened 2024-06-12
More naming convention cleanup

amishatishpatel:
Including:
* rx -> recv, 
* tx -> send,
* Conform to "preamble" rather than "header" size,
* Remove an old mention of a since-removed class.

I looked through katsdpcontroller and I don't think the move to `--send-enabled` (from `--tx-enabled`) breaks anything.

Closes NGC-1158.


src/katgpucbf/recv.py
```diff
@@ -41,9 +41,9 @@
 )
 
 #: Number of chunks before rx sensor status changes
```
bmerry, line 43:
```suggestion
#: Number of chunks before recv sensor status changes
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -1016,7 +1016,7 @@ class XBEngine(DeviceServer):
     src_ibv
         Use ibverbs for input.
     src_affinity
-        Specific CPU core to assign the RX stream processing thread to.
+        Specific CPU core to assign the RECV stream processing thread to.
```
bmerry, line 1019:
```suggestion
        Specific CPU core to assign the RECV stream processing thread to.
```
```suggestion
        Specific CPU core to assign the receive stream processing thread to.
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -1026,7 +1026,7 @@ class XBEngine(DeviceServer):
         The number of consecutive batches to store in the same chunk. The higher
         this value is, the more GPU and system RAM is allocated, the lower,
         the more work the Python processing thread is required to do.
-    rx_reorder_tol
+    recv_reorder_tol
```
bmerry, line 1029: I see we've got further inconsistency, in that parameters also get named `src` and `dst` rather than `recv` and `send`. In the longer term we should unify that; in the shorter term I'd suggest it would be more locally consistent to name this `src_reorder_tol` to match the other `src_*` parameters.

src/katgpucbf/xbgpu/main.py
```diff
@@ -352,7 +352,7 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
     parser.add_argument("--dst-ttl", type=int, default=DEFAULT_TTL, help="TTL for outgoing packets [%(default)s]")
     parser.add_argument("--dst-ibv", action="store_true", help="Use ibverbs for output [no].")
     parser.add_argument(
-        "--tx-enabled",
+        "--send-enabled",
```
bmerry, line 355: doc/control.rst and scratch/xbgpu/run-xbgpu.sh need updating for this change.

bmerry (changes requested):
Generally looks good.

@james-smith-za what do you think about renaming command-line options that start with `--src-` or `--dst-` to instead be `--recv-` or `--send-`? It might have some impact on katsdpcontroller but maybe it's worth biting that bullet now.

james-smith-za:
> Generally looks good.
> 
> @james-smith-za what do you think about renaming command-line options that start with `--src-` or `--dst-` to instead be `--recv-` or `--send-`? It might have some impact on katsdpcontroller but maybe it's worth biting that bullet now.

I wouldn't be opposed to this.

james-smith-za:
That being said though, I don't want to shift the goalposts and increase the scope for the PR.

@amishatishpatel I would be more than fine to finish off the issues with this one and do the src / dst change as @bmerry suggested in a separate PR.

bmerry (approved):

PR #798, opened 2024-06-13
Speed up test_weight_mapping

bmerry:
Speed it up mostly be rewriting the test to reduce latency between changing settings and seeing results (see commit message on first commit).

Also speed the test up generally by reducing buffer sizes to reduce latency.

Relates to NGC-1343.

### DIFF
--- a/qualification/__init__.py
+++ b/qualification/__init__.py
@@ -496,6 +496,7 @@ class TiedArrayChannelisedVoltageReceiver(XBReceiver):
             n_bits_per_sample=self.n_bits_per_sample,
             n_spectra_per_heap=self.n_spectra_per_heap,
             n_samples_between_spectra=self.n_samples_between_spectra,
+            decimation_factor=self.decimation_factor,
             use_ibv=use_ibv,
         )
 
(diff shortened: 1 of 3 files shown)


bmerry:
I don't have a qualification test to verify the speedup yet, because the one I started fell over when the Docker registry did. I'll re-run it this weekend.

james-smith-za (approved):
I like it.

Using gaussian noise just gives you more channels to work with and thus more parallelism?

bmerry:
> Using gaussian noise just gives you more channels to work with and thus more parallelism?

Kinda - but it's not so much about parallelism as only having to configure the F-engines once instead of inside the loop, because every time you change the settings there is a big latency before you see the outcome.

james-smith-za:
> > Using gaussian noise just gives you more channels to work with and thus more parallelism?
> 
> Kinda - but it's not so much about parallelism as only having to configure the F-engines once instead of inside the loop, because every time you change the settings there is a big latency before you see the outcome.

Ah yes that does make a lot of sense.

bmerry:
The latest test passed in 5h26m (compared to about 6h50m for `main`), so I'm going to merge.

PR #801, opened 2024-06-18
{src, dst} -> {recv, send}

amishatishpatel:
As per @bmerry's suggestion on #795.

I've purposefully left the following unchanged, to be discussed during this PR review
* the positional `src` arguments (for fgpu and xbgpu), as well as 
* the `--output` `dst` arguments.

- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match - see https://github.com/ska-sa/katsdpcontroller/pull/745

Closes NGC-1158.

### DIFF
--- a/scratch/xbgpu/run-xbgpu.sh
+++ b/scratch/xbgpu/run-xbgpu.sh
@@ -23,8 +23,8 @@ rx_comp=$rx_affinity
 tx_affinity=$((affinity + 1))
 tx_comp=$tx_affinity
 other_affinity=$tx_affinity
-src_mcast="239.10.10.$((10 + index)):7148"
-dst_mcast="239.10.11.$((10 + index)):7148"
+src="239.10.10.$((10 + index)):7148"
+dst="239.10.11.$((10 + index)):7148"
 channel_offset=$((channels_per_substream * index))
 katcp_port="$((7140 + index))"
 prom_port="$((7150 + index))"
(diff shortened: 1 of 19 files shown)


amishatishpatel:
I requested a review from you both as this update did touch quite a lot of files. Don't necessarily want to put it all on one person to review. Please and thank you in advance.

scratch/xbgpu/run-xbgpu.sh
```diff
@@ -24,7 +24,7 @@ tx_affinity=$((affinity + 1))
 tx_comp=$tx_affinity
 other_affinity=$tx_affinity
 src_mcast="239.10.10.$((10 + index)):7148"
-dst_mcast="239.10.11.$((10 + index)):7148"
+send_mcast="239.10.11.$((10 + index)):7148"
```
bmerry, line 27: For consistency with the argument names (and with run-fgpu.sh), perhaps call these `src` and `dst`. But whatever you call them, be consistent between the two variables.

bmerry (changes requested):
All looks reasonable to me - just a couple of bits of bikeshedding. I'm okay with keeping the names `src` and `dst` for the source and destination addresses - I think it would read awkwardly to try to force them to contain recv/send.

scratch/xbgpu/run-xbgpu.sh
amishatishpatel: Too right, my apologies. I think this line caught a stray bullet when updating instances of `dst_`.

james-smith-za: In this case I think `src` and `dst` are fine - it makes grammatical sense to talk about the source and destination multicast addresses.

james-smith-za (approved):
All looks fine to me, once @bmerry's comments are addressed.

test/xbgpu/conftest.py
```diff
@@ -21,22 +21,22 @@
 
 
 @pytest.fixture
-def n_src_streams() -> int:  # noqa: D401
+def n_recv_streams() -> int:  # noqa: D401
     """Number of source streams for an xbgpu instance."""
```
amishatishpatel, line 25: Argh, I'm making work for myself here. Related to your fgpu/main.py observation @bmerry, should I update the docstring here? This does seem to lend itself more to `src` rather than `recv`; I'm happy to roll back the change here.

bmerry: I wouldn't bother changing the docstring. Things are looking good.

bmerry (approved):

PR #806, opened 2024-06-19
Qualification: share CBFs between packages

bmerry:
Give up on pytest's fixture scoping support, and manage caching of CBFs explicitly. Previous attempts to use session-scoped fixtures have gone wrong because pytest doesn't cope well with per-test parametrisation of session-scoped fixtures (it matches them by position in the list of parameters, rather than the parameter value).

This should significantly speed up tests by allowing each CBF to be reused between the tied_array_channelised_voltage and baseline_correlation_products packages.

The sorting function is a little hacky and there is a small chance it will cause more reconfigurations than necessary. However, correctness is ensured, because a CBF is only reused if the exact config dictionary matches.

- [x] If qualification tests are changed: attach a sample qualification report: [report2.pdf](https://github.com/user-attachments/files/15902148/report2.pdf)

Relates to NGC-1343.

### DIFF
--- a/qualification/antenna_channelised_voltage/conftest.py
+++ b/qualification/antenna_channelised_voltage/conftest.py
@@ -19,7 +19,7 @@
 import pytest
 
 
-@pytest.fixture(scope="package")
+@pytest.fixture
 def n_dsims(n_antennas: int):
     """Give every simulated antenna its own dsim."""
     return n_antennas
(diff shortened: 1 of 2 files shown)


bmerry:
Note that the attached report is on a smaller sample of configurations, and shows that the tests are sharing CBFs. I don't yet know exactly how much faster this will be.

james-smith-za (approved):
Interesting idea.

I guess the proof of the pudding is in the eating, so let's give this a whirl and see how it goes.

PR #807, opened 2024-06-21
Change `heap` to `batch` where appropriate

james-smith-za:
I kept the change to only internal interfaces, so basically the F-engine output.

I must confess I have mixed feelings about this. The smaller changes were straightforward enough but they left "spectra-per-heap" standing out like a sore thumb. So I changed that. But that brings the internal variable names out-of-sync with the ICD-defined sensor (which is implemented in the controller and really has no bearing here).

Closes NGC-1358.

### DIFF
--- a/scratch/xbgpu/benchmarks/beamform_bench.py
+++ b/scratch/xbgpu/benchmarks/beamform_bench.py
@@ -30,7 +30,7 @@ def main():
     parser.add_argument(
         "--channels-per-substream", type=int, default=16, help="Channels processed by one engine [%(default)s]"
     )
-    parser.add_argument("--spectra-per-heap", type=int, default=256, help="Spectra in each batch [%(default)s]")
+    parser.add_argument("--spectra-per-batch", type=int, default=256, help="Spectra in each batch [%(default)s]")
     parser.add_argument("--heaps-per-fengine-per-chunk", type=int, default=32, help="Batches per chunk [%(default)s]")
     parser.add_argument("--beams", type=int, default=4, help="Number of dual-pol beams [%(default)s]")
     parser.add_argument("--passes", type=int, default=10000, help="Number of times to repeat the test [%(default)s]")
--- a/scratch/xbgpu/benchmarks/beamform_bench.py
+++ b/scratch/xbgpu/benchmarks/beamform_bench.py
@@ -38,7 +38,7 @@ def main():
 
     ctx = katsdpsigproc.accel.create_some_context()
     command_queue = ctx.create_command_queue()
-    template = BeamformTemplate(ctx, [0, 1] * args.beams, n_spectra_per_batch=args.spectra_per_batch)
+    template = BeamformTemplate(ctx, [0, 1] * args.beams, n_spectra_per_batch=args.spectra_per_heap)
     fn = template.instantiate(
         command_queue,
         n_batches=args.heaps_per_fengine_per_chunk,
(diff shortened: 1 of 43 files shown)


bmerry:
> I must confess I have mixed feelings about this. The smaller changes were straightforward enough but they left "spectra-per-heap" standing out like a sore thumb. So I changed that. But that brings the internal variable names out-of-sync with the ICD-defined sensor (which is implemented in the controller and really has no bearing here).

My feeling is that a "spectrum" is a collection of information about a signal at a range of frequencies at a single point in time. Thus, heaps do contain spectra - they're just spectra with less bandwidth than the spectra contained in batches. There is nothing fundamental about the bandwidth of a batch (in the narrowband case, it's not even the same as the digitised bandwidth), so I don't think that spectra-per-batch is any more "right" than spectra-per-heap. Hence, we might as well stick to spectra-per-heap for consistency with the ICD sensor (which in turn we should keep for backwards compatibility).

I'll hold off detailed review until we decide where to go with this.

james-smith-za:
OK. The argument is persuasive.
I kept the spectra-per-heap work to a single commit so I should just be able to reverse that commit.

scratch/xbgpu/benchmarks/beamform_bench.py
```diff
@@ -30,7 +30,7 @@ def main():
     parser.add_argument(
         "--channels-per-substream", type=int, default=16, help="Channels processed by one engine [%(default)s]"
     )
-    parser.add_argument("--spectra-per-heap", type=int, default=256, help="Spectra in each batch [%(default)s]")
+    parser.add_argument("--spectra-per-batch", type=int, default=256, help="Spectra in each batch [%(default)s]")
```
bmerry, line 33: I think you missed reverting this.

scratch/xbgpu/benchmarks/correlate_bench.py
```diff
@@ -29,7 +29,7 @@ def main():
     parser.add_argument(
         "--channels-per-substream", type=int, default=16, help="Channels processed by one engine [%(default)s]"
     )
-    parser.add_argument("--spectra-per-heap", type=int, default=256, help="Spectra in each batch [%(default)s]")
+    parser.add_argument("--spectra-per-batch", type=int, default=256, help="Spectra in each batch [%(default)s]")
```
bmerry, line 32: I think you missed reverting this.

src/katgpucbf/fgpu/compute.py
```diff
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2020-2023, National Research Foundation (SARAO)
+# Copyright (c) 2020-2024, National Research Foundation (SARAO)
```
bmerry, line 2: You've made copyright bumps on a number of files without changing them. Is that left over from the revert?

src/katgpucbf/fgpu/engine.py
```diff
@@ -863,7 +863,7 @@ async def run_processing(self) -> None:
                 await self._flush_out(end_timestamp)
 
             if end_timestamp >= max_end_in:
-                # We've exhausted the input buffer.
+                # We'v eexhausted the input buffer.
```
bmerry, line 866: I prefer this the way it was...

bmerry (changes requested):
The meat of it looks good - just some cleanup needed around your reverted change.

src/katgpucbf/fgpu/engine.py
james-smith-za: oops.

james-smith-za:
Closing in favour of #810

PR #811, opened 2024-06-25
Take advantage of Python 3.12

bmerry:
This is the low-hanging fruit of NGC-1380. I haven't yet used task groups or exception groups, because that's going to require deeper thinking about what the code is doing (plus reading up about the feature). I also haven't changed the generics to use PEP 695 because mypy doesn't support it yet.


bmerry:
There's a lot here, but the bulk of it is in the first commit which is run by machine (pyupgrade).

qualification/report/generate_pdf.py
```diff
@@ -670,7 +670,7 @@ def collate_results(results: Iterable[Result]) -> list[ResultSet]:
     for result_list in result_lists.values():
         result_set = ResultSet(result_list)
         # Check whether the CBF titles alone give unique subtitles
-        cbf_titles = set(result.cbf.title for result in result_list if result.cbf is not None)
+        cbf_titles = {result.cbf.title for result in result_list if result.cbf is not None}
```
james-smith-za, line 673: Interesting new syntax.

requirements-dev.txt
```diff
@@ -40,7 +40,6 @@ async-timeout==4.0.3
     # via
     #   -c requirements.txt
     #   aiokatcp
```
james-smith-za, line 42: Looks like it's still in requirements.txt based on this?
(the aiokatcp dependency can probably wait for another day)

qualification/report/generate_pdf.py
bmerry: Actually that's been around a long time, although I can't immediately find which Python version introduced it.

requirements-dev.txt
bmerry: Yes, it's only removed as a direct requirement. I expect it to still be pulled in indirectly for the foreseeable future.

qualification/report/generate_pdf.py
bmerry: Looks like it dates right back to Python 3.0 (and 2.7).

james-smith-za (approved):

PR #805, opened 2024-06-19
Remove prom_get helper function in xbgpu test

amishatishpatel:
It wasn't really needed anymore, even though there is still one intermediate usage of a PromDiff.

Checklist (if not applicable, edit to add `(N/A)` and mark as done):

- [x] (n/a) If dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`
- [x] (n/a) If modules are added/removed: use `sphinx-apidoc -efo doc/ src/` to update files in `doc/`
- [x] Ensure copyright notices are present and up-to-date
- [x] (n/a) If qualification tests are changed: attach a sample qualification report
- [x] (n/a) If design has changed: ensure documentation is up to date
- [x] (n/a) If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match

Closes NGC-1316.

PR #812, opened 2024-06-26
Add Prometheus get_sample* helper class

amishatishpatel:
And update uses of `PromDiff.get_sample_diff` accordingly.

I mainly went with this MO because I focused on [your original critique/observation](https://github.com/ska-sa/katgpucbf/pull/805#issuecomment-2178718952) of
> [...] the helper function is still valuable for readability

I know it's not common to request a review before Jenkins is done, but I have run the affected/
updated tests manually (on cbfgpu033) and they all passed fine.

Closes NGC-1316.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -95,6 +95,58 @@ class PromDiff:
             return after
 
 
+class PromGetHelp:
+    """Helper class for obtaining Prometheus metric values.
+
+    This is used when the developer is condifent on the return type and aides
+    code readability. It only needs the :class:`PromDiff` upon initialisation
+    and the user can continually obtain metric values against it.
+
+    Typical usage is::
+
+        labels = {"stream": stream_name}
+        promget = PromGetHelp(prom_diff, labels)
+        assert promget.value(name) == expected_value
+        new_labels = {"pol": "0"}
+        assert promget.diff(name, label_override=new_labels) == expected_diff
+
+    Parameters
+    ----------
+    prom_diff
+        Instance of :class:`PromDiff` that is used during a unit test.
+    labels
+        Optional identifiers for Prometheus metrics under test.
+    """
+
+    def __init__(self, prom_diff: PromDiff, labels: dict[str, str] | None = None) -> None:
+        self.prom_diff = prom_diff
+        self.labels = labels
+
+    def value(self, name: str, label_override: dict[str, str] | None = None) -> float:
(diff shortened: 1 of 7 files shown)


bmerry:
I'm not loving the approach of "this class is being unhelpful, so we'll wrap it in another class" instead of just fixing the first class; particularly since they don't use consistent naming.

Can you take a look at the NGC-1316-prom-diff-improve branch I've just pushed? It instead augments the existing PromDiff class to be more helpful, and it adds less code. I haven't checked to ensure that it's applied everywhere it could be.

By the way, I'm open to renaming the `get_sample_diff` method to just `diff` to keep things more compact.

bmerry (commented):
See comment above.

amishatishpatel:
Firstly, Re:
> I'm not loving the approach of "this class is being unhelpful, so we'll wrap it in another class" 

* I'm absolutely with you

Second, Re:
> instead of just fixing the first class; particularly since they don't use consistent naming.

* I must admit, I was under the (perhaps unfounded) impression that the PromDiff implementation was following some guide (style or otherwise) from Prometheus
* Which is why I didn't want to touch the PromDiff implementation itself
  * e.g. maybe a metric reading can be `None`? 

> Can you take a look at the NGC-1316-prom-diff-improve branch I've just pushed? It instead augments the existing PromDiff class to be more helpful, and it adds less code. I haven't checked to ensure that it's applied everywhere it could be.

* Having a look now and it's certainly more promising and feels much better as a change/update.
* Will make comments on the commit as I see fit.

> By the way, I'm open to renaming the `get_sample_diff` method to just `diff` to keep things more compact.

* Again, I welcome this wholeheartedly. Especially now that I know we control the implementation.

Can I close this PR?

bmerry:
> I must admit, I was under the (perhaps unfounded) impression that the PromDiff implementation was following some guide (style or otherwise) from Prometheus

Nope, it's just something I dreamt up.

> maybe a metric reading can be None?

I think Prometheus has a concept of a "tombstone" reading to indicate that a metric has gone away, but I think that's something on the server side. For our tests, a metric might not be present with particular labels, but if it's present it'll be float.

> Can I close this PR?

Sure. Feel free to take over the branch I created and make any additional changes you think are needed (e.g. renaming the query functions to shorter names).

amishatishpatel:
Closing this PR and building on @bmerry's (much better) suggestion in katgpucbf/NGC-1316-prom-diff-improve.

PR #813, opened 2024-06-26
Implement larger window for dig-rms-dbfs sensor

bmerry:
The sensor now reports on roughly the last one second of data, instead of just the last OutQueueItem. The exact interval is a whole number of output chunks.

The chunks are divided into windows according to their timestamps, with the same number of chunks in each window. The sensor is updated at the end of each window with the average power for the window.

Closes NGC-1369.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -59,7 +59,8 @@ from ..utils import (
     gaussian_dtype,
     steady_state_timestamp_sensor,
 )
-from . import DIG_RMS_DBFS_HIGH, DIG_RMS_DBFS_LOW, INPUT_CHUNK_PADDING, recv, send
+from . import DIG_RMS_DBFS_HIGH, DIG_RMS_DBFS_LOW, DIG_RMS_DBFS_WINDOW, INPUT_CHUNK_PADDING, recv, send
+from .accum import Accum
 from .compute import Compute, ComputeTemplate, NarrowbandConfig
 from .delay import AbstractDelayModel, AlignedDelayModel, LinearDelayModel, MultiDelayModel, wrap_angle
 from .output import NarrowbandOutput, Output, WidebandOutput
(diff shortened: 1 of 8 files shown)


src/katgpucbf/fgpu/engine.py
```diff
@@ -901,27 +902,42 @@ async def _chunk_send_and_cleanup(
...
-            if np.all(out_item.present):
-                for pol, trg in enumerate(dig_total_power):
-                    total_power = float(trg)
-                    avg_power = total_power / (out_item.n_spectra * self.output.spectra_samples)
+        all_present = np.all(out_item.present)
+        for pol, (accum, trg) in enumerate(zip(dig_total_power_accums, dig_total_power)):
+            power: int | None = int(trg)
+            if not all_present:
+                power = None
+            if measurement := accum.add(out_item.timestamp, out_item.next_timestamp, power):
```
james-smith-za, line 936: Walrus operator warming the cockles of my heart...

Perhaps I have misunderstood, but I thought that if `measurement` is `None` then you also want to update the sensor with Failure but it doesn't look as though it happens. Currently if `measurement` isn't none but `measurement.value` is none, there's a Failure update.

test/fgpu/test_accum.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_add_to_invalid(invalid_accum: Accum) -> None:
+    """Extending an invalid window does not validate it."""
+    assert invalid_accum.add(220, 300, 10) == Measurement(200, 300, None)
+
+
+def test_add_gap(valid_accum: Accum) -> None:
+    """Extending a window with a gap invalidates it."""
+    assert valid_accum.add(230, 250, 10) is None
+    assert valid_accum.add(250, 300, 20) == Measurement(200, 300, None)
```
james-smith-za, line 76: This one doesn't make sense to me.
The fixture has [200,220) with value 200
then we add [230,250) with value 10 but there's a gap from [220,230)
How does it become valid again when we add from [250,300)
?

src/katgpucbf/fgpu/engine.py
bmerry: `measurement is None` means "not enough time has passed since the last measurement." `measure.value is None` means "enough time has passed and we missed some data since the last measurement".

test/fgpu/test_accum.py
bmerry: My previous comment may help. Note that the *value* is None, which indicates that data was missing, but that we've seen enough to know we can issue this update.

james-smith-za: Ah yes that does make sense. Thanks for clarifying.

james-smith-za (approved):

PR #814, opened 2024-06-27
Make use of TaskGroup

bmerry:
Where asyncio.gather was used to start, run and await multiple tasks, use the new TaskGroup instead, which has better cancellation behaviour (cancels the other tasks as soon as one task fails, and potentially report multiple exceptions via an ExceptionGroup).

This is not used everywhere it could be yet. In particular, tasks started by engines are managed by aiokatcp's add_service_task, which does similar things to TaskGroup (although it does not raise exception groups). There are a number of other places where the creation of and waiting for a task are widely separated such that it would be inconvenient to use a context manager.

See NGC-1380.

### DIFF
--- a/scratch/benchmarks/remote.py
+++ b/scratch/benchmarks/remote.py
@@ -164,15 +164,14 @@ async def run_tasks(
         async with asyncio.timeout(timeout):
             for i in range(n):
                 # Wait until either the port is ready or the process dies
-                tasks = [
-                    asyncio.create_task(wait_port(server, port_base + i)),
-                    asyncio.create_task(procs[i].wait(check=True)),
-                ]
-                done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
-                for task in pending:
-                    task.cancel()
-                for task in done:
-                    task.result()  # Re-raise any exceptions
+                async with asyncio.TaskGroup() as tg:
+                    tasks = [
+                        tg.create_task(wait_port(server, port_base + i)),
+                        tg.create_task(procs[i].wait(check=True)),
+                    ]
+                    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                    for task in pending:
+                        task.cancel()
                 if tasks[1] in done:
                     raise RuntimeError("process shut down before becoming ready")
         return stack.pop_all()
--- a/src/katgpucbf/dsim/main.py
+++ b/src/katgpucbf/dsim/main.py
@@ -152,8 +152,12 @@ def first_timestamp(time_converter: TimeConverter, now: float, align: int) -> in
     return samples
 
 
-async def async_main() -> None:
-    """Asynchronous main entry point."""
+async def _async_main(tg: asyncio.TaskGroup) -> None:
+    """Real implementation of :func:`async_main`.
+
+    This is split into a separate function to avoid having to indent the whole
+    thing inside the `tg` context manager.
+    """
     args = parse_args()
     heap_size = args.heap_samples * args.sample_bits // BYTE_BITS
 
@@ -190,7 +194,7 @@ async def async_main() -> None:
     # Start descriptor sender first so descriptors are sent before dsim data.
     descriptor_heap = descriptors.create_descriptors_heap()
     descriptor_sender = DescriptorSender(descriptor_stream, descriptor_heap, SPEAD_DESCRIPTOR_INTERVAL_S)
-    descriptor_task = asyncio.create_task(descriptor_sender.run())
+    tg.create_task(descriptor_sender.run())
 
     if args.dither_seed is None:
         args.dither_seed = np.random.SeedSequence().entropy  # Generate a random seed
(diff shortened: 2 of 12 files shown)


scratch/benchmarks/remote.py
```diff
@@ -164,15 +164,14 @@ async def run_tasks(
...
-                for task in done:
-                    task.result()  # Re-raise any exceptions
+                async with asyncio.TaskGroup() as tg:
+                    tasks = [
+                        tg.create_task(wait_port(server, port_base + i)),
+                        tg.create_task(procs[i].wait(check=True)),
+                    ]
+                    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+                    for task in pending:
+                        task.cancel()
```
amishatishpatel, line 174: Ah, I gather the move to using a `TaskGroup` here is to avoid the old Line 175 of re-raising any exceptions?

src/katgpucbf/dsim/main.py
```diff
@@ -152,8 +152,12 @@ def first_timestamp(time_converter: TimeConverter, now: float, align: int) -> in
...
 
 
-async def async_main() -> None:
-    """Asynchronous main entry point."""
+async def _async_main(tg: asyncio.TaskGroup) -> None:
+    """Real implementation of :func:`async_main`.
+
+    This is split into a separate function to avoid having to indent the whole
+    thing inside the `tg` context manager.
+    """
```
amishatishpatel, line 160: Good call.

src/katgpucbf/fsim/main.py
```diff
@@ -312,15 +316,23 @@ async def async_main() -> None:
...
     sync_time = time.time()
-    await asyncio.gather(*(sender.run(sync_time, args.run_once) for sender in senders))
-    for sender in descriptor_senders:
-        sender.halt()
-    await asyncio.gather(*descriptor_tasks)
+    async with asyncio.TaskGroup() as sender_tg:
+        for sender in senders:
+            sender_tg.create_task(sender.run(sync_time, args.run_once))
+    for descriptor_sender in descriptor_senders:
+        descriptor_sender.halt()
```
amishatishpatel, line 329: Appreciate the rename as well (from `sender in descriptor_senders`).

test/fsim/test_main.py
```diff
@@ -79,12 +79,11 @@ def sender(args: argparse.Namespace, mock_queue) -> Sender:
...
-    task = asyncio.create_task(sender.run(0, False))
-    yield task
-    task.cancel()
-    with pytest.raises(asyncio.CancelledError):
-        await task
+    async with asyncio.TaskGroup() as tg:
+        await sender.stream.async_send_heap(sender.descriptor_heap, rate=0.0)
+        task = tg.create_task(sender.run(0, False))
+        yield task
+        task.cancel()
```
amishatishpatel, line 86: Seeing as you removed the `with pytest.raises`, does that mean we don't need to explicitly acknowledge(/catch?) the `asyncio.CancelledError`?
* I gather we're relying on the `TaskGroup` to step in somehow
* At the moment I'm just seeing them as two different entities (`pytest.raises` has a different job to `asyncio.TaskGroup`)

amishatishpatel (commented):
Couple comments for clarification before explicitly approving. Looks really neat.

scratch/benchmarks/remote.py
bmerry: I was generally trying to use TaskGroups where they fit. But yes, it's an advantage of TaskGroups. It will also better handle the case where multiple completed tasks raised exceptions (they'll all get bundled into an ExceptionGroup, while the old code would just raise the first one and drop the rest on the floor). I also realised that the old code doesn't actually wait for the cancelled tasks to complete (which might not be immediately if they have `final` code).

src/katgpucbf/fsim/main.py
bmerry: Thank mypy - it didn't like having `sender` be used as the loop variable for two loops with different types.

test/fsim/test_main.py
bmerry: TaskGroup's context manager swallows `CancelledError`. Technically I've changed the behaviour since I'm not longer verifying that a CancelledError is raised, but I think I was just abusing `pytest.raises` in the first place as a way to suppress the exception from bubbling out rather than to actually verify that it happened.

scratch/benchmarks/remote.py
amishatishpatel: > I also realised that the old code doesn't actually wait for the cancelled tasks to complete

Aha, it simply executed `task.cancel()` and moved on with life. Not to drag on, but out of interest
* Here we/you seem to be doing a similar thing (`task.cancel()`), but
* Now wrapped with an `async with asyncio.TaskGroup()`

How does the TaskGroup context help "resolve" incomplete cancelled tasks?

amishatishpatel: Ah, reading your reply on my comment regarding `pytest.raises`, I'm thinking
> the context manager swallows any `CancelledError`s.

test/fsim/test_main.py
amishatishpatel: Noted and noted.

amishatishpatel (approved):
Just one comment for my own slightly-related understanding. I'll happily research it in my own time if you've got other things to get to.

scratch/benchmarks/remote.py
bmerry: Yes. It also waits for all tasks in the task group to complete.

PR #815, opened 2024-06-28
Improve PromDiff for unit tests

amishatishpatel:
Can't take much credit for this one, this is all @bmerry. My attempt at #812 
was largely because I thought we didn't "own" our implementation of `PromDiff`.

I tweaked the class's docstring and renamed `get_sample_{value, diff}` to just
`{value, diff}`.

I noticed katgpucbf doesn't adopt/use the convention of adding a 
```
Raises
------
```
section in the docstring, so I left the discussion as-is in `PromDiff.{value, diff}`.

Closes NGC-1316.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -16,7 +16,8 @@
 # limitations under the License.
 ################################################################################
 
-from typing import Self
+import copy
+from typing import Mapping, Self
 
 import numpy as np
 import prometheus_client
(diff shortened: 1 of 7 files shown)


test/__init__.py
```diff
@@ -68,31 +101,37 @@ def _get_value(
...
-            return after - before
-        elif after is not None:
-            raise ValueError(f"Metric {name}{labels} did not exist at start")
-        else:
-            return after
+        effective_labels = self._get_labels(labels)
+        before = self._get_value(self._before, name, effective_labels)
+        after = self._get_value(self._after, name, effective_labels)
+        assert before is not None, f"Metric {name}{labels} did not exist at start"
+        assert after is not None, f"Metric {name}{labels} did not exist at start"
```
bmerry, line 123:
```suggestion
        assert after is not None, f"Metric {name}{labels} did not exist at end"
```

bmerry (changes requested):
LGTM. Just one typo (probably my fault) - feel free to fix and merge without another review.

PR #816, opened 2024-06-28
Upgrade pytest to v8.2.2

amishatishpatel:
This is the latest version of pytest at the moment. This did also
require a tweak to some other package versions.

See ticket for a brief recap of builds trialled et al.

- [ ] If qualification tests are changed: attach a sample qualification report - I can't quite remember how to fix my access to this, will attach a report once I do.

Closes NGC-1330.


setup.cfg
```diff
@@ -47,7 +47,8 @@ doc =
 test =
     katsdpsigproc[CUDA]
     async-solipsism
-    pytest
+    async-timeout
+    pytest>8
     pytest-asyncio>=0.18,<0.22
```
bmerry, line 52: I'm a bit torn. To avoid anyone else getting the cryptic incompatibilities, I'd consider
```suggestion
    pytest-asyncio>=0.21.2,<0.22
```
On the other hand if pytest 8.0.0 and pytest-asyncio 0.18 work together then this could be left as is. But I don't think it's worth spending the time to test that combination.

setup.cfg
```diff
@@ -47,7 +47,8 @@ doc =
 test =
     katsdpsigproc[CUDA]
     async-solipsism
-    pytest
+    async-timeout
```
bmerry, line 50: I think this might be a merge issue - async-timeout was removed as part of the Python 3.12 work, and it's snuck back in here.

setup.cfg
```diff
@@ -47,7 +47,8 @@ doc =
...
+    pytest>8
```
bmerry, line 51:
```suggestion
    pytest>=8
```
It's standard for lower bounds to be `>=`. What you've written means "pytest 8.x+, except for 8.0.0".

bmerry (changes requested):

setup.cfg
amishatishpatel: Fair points all around. Will update to set the lower bound at 0.21.2

setup.cfg
amishatishpatel: Yeah I was wondering where that came from. I put it down to me manually removing packages before running `pip-compile` - which, in hindsight, shouldn't have affected `setup.cfg`.

setup.cfg
amishatishpatel: I think this was my attempt at ~coercing it to get a version newer than 8.0.x. In the end I manually set it to 8.2.2 in `requirements-dev.txt`.
* I'll update to `>=`

bmerry (approved):

amishatishpatel:
Thanks for the unprompted approval. I was waiting for Jenkins to build it - which I see it didn't even pick up.

Either way, merging now.

PR #818, opened 2024-07-02
Run the microbenchmarks under Jenkins

bmerry:
This will make it more likely that we'll notice when they get broken by
API changes.

Closes NGC-1141.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -71,12 +71,35 @@ pipeline {
 
     stage('Parallel stage') {
       parallel {
-        stage('Compile and test memcpy_loop') {
+        stage('Compile and test microbenchmarks') {
+          options { timeout(time: 5, unit: 'MINUTES') }
           steps {
             dir('scratch') {
               sh 'make clean'
               sh 'make'
               sh './memcpy_loop -T'
+
+              // We just want to know if they run without crashing, so we use a small
+              // number of passes to speed things up.
+              sh 'fgpu/benchmarks/compute_bench.py --kernel all --passes 10'
+              sh 'fgpu/benchmarks/compute_bench.py --kernel ddc --narrowband --passes 10'
+              sh 'fgpu/benchmarks/compute_bench.py --kernel pfb_fir --passes 10'
+              sh 'fgpu/benchmarks/compute_bench.py --kernel fft --passes 10'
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -71,12 +71,35 @@ pipeline {
...
+              sh 'fgpu/benchmarks/fft_bench.py --mode r2c --passes 10'
+              sh 'fgpu/benchmarks/fft_bench.py --mode c2c --passes 10'
+
+              sh 'xbgpu/benchmarks/beamform_bench.py --passes 10'
+              sh 'xbgpu/benchmarks/correlate_bench.py --passes 10'
+
+              sh './gpu_copy.py htod --repeat 10'
+              sh './gpu_copy.py dtoh --repeat 10'
+              sh './gpu_copy.py dtod --repeat 10'
+              sh './gpu_copy.py htod --mem huge --fill 1 --repeat 10'
```
amishatishpatel, line 102: > [...] so we use a small number of passes to speed things up

Again, out of interest, are these normally run with `--passes 100` or more?

Jenkinsfile
```diff
@@ -71,12 +71,35 @@ pipeline {
 
     stage('Parallel stage') {
       parallel {
-        stage('Compile and test memcpy_loop') {
+        stage('Compile and test microbenchmarks') {
+          options { timeout(time: 5, unit: 'MINUTES') }
```
amishatishpatel, line 75: I see the last successful run took 24s. 
* Do you envision it ever hitting this timeout? I mean, more tests obviously take more time.
* And/or was it just randomly chosen at 5 minutes?

amishatishpatel (approved):
I mean, Jenkins doesn't lie, it works. What did you have to resolve to get this new stage to pass?

Jenkinsfile
bmerry: Yes normally much higher to ensure that the GPU is fully warmed up. Run the commands with `--help` to see the defaults.

Jenkinsfile
bmerry: I don't expect it to time out, but a breakage could lead to something getting totally wedged and hanging.

I just noticed that we'd never supplied a timeout, so I copied the one used for building documentation.

PR #819, opened 2024-07-03
Make qualification report cope better with CBFs that don't start

bmerry:
Introduce a FailedCBF class which represents a CBF that we tried to configure but which failed. It shares (via CBFBase) the basic properties needed to report it as a configuration in the qualification test report, but retains the exception that caused the failure. That allows for a few changes:
1. We can cache it in the CBFCache, so that we don't keep retrying a broken configuration (so that we fail faster, although on the flip side we not longer retry for the next test so transient failures will affect all tests for a configuration).
2. We can re-raise the exception in the `cbf` fixture, so that the console output correctly reflects the original error that occurred.
3. We can treat this broken configuration almost the same as working configurations in the report. That fixes up the weird appearances reported in NGC-951.
4. The failure can appear just once in the report, under the CBF configuration section; the individual tests can then just report that the CBF failed to start.

Note that the first three commits are refactoring to make the codebase more manageable, and don't change behaviour.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/user-attachments/files/16082621/report.pdf)

Closes NGC-951.


qualification/report/generate_pdf.py
```diff
@@ -243,7 +243,8 @@ class CBFConfiguration:
 
     mode_config: dict
     uuid: UUID
-    tasks: dict[str, Task]
+    tasks: dict[str, Task]  # Empty if the CBF failed
+    error: str | None  # Empty if the CBF started successfully
```
amishatishpatel, line 247: Hate to be pedantic, but will it 'Empty' or `None`? i.e. the actual value.
* I see on Line 564 you don't have a default for the `msg.get("error")`, 
* So I'm leaning towards `None` if the CBF started successfully

qualification/report/generate_pdf.py
```diff
@@ -1050,7 +1058,11 @@ def _doc_result(section: Container, result: Result, tmp_dir: pathlib.Path, figur
                     procedure_table.add_row((MultiColumn(2, align="|c|", data=mp),))
                 procedure_table.add_hline()
 
-    if result.failure_messages:
+    if result.cbf is not None and result.cbf.error is not None:
+        # In this case we'll already have details about the failure as part
+        # under the heading for the CBF, so don't repeat the whole thing here.
+        section.append(TextColor("red", f"CBF failed to start: {result.cbf.error.splitlines()[-1]}"))
```
amishatishpatel, line 1064: > [...] we'll already have details about the failure **as part under the heading** for the CBF, [...]

"as part under the heading"? 
* Typo? Shorthand? Not quite understanding the comment.

amishatishpatel (changes requested):
Minor tweaks requested. More generally, it seems you've changed the layout/info reported in the Result Summary - no complaints there. It's likely that my understanding of how the qualification tests run is wholly incomplete, but worth asking (Re: attached report.pdf and extra-long screenshot)
* I see the ACV tests only used one configuration to run (Configuration 3),
* And that configuration happened to be the only one that failed,
* Perhaps point 1 in your PR blurb explains why the test wasn't run for other configurations
> We can cache it in the CBFCache, so that we don't keep retrying a broken configuration (so that we fail faster, although on the flip side **we not longer retry for the next test so transient failures will affect all tests for a configuration**).

image

qualification/conftest.py
```diff
@@ -636,6 +438,10 @@ async def cbf(
     been reset to a default state, with the dsim outputting zeros.
     """
     cbf = await cbf_cache.get_cbf(cbf_config, cbf_mode_config)
+    pdf_report.config(cbf=str(cbf.uuid))
+    if isinstance(cbf, FailedCBF):
+        raise cbf.exc
```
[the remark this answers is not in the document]
bmerry: pytest.fail will log a traceback of the code running at the point `pytest.fail` was called, but you'd lose the traceback of the original exception (which shows where you were when trying to start the CBF failed). The `cbf.exc` object contains the original traceback, which will be shown by pytest.

qualification/report/generate_pdf.py
bmerry: Good catch - empty is not the same as None. I'll fix.

bmerry:
> Minor tweaks requested. More generally, it seems you've changed the layout/info reported in the Result Summary - no complaints there. It's likely that my understanding of how the qualification tests run is wholly incomplete, but worth asking (Re: attached report.pdf and extra-long screenshot)
> 
>     * I see the ACV tests only used one configuration to run (Configuration 3),
> 
>     * And that configuration happened to be the only one that failed,
> 
>     * Perhaps point 1 in your PR blurb explains why the test wasn't run for other configurations
> 
> 
> > We can cache it in the CBFCache, so that we don't keep retrying a broken configuration (so that we fail faster, although on the flip side **we not longer retry for the next test so transient failures will affect all tests for a configuration**).

This report is just one where I manually stopped the test after running one configuration, which I manually failed. In practice, other configurations would still proceed and be tested.

amishatishpatel (approved):
Noted, noted and noted.

PR #821, opened 2024-07-04
Handle decimation factors less than 8

bmerry:
The padding added to each input chunk needs to be a multiple of BYTE_BITS (8), but that doesn't necessarily happen when the narrowband decimation is small.

Closes NGC-1393.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -1233,6 +1233,7 @@ class Engine(aiokatcp.DeviceServer):
         self._copy_queue = context.create_command_queue()
 
         extra_samples = max_delay_diff + max(output.window for output in outputs)
+        extra_samples = accel.roundup(extra_samples, BYTE_BITS)
         if extra_samples > self.recv_layout.chunk_samples:
             raise RuntimeError(f"chunk_samples is too small; it must be at least {extra_samples}")
         self.n_samples = self.recv_layout.chunk_samples + extra_samples


bmerry:
Let me know if you think this should have a unit test. I considered that low priority since this isn't a made we're imminently planning to use, and I implemented this just so that I could briefly check the performance.

james-smith-za (approved):

james-smith-za:
> Let me know if you think this should have a unit test. I considered that low priority since this isn't a made we're imminently planning to use, and I implemented this just so that I could briefly check the performance.

What would you test exactly? That 4, 8 and 16 all don't crash something?

bmerry:
> > Let me know if you think this should have a unit test. I considered that low priority since this isn't a made we're imminently planning to use, and I implemented this just so that I could briefly check the performance.
> 
> What would you test exactly? That 4, 8 and 16 all don't crash something?

Ideally that they produce correct results end-to-end. Currently I think TestEngine only tests with decimation=8.

james-smith-za:
> > > Let me know if you think this should have a unit test. I considered that low priority since this isn't a made we're imminently planning to use, and I implemented this just so that I could briefly check the performance.
> > 
> > 
> > What would you test exactly? That 4, 8 and 16 all don't crash something?
> 
> Ideally that they produce correct results end-to-end. Currently I think TestEngine only tests with decimation=8.

I wouldn't be opposed to this.

PR #822, opened 2024-07-04
Run Jenkins pytests with -n 4

bmerry:
This brings branch builds to < 5 minutes. Relates to NGC-1363, in that that workaround for that ticket involves running only one build at a time per host and hence we have enough resources to run things in parallel. We might be able to increase the parallelism further, but the pre-commit checks would then become a bottleneck.


james-smith-za (approved):
Nice! I like it.
This will go some ways towards not blocking each other while we wait for builds.

PR #825, opened 2024-07-08
Make global memory access for dithering more efficient

bmerry:
Save and restore only the state that's actually needed. This reduces memory usage and global memory traffic for beamforming. In some microbenchmarks it makes beamforming 1.6x more efficient. It also simplifies some of the code by eliminating the need to measure the size and alignment of curandStateXORWOW_t at runtime.

Also removed a redundant initialisation of random state in beamform_bench.py.

This is indirectly related to NGC-1264: this improvement is needed to make F-engine dithering feasible.

### DIFF
--- a/src/katgpucbf/curand_helpers.py
+++ b/src/katgpucbf/curand_helpers.py
@@ -16,15 +16,9 @@
 
 """Helpers to initialise random state with curand.
 
-The :c:struct:`curandXORWOW_t` struct defined by curand is unnecessarily large
-for our purposes, because it retains state needed to generate Gaussian
-distributions (Box-Muller transform). To reduce global memory traffic, we use
-a different type we define (:c:struct:`randState_t`) to hold random states in
-global memory, together with helpers that save and restore this smaller state
-from a private :c:struct:`curandXORWOW_t` used within a kernel.
-
-Code using this module should thus NOT generate Gaussian distributions without
-understanding the implications.
+See :ref:`dithering` for an explanation of why we introduce a separate
+:c:struct:`randState_t` structure. Code using this module should **not**
+generate Gaussian distributions without understanding the implications.
 """
 
 from importlib import resources
@@ -37,7 +31,7 @@ from katsdpsigproc.abc import AbstractCommandQueue, AbstractContext
 RAND_STATE_SIZE = 24
 #: alignof(randState_t)
 RAND_STATE_ALIGNMENT = 8
-#: dtype corresponding to randState_t
+#: opaque dtype corresponding to randState_t (only size and alignment matter)
 RAND_STATE_DTYPE = np.dtype(
     {"names": ["_align"], "formats": [np.dtype(f"u{RAND_STATE_ALIGNMENT}")], "itemsize": RAND_STATE_SIZE}, align=True
 )
--- a/src/katgpucbf/curand_helpers.py
+++ b/src/katgpucbf/curand_helpers.py
@@ -51,7 +51,7 @@ class RandomStateBuilder:
     def __init__(self, context: AbstractContext) -> None:
         with resources.as_file(resources.files(__package__)) as resource_dir:
             program = accel.build(context, "kernels/curand_init.mako", extra_dirs=[str(resource_dir)])
-        self._init_kernel = program.get_kernel("init_randState_t")
+        self._init_kernel = program.get_kernel("rand_state_init")
 
     def make_states(
         self,
--- a/src/katgpucbf/curand_helpers.py
+++ b/src/katgpucbf/curand_helpers.py
@@ -16,7 +16,7 @@
 
 """Helpers to initialise random state with curand.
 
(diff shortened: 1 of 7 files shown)


src/katgpucbf/kernels/curand_helpers.mako
```diff
@@ -0,0 +1,68 @@
...
+    uint2 raw[3];
+} randState_t;
+
+// Check that CUDA hasn't changed the internals of curandStateXORWOW_t
+static_assert(offsetof(curandStateXORWOW_t, d) == 0);
+static_assert(std::is_same_v<decltype(std::declval<curandStateXORWOW_t>().d), unsigned int>);
+static_assert(offsetof(curandStateXORWOW_t, v) == 4);
+static_assert(std::is_same_v<decltype(std::declval<curandStateXORWOW_t>().v), unsigned int[5]>);
+
+DEVICE_FN static inline void rand_state_load(curandStateXORWOW_t * RESTRICT out, const GLOBAL randState_t * RESTRICT in)
```
amishatishpatel, line 44: A question about your choice of function names here - `rand_state_load` and `rand_state_save`.
* Is there some kind of generally accepted convention between kernel syntax and `DEVICE_FN` syntax?
* `curand_init.mako` implements an `init_randState_t` kernel, with the actual/unedited data type in the name

src/katgpucbf/kernels/curand_helpers.mako
```diff
@@ -0,0 +1,68 @@
...
+extern "C++"  // PyCUDA wraps the whole file in extern "C"
+{
+#include <curand_kernel.h>
+#include <type_traits>
+#include <utility>
+}
+
+/* Holds the same random state as curandStateXORWOW_t, excluding the state for
+ * the Box-Muller transform. It uses an array of uint2 rather than uint so
+ * that global memory instructions use 64-bit rather than 32-bit access.
```
amishatishpatel, line 28: Aha, this is where the magic sauce is used to reduce memory traffic - moving to `uint2`.

amishatishpatel (commented):
A couple comments before explicitly approving. One day I'll be more in tune with the dark arts to give more targeted feedback. Otherwise, seems sound enough.

Very cool.

src/katgpucbf/xbgpu/kernels/beamform.mako
```diff
@@ -96,7 +96,8 @@ KERNEL REQD_WORK_GROUP_SIZE(BLOCK_SPECTRA, 1, 1) void beamform(
     rand_states += (batch * get_num_groups(1) + channel) * n_spectra + spectrum;
 
     curandStateXORWOW_t rand_state;
-    rand_state_load(&rand_state, rand_states);
+    if (valid)
+        rand_state_load(&rand_state, rand_states);
```
amishatishpatel, line 100: Aha, good catch.

amishatishpatel:
> The big gain comes when there are few antennas. With many antennas, the kernel running time is dominated by loading and processing the input data, and the cost of dithering the output is amortised; there is still about a 2% gain though.

Noted.

> @amishatishpatel I added documentation to the Python module, but let you know if you think it would be more appropriate to add a section to the doc/ directory.

I think it would be nice to add/move this documentation discussion (in `curand_helpers.py`) to e.g. [Beamforming::Dithering](https://katgpucbf.readthedocs.io/en/latest/xbgpu.design.b.html#dithering) in `doc/`. 
* Largely because it is quite a specific design choice that would fit well there.
* That being said, does it need a brief mention in the `.py` file?

src/katgpucbf/curand_helpers.py
```diff
@@ -14,51 +14,67 @@
...
+Code using this module should thus NOT generate Gaussian distributions without
+understanding the implications.
+"""
+
+from importlib import resources
 
 import numpy as np
-from katsdpsigproc.accel import DeviceArray, roundup
-from katsdpsigproc.cuda import Context
+from katsdpsigproc import accel
```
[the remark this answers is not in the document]
bmerry: I just wanted to get off the slippery slope of this import getting out of hand. Everywhere else in `src/` I import the module itself (I see there are a couple of tests that import from accel).

src/katgpucbf/kernels/curand_helpers.mako
bmerry: No actually - using uint2 is to make the memory traffic that exists require fewer instructions; I expect that to make a very difference if any. The reduction comes from using a smaller structure. Here's curandStateXORWOW_t:
```c++
struct curandStateXORWOW {
    unsigned int d, v[5];
    int boxmuller_flag;
    int boxmuller_flag_double;
    float boxmuller_extra;
    double boxmuller_extra_double;
};
```
We're jettisoning all the boxmuller fields since they're not needed for uniform random distributions.

amishatishpatel (approved):
Fantastic.

PR #827, opened 2024-07-11
Replace --spectra-per-heap with --jones-per-batch in scratch scripts

bmerry:
While the primary executables now use `--jones-per-batch` (or in the case of fgpu, a parameter to `--wideband/--narrowband`) and compute spectra-per-heap from it, some of the scratch scripts still used `--spectra-per-heap` directly with a default of 256. That means that unless it was overridden, we'd be benchmarking something different from what we run in production. The scripts now take `--jones-per-batch` with a default that matches the default of the primary executables.

I also had to address some bit-rot in compare_fengs.py.

Closes NGC-1392.

### DIFF
--- a/src/katgpucbf/xbgpu/main.py
+++ b/src/katgpucbf/xbgpu/main.py
@@ -245,7 +245,7 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         "--channels",
         type=int,
         required=True,
-        help="Total number of channels out of the F-Engine PFB.",
+        help="Total number of channels in the stream.",
     )
     parser.add_argument(
         "--channels-per-substream",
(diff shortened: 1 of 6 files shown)


bmerry:
(the branch was named with the incorrect ticket)

src/katgpucbf/xbgpu/main.py
```diff
@@ -245,7 +245,7 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         "--channels",
         type=int,
         required=True,
-        help="Total number of channels out of the F-Engine PFB.",
+        help="Total number of channels in the stream.",
```
amishatishpatel, line 248: That's a nice touch. Much appreciated.

amishatishpatel (approved):
It's very handy that you recently updated Jenkins to run the scratch scripts as well. I can't see anything too suspicious with the change. Please and thank you for catching those outstanding `_per_stream` -> `_per_substream`.

PR #833, opened 2024-07-15
Attempt to fix a sporadic dsim test failure

bmerry:
I'm not able to reproduce the failure on my laptop, which makes testing this fix difficult. However, it shouldn't break anything.

Maybe closes NGC-1415.

### DIFF
--- a/test/dsim/test_send.py
+++ b/test/dsim/test_send.py
@@ -126,6 +126,10 @@ async def test_sender(
     # Now proceed with DSim data using received descriptors (in ItemGroup)(ig)
     with PromDiff(namespace=send.METRIC_NAMESPACE) as prom_diff:
         await sender.run(0, TimeConverter(time.time(), ADC_SAMPLE_RATE))
+        # sender.run can return with some future callbacks (which update
+        # Prometheus counters) still scheduled for the next event loop
+        # iteration. Ensure they get a chance to run before exiting PromDiff.
+        await asyncio.sleep(0)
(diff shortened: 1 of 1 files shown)


test/dsim/test_send.py
```diff
@@ -126,6 +126,10 @@ def wrapped_send_heaps(*args, **kwargs):
     # Now proceed with DSim data using received descriptors (in ItemGroup)(ig)
     with PromDiff(namespace=send.METRIC_NAMESPACE) as prom_diff:
         await sender.run(0, TimeConverter(time.time(), ADC_SAMPLE_RATE))
+        # sender.run can return with some future callbacks (which update
+        # Prometheus counters) still scheduled for the next event loop
+        # iteration. Ensure they get a chance to run before exiting PromDiff.
+        await asyncio.sleep(0)
```
amishatishpatel, line 132: Noted. My mind immediately went to "Is this some kind of `noop`?"

Still, neat.

amishatishpatel (approved):

test/dsim/test_send.py
bmerry: > Noted. My mind immediately went to "Is this some kind of `noop`?"

I do think Python should have a function specifically for this, but here's what the docs for asyncio.sleep say:

> Setting the delay to 0 provides an optimized path to allow other tasks to run. This can be used by long-running functions to avoid blocking the event loop for the full duration of the function call.

In this case, it allows pending callbacks to run.

PR #837, opened 2024-08-12
Bump aiohttp to the latest version

bmerry:
To keep dependabot happy.


james-smith-za (approved):
Dependency creep. Not entirely making me happy but I don't have a good alternative proposal at the moment.

PR #841, opened 2024-08-14
fgpu: add a knob to allow dithering to be disabled

bmerry:
The --wideband and --narrowband options can now be passed dither=false to disable dithering.

Contributes to NGC-1429.

### DIFF
--- a/src/katgpucbf/fgpu/main.py
+++ b/src/katgpucbf/fgpu/main.py
@@ -79,6 +79,7 @@ class _OutputDict(TypedDict, total=False):
     dst: list[Endpoint]
     taps: int
     w_cutoff: float
+    dither: bool
 
 
 class _WidebandOutputDict(_OutputDict, total=False):
@@ -111,6 +112,7 @@ def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str]
     `kws`, the key and the value. If it does not recognise the key, it should
     raise ValueError.
     """
+    bool_map = {"false": False, "true": True}
     for part in value.split(","):
         match part.split("=", 1):
             case [key, data]:
@@ -125,10 +127,17 @@ def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str]
                         kws[key] = float(data)
                     case "dst":
                         kws[key] = endpoint_list_parser(DEFAULT_PORT)(data)
+                    case "dither":
+                        try:
+                            kws[key] = bool_map[data.lower()]
+                        except KeyError:
+                            raise ValueError("dither must be set to 'true' or 'false'") from None
                     case _:
                         field_callback(kws, key, data)
             case _:
                 raise ValueError(f"missing '=' in {part}")
+    # ignore due to https://github.com/python/mypy/issues/17674
+    kws.setdefault("dither", True)  # type: ignore
     for key in ["name", "channels", "dst"]:
         if key not in kws:
             raise ValueError(f"{key} is missing")
@@ -194,7 +203,6 @@ def parse_narrowband(value: str) -> NarrowbandOutput:
                 raise ValueError(f"{key} is missing")
         # Note that using **kws at the end means these are only defaults which
         # can be overridden by the user.
-        # The ignore is to work around https://github.com/python/mypy/issues/9408
         kws = {
             "taps": DEFAULT_TAPS,
             "w_cutoff": DEFAULT_W_CUTOFF,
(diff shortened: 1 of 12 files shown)


src/katgpucbf/fgpu/main.py
```diff
@@ -125,10 +127,17 @@ def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str]
                         kws[key] = float(data)
                     case "dst":
                         kws[key] = endpoint_list_parser(DEFAULT_PORT)(data)
+                    case "dither":
+                        try:
+                            kws[key] = bool_map[data.lower()]
+                        except KeyError:
+                            raise ValueError("dither must be set to 'true' or 'false'") from None
```
amishatishpatel, line 134: Must admit, maybe the first time I'm seeing `raise ... from ...`
* I gather it's because you want to treat the `KeyError` as a `ValueError`, actually
* You do give it a default on Line 140, but this is more to ensure the user gives a proper/explicit instruction?

amishatishpatel (approved):
Just the one comment, otherwise good to go. Just one higher-level question Re: katsdpcontroller update (to use `dither` option)
* It seems this PR can be merged without a katsdpcontroller update just yet
* Would you prefer they were merged together?

src/katgpucbf/fgpu/main.py
bmerry: `raise ... from None` suppresses the KeyError in the traceback. Without it, Python would tell you about the KeyError (with traceback), then it would say that while processing the KeyError, another exception occurred (the ValueError) and you get that traceback.

Overall this is to ensure you get a helpful message if you try to set it to something that is neither true nor false.

bmerry:
@amish When I got to the katsdpcontroller update I decided to future-proof it a bit by making the dither options `uniform` and `none` rather than `true` and `false`. That way we could add alternative dithering algorithms in future if there was demand. I'm thinking it would probably make sense to change this PR to do the same, so that katsdpcontroller can just pass the value straight through, and katgpucbf is similarly future-proofed. What do you think?

src/katgpucbf/fgpu/main.py
amishatishpatel: Indeed indeed, you worded it a lot better. Still, really neat.

amishatishpatel:
> @amish When I got to the katsdpcontroller update I decided to future-proof it a bit by making the dither options `uniform` and `none` rather than `true` and `false`. That way we could add alternative dithering algorithms in future if there was demand. I'm thinking it would probably make sense to change this PR to do the same, so that katsdpcontroller can just pass the value straight through, and katgpucbf is similarly future-proofed. What do you think?

Hmmmm, I saw that. Of course, excellent idea all around. However, this does sound like the update here won't be as simple as accepting a boolean `dither` flag.
* You might already have a complete idea of how to change it to be more parameterisable/scalable, but
* Do you "just need to tweak" e.g. `main.py::parse_stream` to check if `dither in ["uniform", "none"]` and all subsequent uses?

I like that the logical progression of this will be contained in this PR, so I'm absolutely for this change. I'll keep an eye out for the removal of the remapping in https://github.com/ska-sa/katsdpcontroller/pull/760

bmerry:
> > @amish When I got to the katsdpcontroller update I decided to future-proof it a bit by making the dither options `uniform` and `none` rather than `true` and `false`. That way we could add alternative dithering algorithms in future if there was demand. I'm thinking it would probably make sense to change this PR to do the same, so that katsdpcontroller can just pass the value straight through, and katgpucbf is similarly future-proofed. What do you think?
> 
> Hmmmm, I saw that. Of course, excellent idea all around. However, this does sound like the update here won't be as simple as accepting a boolean `dither` flag.
> 
>     * You might already have a complete idea of how to change it to be more parameterisable/scalable, but
> 
>     * Do you "just need to tweak" e.g. `main.py::parse_stream` to check if `dither in ["uniform", "none"]` and all subsequent uses?
> 
> 
> I like that the logical progression of this will be contained in this PR, so I'm absolutely for this change. I'll keep an eye out for the removal of the remapping in [ska-sa/katsdpcontroller#760](https://github.com/ska-sa/katsdpcontroller/pull/760)

I've pushed an update for this PR. Will update the katsdpcontroller PR next.

amishatishpatel (commented):
Just one comment (that I could genuinely look up/debug myself). A more pressing comment for clarity on the katsdpcontroller PR. Otherwise, this is good to go.

Don't want to approve just yet in case my question on the other PR influences this one.

test/fgpu/test_compute.py
```diff
@@ -20,13 +20,14 @@
...
 from katgpucbf.fgpu.engine import generate_ddc_weights
+from katgpucbf.utils import DitherType
 
 pytestmark = [pytest.mark.cuda_only]
 
 
 @pytest.mark.parametrize("mode", ["wideband", "narrowband"])
-@pytest.mark.parametrize("dither", [True, False])
-def test_compute(context: AbstractContext, command_queue: AbstractCommandQueue, mode: str, dither: bool) -> None:
+@pytest.mark.parametrize("dither", DitherType)
```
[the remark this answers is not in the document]
bmerry: It's nothing special in pytest. Enum classes implement the iterator protocol, so you can iterate over them to get all their members. That's not true for classes in general, so it won't work for `bool`.

PR #842, opened 2024-08-14
Reduce xbgpu startup GPU mem usage

bmerry:
It calls ensure_all_bound (via zero_visibilities) on the Correlation op early in gpu_proc_loop, before binding any input item. This causes a temporary input buffer to be created, which will subsequently be garbage collected.

Fix by having zero_visibilities only ensure_bound the required buffer.

Possibly contributes to NGC-938.

### DIFF
--- a/src/katgpucbf/xbgpu/correlation.py
+++ b/src/katgpucbf/xbgpu/correlation.py
@@ -292,7 +292,7 @@ class Correlation(accel.Operation):
 
     def zero_visibilities(self) -> None:
         """Zero all the values in the internal buffer."""
-        self.ensure_all_bound()
+        self.ensure_bound("mid_visibilities")
         self.buffer("mid_visibilities").zero(self.command_queue)
 
     @staticmethod


src/katgpucbf/xbgpu/correlation.py
```diff
@@ -292,7 +292,7 @@ def reduce(self) -> None:
 
     def zero_visibilities(self) -> None:
         """Zero all the values in the internal buffer."""
-        self.ensure_all_bound()
```
amishatishpatel, line 295: Why did we do `ensure_all_bound` in the first place? Was it just easier? Slash, we didn't really envision being in this situation?

amishatishpatel (approved):

src/katgpucbf/xbgpu/correlation.py
bmerry: I think `ensure_bound` was only added to katsdpsigproc later, to address this sort of situation.

PR #843, opened 2024-08-18
Qual: speed up checking whether chunks are complete

bmerry:
For baseline-correlation-products chunks, it was using a particularly slow way to check if any element equals -2^31. Speed it up by only checking one channel per heap, since this flagging is channel-independent within each engine.

This seems to fix the sporadic failures of test_accum_length, and in general should make the tests more reliable.

Contributes to NGC-938.

### DIFF
--- a/qualification/recv.py
+++ b/qualification/recv.py
@@ -92,6 +92,10 @@ class XBReceiver:
         self.cbf = cbf
         self._acv_name = acv_name
 
+    def is_complete_chunk(self, chunk: katgpucbf.recv.Chunk) -> bool:
+        """Check whether this chunk is complete (no missing data)."""
+        return bool(np.all(chunk.present))
+
     # The overloads ensure that when all_timestamps is known to be False, the
     # returned chunks are inferred to not be optional.
     @overload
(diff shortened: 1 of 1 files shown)


qualification/recv.py
```diff
@@ -157,12 +161,8 @@ async def complete_chunks(
                     timestamp = chunk.chunk_id * self.timestamp_step
                     if min_timestamp is not None and timestamp < min_timestamp:
                         logger.debug("Skipping chunk with timestamp %d (< %d)", timestamp, min_timestamp)
-                    elif not np.all(chunk.present):
+                    elif not self.is_complete_chunk(chunk):
                         logger.debug("Incomplete chunk %d", chunk.chunk_id)
-                    elif (chunk.data.dtype == np.dtype(np.int32) and np.any(chunk.data == -(2**31))) or (
-                        chunk.extra is not None and np.min(chunk.extra) < self.n_ants
```
amishatishpatel, line 163: I know this was a slightly different check to that now on Line 358, *but*
- I see this had a `np.min(chunk.extra) < self.n_ants`, and
- Now it's `== self.n_ants`

Was the check not entirely correct to begin with? Or has something changed in how we(/you) make use of `chunk.extra` now?

qualification/recv.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, cbf: CBFRemoteControl, stream_names: Sequence[str]) -> None:
         self.cbf = cbf
         self._acv_name = acv_name
 
+    def is_complete_chunk(self, chunk: katgpucbf.recv.Chunk) -> bool:
+        """Check whether this chunk is complete (no missing data)."""
+        return bool(np.all(chunk.present))
```
amishatishpatel, line 97: I gather casting it to a plain Python bool has done wonders for speeding up the check. Interesting indeed.

amishatishpatel (approved):
Let me not hold up the merge of this tweak any longer. Good catch indeed.

qualification/recv.py
bmerry: It was correct in the original code. But the original code is a check for an *incomplete* chunk. The new code is a check for a *complete* chunk, so the test needed to be inverted. Technically inverting it would be `>= self.n_ants` but it can't be more than `n_ants` (and if it is, we probably want to reject it anyway).

qualification/recv.py
bmerry: No, that cast is just to keep mypy happy. The big speedup is in the BaselineCorrelationProductsReceiver.is_complete_chunk overload where I now only check the first channel in each heap.

qualification/recv.py
amishatishpatel: Noted and noted.

qualification/recv.py
amishatishpatel: Aha, I realised this after reading your updates on the ticket. Really, excellent.

PR #844, opened 2024-08-18
Qualification: pin threads to cores

bmerry:
This should make the performance of qualification tests a little more predictable by always running the same tasks on the same cores.

Possibly helps with NGC-938.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -477,7 +477,7 @@ class CoreAllocator:
 
     It is initialised with a list of cores (from pytest config), with earlier
     entries considered better than later ones. Cores are allocated in this
-    order. There is no mechanism to return cores; simplify create a new
+    order. There is no mechanism to return cores; simply create a new
     allocator to start fresh.
     """
 
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -21,10 +21,11 @@ import asyncio
 import copy
 import inspect
 import logging
+import os
 import subprocess
 import time
-from collections import namedtuple
-from collections.abc import AsyncGenerator, Generator
+from collections import deque, namedtuple
+from collections.abc import AsyncGenerator, Generator, Iterable, Sequence
 
 import matplotlib.style
 import pytest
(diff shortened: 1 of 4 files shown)


bmerry:
I've marked this draft for now because there is still an open issue: the code for deciding which cores are available when not are explicitly specified doesn't work right. It works for the first test, but then for subsequent tests the `sched_getaffinity` will return the `sched_setaffinity` used for the previous test. We need a session-scoped fixture for the available cores.

qualification/conftest.py
```diff
@@ -470,9 +472,56 @@ async def cbf(
             await pcc.request("capture-stop", name)
 
 
+class CoreAllocator:
+    """Provide CPU cores to receivers that need them.
+
+    It is initialised with a list of cores (from pytest config), with earlier
+    entries considered better than later ones. Cores are allocated in this
+    order. There is no mechanism to return cores; simplify create a new
+    allocator to start fresh.
```
amishatishpatel, line 481: *simply create a new allocator to start fresh.

qualification/pytest-jenkins.ini
```diff
@@ -8,6 +8,8 @@ product_name = jenkins_qualification_cbf
 interface = enp193s0np0
 interface_gbps = 150
 use_ibv = true
+# This core ordering distributes load across the CCDs
+cores = 0 4 8 12 1 5 9 13 2 6 10 14 3 7 11 15
```
amishatishpatel, line 12: For anyone else coming back to this PR, discussing with Bruce on the ordering and he informs me that
1. It is a 'custom list', but by design (as per the comment), and
2. It's quite specific to our/a 16-core CPU.

This is, of course, for our own internal machines, so it's not like we need to accommodate different CPU core counts (just yet).

amishatishpatel (approved):
As  always, pleasantly annoyed by how neat your solution is. Also, certainly appreciate the comments/notes/explainers dotted about.

qualification/conftest.py
bmerry: Thanks for catching.

qualification/pytest-jenkins.ini
bmerry: I've added a comment that it's for an Epyc 7313P.

PR #845, opened 2024-08-19
Qual: subscribe only to initial sensors

bmerry:
This avoids being subjected to the firehose of sensor updates, which should make the test more reliable. Most of the sensors we query are static anyway. For the few exceptions we just query the product controller as needed.

Possibly contributes to NGC-938.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -205,8 +205,8 @@ async def test_delay_application_rate(cbf: CBFRemoteControl, pdf_report: Reporte
     enough.
     """
     pdf_report.step("Query rate of spectra.")
-    n_samples_between_spectra = cbf.sensors["antenna-channelised-voltage.n-samples-between-spectra"].value
-    scale_factor_timestamp = cbf.sensors["antenna-channelised-voltage.scale-factor-timestamp"].value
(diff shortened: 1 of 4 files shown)


qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -240,23 +241,23 @@ async def test_delay_sensors(
         delay_strs.append(f"{delay},{delay_rate}:{phase},{phase_rate}")
         delay_tuples.append((load_ts, delay, delay_rate, phase, phase_rate))
 
-    def delay_sensor_value(label: str) -> tuple:
-        return literal_eval(cbf.sensors[f"antenna-channelised-voltage.{label}.delay"].value.decode())
+    async def delay_sensor_value(label: str) -> tuple:
+        return literal_eval(await pcc.sensor_value(f"antenna-channelised-voltage.{label}.delay", str))
```
amishatishpatel, line 245: Good call (why was this blocking anyway?).

amishatishpatel (approved):
Brilliant, 10/10 - IGN

qualification/antenna_channelised_voltage/test_delay.py
bmerry: Previously the sensor watcher pushed updates into the internal mirror of the sensors - so when you wanted the latest value you just looked at what was in the mirror. Switching from a push model to a pull model means this has to be asynchronous now because there is a request/reply cycle.

PR #847, opened 2024-08-21
Add dither option to xbgpu

bmerry:
The kernel code is more complex than for fgpu, because one kernel
handles multiple beams, and they could have individual choices of
dithering. This is already handled for the polarisations, so we use the
same pattern: the choice is baked into the BeamformTemplate, and encoded
into the kernel source as a constant array generated at compile time.
The indices into the array should also be known to the compiler
(assuming it unrolls loops well) and hence it shouldn't add any cost.
In addition to this mechanism, if none of the beams have dithering, we
remove the random state array entirely.

Closes NGC-1429.

### DIFF
--- a/src/katgpucbf/xbgpu/beamform.py
+++ b/src/katgpucbf/xbgpu/beamform.py
@@ -17,6 +17,7 @@
 """Implement the calculations for beamforming."""
 
 from collections.abc import Sequence
+from dataclasses import dataclass
 from importlib import resources
 
 import numpy as np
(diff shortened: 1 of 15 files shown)


bmerry:
It looks like there is a bug somewhere in the kernel, because the unit tests are failing (but only with `--all-combinations`, which is why I didn't spot it earlier). So it might be best to start reviewing with the other parts of the code while I debug.

bmerry:
> It looks like there is a bug somewhere in the kernel, because the unit tests are failing (but only with --all-combinations, which is why I didn't spot it earlier). So it might be best to start reviewing with the other parts of the code while I debug.

The tolerance (was which just a thumb-suck anyway) was a little too tight. I'm slightly surprised that the errors are that big since it's all floating point, but there may be some catastrophic cancellation. It's still only about 1.6e-05 relative to the saturation point of 127.

src/katgpucbf/xbgpu/beamform.py
```diff
@@ -147,22 +169,27 @@ def __init__(
...
+                (n_batches, n_channels_per_substream, n_spectra_per_batch),
+                seed=seed,
+                sequence_first=sequence_first,
+                sequence_step=sequence_step,
+            )
+            self.bind(rand_states=rand_states)
+
+    @property
+    def _dither_enabled(self) -> bool:
+        return any(beam.dither != DitherType.NONE for beam in self.template.beams)
```
amishatishpatel, line 192: I know it absolutely makes sense for this to be here (in `Beamform`), but
* considering the exact same incantation is in `beamform.mako`,
* would you be opposed to making this a property of `BeamformTemplate`?
  * and passing the `any_dither` bool into `accel.build`.
* It would also take the arglist back down to two parameters (if that matters).
  * You wouldn't need to pass `DitherType` anymore.

amishatishpatel (approved):
One small observation/suggestion. Otherwise good to go. Quite glad I didn't put my hand up to make this change; I would not have got to such an elegant solution so quickly.

Worth asking, since your observation on tolerances for the unit test
* Do we need to tweak anything qualification-related?
* Or will we only find out once it's merged?

amishatishpatel (approved):

PR #848, opened 2024-08-21
Fix documentation of type of rand_states slot

bmerry:
The actual type was changed in #825, but this bit of documentation was missed.


amishatishpatel (approved):
Aha, nice one.

PR #849, opened 2024-08-22
Qualification: use multiple cores for baseline-correlation-products

bmerry:
This should hopefully reduce the number of dropped heaps.

See NGC-938.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -531,7 +531,7 @@ async def receive_baseline_correlation_products(
     receiver = BaselineCorrelationProductsReceiver(
         cbf=cbf,
         stream_name="baseline-correlation-products",
-        core=core_allocator.allocate(1)[0],
+        cores=core_allocator.allocate(4),
         interface_address=interface_address,
         use_ibv=use_ibv,
     )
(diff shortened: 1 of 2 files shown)


qualification/conftest.py
```diff
@@ -531,7 +531,7 @@ async def receive_baseline_correlation_products(
     receiver = BaselineCorrelationProductsReceiver(
         cbf=cbf,
         stream_name="baseline-correlation-products",
-        core=core_allocator.allocate(1)[0],
+        cores=core_allocator.allocate(4),
```
amishatishpatel, line 534: I gather this is more because how the cores are 'grouped' (to be spread across CCDs)
* i.e. Grab a whole group of four anyway,
* Rather than "we know we will always need/use four" (?)

james-smith-za (approved):
Moar cores. I like it.

amishatishpatel (approved):
I think I worded my question/statement poorly. Still, LGTM and see how it goes overnight.

qualification/conftest.py
bmerry: I didn't particularly think about the CCDs. I wanted a power of 2 (to evenly split the multicast groups), and I was worried that the gains from going from 1 to 2 might be wiped out by the overheads of the cores coordinating with each other, so I went for 4.

Additionally, katsdpcontroller always makes at least 4 XB-engines, so we're guaranteed to have something for each core to do.

PR #853, opened 2024-08-30
Add more information when general/control qual test fails

bmerry:
This should shed some light on whether it just needs more time for transmission to stabilise or is a more serious issue.

See NGC-1265.


bmerry:
I don't have a qualification test to attach because I haven't reproduced the issue yet.

james-smith-za (approved):
Looks fine, but this will only tell you about the last one. So if there are 2 or 5 or 28, you'll get no info as to whether they're all bunched up somewhere or spread over the time between 0 and N.

If that's what you want then it's fine I guess?

bmerry:
> If that's what you want then it's fine I guess?

For now, yes. We know that reception is unreliable at the start, and we're already throwing away the first 2s. This will tell us whether we just need to throw away a little more (e.g. 5s) or if they're spread out over the whole 30s.

PR #854, opened 2024-08-30
Bring the README up to date

bmerry:
I noticed it was rather outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,8 @@
 # Karoo Array Telescope GPU-Accelerated Correlator/Beamformer (katgpucbf)
 
-This repository implements both F- and X-engines for a GPU-based correlator,
-developed for the MeerKAT Extension Radio Telescope by the South African Radio
-Astronomy Observatory (SARAO). A B-engine for beamforming is planned, but not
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,8 @@
 # Karoo Array Telescope GPU-Accelerated Correlator/Beamformer (katgpucbf)
 
-This repository implements both F- and X-engines for a GPU-based correlator,
-developed for the MeerKAT Extension Radio Telescope by the South African Radio
-Astronomy Observatory (SARAO). A B-engine for beamforming is planned, but not
-yet included.
+This repository implements both F-, X-, and B-engines for a GPU-based correlator,
```
james-smith-za, line 3: My only nit-pick is that "both" applies when you list only two things (as far as I understand it anyway).

```suggestion
This repository implements F-, X-, and B-engines for a GPU-based correlator,
```

james-smith-za (approved):

README.md
bmerry: Thanks, I'd missed that when I added B-engines.

PR #857, opened 2024-09-04
Improvements to test_control

bmerry:
This should improve (but probably not fix) the issues with this test (NGC-1265).

- Use real-time scheduling for the qualification test
- Set the gains at the start of the test, for its side-effect of updating steady-state-timestamp, and hence ensuring we flush out any data that was already in the pipeline rather than using it.
- Remove STARTUP_TIME, which is no longer needed due to the above.
- Increase TEST_TIME_TOL, which was too low after the above.

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/user-attachments/files/16869311/report.pdf)


qualification/general/test_control.py
```diff
@@ -236,6 +227,10 @@ async def test_control(
     pdf_report.detail(f"Subscribed to {len(sensor_watcher.sensors)} sensors.")
 
     pdf_report.step("Set up periodic control.")
+    # We don't actually care about the gains, but setting it updates the
+    # steady-state timestamp, and ensures that consume_chunks sees only
+    # "fresh" data.
+    await cbf.product_controller_client.request("gain-all", "antenna-channelised-voltage", "default")
```
james-smith-za, line 233: Might be unnecessary after NGC-1462?

james-smith-za (approved):

qualification/general/test_control.py
bmerry: No, that's unrelated. The fixture setup will listen to TACV heaps until it gets a good one, and during that time, BCP heaps will also be received and go into the ringbuffer. We want to ignore those.

PR #858, opened 2024-09-04
Qual: use np.ones to create chunks

bmerry:
It guarantees paged-in memory, which should reduce the amount of data lost once we actually subscribe to the streams.

Relates to NGC-1265 (but might have no effect).


james-smith-za (approved):
Isn't there a more explicit method of doing this?

We talked about using page-locked memory a while ago in a different context, does that concept not apply here?

PR #865, opened 2024-10-02
Attempt to make qual test start more robust

bmerry:
- Introduce DEFAULT_TIMEOUT to replace the hard-coded 10s timeout
- Have the receiver fixtures wait 3 * DEFAULT_TIMEOUT, to account for occasionally slow startup of a correlator
- Introduce wait_complete_chunk, which is more efficient that next_complete_chunk because it doesn't make and return a copy of the data.
- Improve docstring for next_complete_chunk.

### DIFF
--- a/qualification/recv.py
+++ b/qualification/recv.py
@@ -225,7 +225,9 @@ class XBReceiver:
     ) -> int:
         """Wait for a complete chunk, but do not return it.
 
-        Only the timestamp is returned.
+        Only the timestamp is returned. This is more efficient than
+        :meth:`next_complete_chunk` because it does not need to copy the data
+        from the chunk.
 
         Parameters
         ----------
--- a/qualification/recv.py
+++ b/qualification/recv.py
@@ -44,6 +44,7 @@ from katgpucbf.utils import TimeConverter
 
 from .cbf import DEFAULT_MAX_DELAY, CBFRemoteControl
 
+DEFAULT_TIMEOUT = 10.0
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


bmerry:
Haven't actually tested this yet. Will do so today.

qualification/recv.py
```diff
@@ -199,20 +200,60 @@ async def next_complete_chunk(
...
+    async def wait_complete_chunk(
+        self,
+        min_timestamp: int | None = None,
+        *,
+        max_delay: int = DEFAULT_MAX_DELAY,
+        timeout: float | None = DEFAULT_TIMEOUT,
+    ) -> int:
+        """Wait for a complete chunk, but do not return it.
+
+        Only the timestamp is returned.
```
amishatishpatel, line 228: Could I request a small update/bit of bikeshedding, considering both this and `next_complete_chunk` exist
* Perhaps a tweak/addition to this docstring as per your commit/PR blurb clarifying the difference between the two functions,
* Or even just why you/one would use one or the other

> [wait_complete_chunk] is more efficient that next_complete_chunk because it doesn't make and return a copy of the data.

amishatishpatel (approved):
One bit of bikeshedding/suggestion for improving clarity with the new function.

Otherwise, good to go.

qualification/recv.py
```diff
@@ -199,20 +200,60 @@ async def next_complete_chunk(
...
             See :meth:`complete_chunks`
+        timeout
+            Maximum time to wait
+
+        Raises
+        ------
+        TimeoutError
+            if a complete chunk is not received in time
+        RuntimeError
+            if the stream is stopped before a complete chunk is received
```
amishatishpatel, line 211: I will admit, I would've liked the first letter of these sentences to be capitalised (like other uses/implementations). Don't feel strongly enough to request a single commit just for that, but if you do enact my other docstring suggestion please do tend to this as well.

PR #870, opened 2024-10-15
Test the filter-group-delay sensor

bmerry:
This required something of a rewrite of the test to improve the sensitivity. See the relevant commit for more details. I'm not completely happy that the sensitivity of the test depends on the number of antennas used (`default_antennas` setting). A potential alternative would be to fix the number of channels per dsim, then set the number of seconds of data to collect in inverse proportion to the number of antennas, which means that the test run-time would be variable and the precision would be fixed, rather than vice versa.

Note that the dsim image used needs #869, and the product controller needs ska-sa/katsdpcontroller#773.

- [x] If qualification tests are changed: attach a sample qualification report: [report8.pdf](https://github.com/user-attachments/files/17374113/report8.pdf)

Closes NGC-1464.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -599,6 +599,12 @@ async def test_group_delay(
     cancels out when measuring phase differences, but ensures that there is
     variation in the phases and thus helps prevent systematic bias in the
     quantisation errors.
+
+    The test is replicated across many channels to increase the number of
+    samples and hence improve signal-to-noise. The channels are divided amongst
+    the simulated antennas then combined by beamforming, which allows for
+    larger amplitudes without saturation in the time domain, and also speeds up
+    the test.
     """
     receiver = receive_tied_array_channelised_voltage
     client = cbf.product_controller_client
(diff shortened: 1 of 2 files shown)


qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -635,13 +649,7 @@ async def test_group_delay(
...
     pdf_report.detail(f"Set delays to {delay_model} for all inputs.")
 
-    pdf_report.step("Set beamformer weights to use only one antenna.")
-    weights = np.zeros(len(receiver.source_indices[0]))
-    weights[0] = 1.0
-    await client.request("beam-weights", receiver.stream_names[0], *weights)
-    pdf_report.detail(f"Set weights on {receiver.stream_names[0]}.")
-
-    async def measure_once(freqs: tuple[float, float]) -> tuple[float, float, float]:
+    async def measure_once(rel_freqs: tuple[float, float]) -> tuple[float, float, float]:
```
james-smith-za, line 652: So this function's signature is now somewhat confusing and I had to go back and forth a few times to understand what it was doing. 

The way it was being used in the test had me thinking you were doing a test at DC and 12Hz which would obviously not give you meaningful data.

So I'd suggest changing it something like this:
```suggestion
    async def measure_once(frequency_delta: float) -> tuple[float, float, float]:
```

The docstring could probably also use a bit of an update.

qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -657,16 +665,25 @@ async def measure_once(freqs: tuple[float, float]) -> tuple[float, float, float]
...
         """
-        signal = f"cw({amplitude}, {freqs[0]}); cw({amplitude}, {freqs[1]});"
-        await cbf.dsim_clients[0].request("signals", signal, dsim_period)
-        pdf_report.detail(f"dsim signal set to {signal}.")
+        async with asyncio.TaskGroup() as tg:
+            for dsim_client, dsim_freqs in zip(cbf.dsim_clients, cfreqs_by_dsim):
+                if dsim_freqs:
+                    signal = " ".join(
+                        f"multicw({len(dsim_freqs)}, {amplitude}, 0.0, {dsim_freqs[0] + rfreq}, {freq_step_dsim});"
+                        for rfreq in rel_freqs
```
james-smith-za, line 673: Going along with my above comment:

```suggestion
                        f"multicw({len(dsim_freqs)}, {amplitude}, 0.0, {dsim_freqs[0] + rfreq}, {freq_step_dsim});"
                        for rfreq in [0, frequency_delta]
```

james-smith-za (commented):
I like the change overall but one part I think could use some change to facilitate the reader's understanding of things.

It's not a hill I'm willing to die on though so I could be persuaded otherwise by a good argument.

qualification/antenna_channelised_voltage/test_delay.py
bmerry: I'm not following. You've replaces two degrees of freedom (a 2-tuple) with one - how do I reconstruct the original values?

I agree that the docstring could be made more explicit.

qualification/antenna_channelised_voltage/test_delay.py
bmerry: That would change the second pass, which currently tests a symmetric range around the centre of the channel.

james-smith-za (approved):
Much better thanks

PR #873, opened 2024-10-18
Update testing of xbgpu sensor timestamps

amishatishpatel:
Specifically in the end-to-end test, check that requests issued before and during data
processing are enacted accordingly.

Closes NGC-1379.

### DIFF
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -584,7 +584,14 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
         self.send_stream.enable_beam(beam_id=stream_id, enable=enable, timestamp=timestamp)
 
     def _weights_updated(self) -> int:
-        """Update version tracking when weight-related parameters are updated."""
+        """Update version tracking when weight-related parameters are updated.
+
+        Returns
+        -------
+        int
+            :attr:`_weights_steady`, the timestamp that includes the effect of
+            a beam request update made now.
+        """
         self._weights_version += 1
         self.engine.update_steady_state_timestamp(self._weights_steady)
         return self._weights_steady
@@ -598,6 +605,12 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             The index of the beam whose weights are being set.
         weights
             A 1D array containing real-valued weights (per input).
+
+        Returns
+        -------
+        int
+            :attr:`_weights_steady`, the timestamp that includes the effect of
+            a beam-weights update made now.
         """
         self._weights[stream_id] = weights
         return self._weights_updated()
@@ -611,6 +624,12 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             The index of the beam whose quantisation gain is being set.
         gain
             Real-valued quantisation gain.
+
+        Returns
+        -------
+        int
+            :attr:`_weights_steady`, the timestamp that includes the effect of
+            a beam-quant-gain update made now.
         """
         self._quant_gains[stream_id] = gain
         return self._weights_updated()
@@ -627,6 +646,12 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             inputs. The second axis has length two, with the first element
             containing the delay in seconds, and the second containing a
             channel-independent phase rotation in radians.
+
+        Returns
+        -------
+        int
+            :attr:`_weights_steady`, the timestamp that includes the effect of
+            a beam-delay update made now.
         """
         self._delays[stream_id] = delays
         return self._weights_updated()
--- a/src/katgpucbf/xbgpu/engine.py
+++ b/src/katgpucbf/xbgpu/engine.py
@@ -583,12 +583,13 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
     def capture_enable(self, *, stream_id: int, enable: bool = True, timestamp: int = 0) -> None:  # noqa: D102
         self.send_stream.enable_beam(beam_id=stream_id, enable=enable, timestamp=timestamp)
 
-    def _weights_updated(self) -> None:
+    def _weights_updated(self) -> int:
         """Update version tracking when weight-related parameters are updated."""
         self._weights_version += 1
         self.engine.update_steady_state_timestamp(self._weights_steady)
+        return self._weights_steady
 
-    def set_weights(self, stream_id: int, weights: np.ndarray) -> None:
+    def set_weights(self, stream_id: int, weights: np.ndarray) -> int:
         """Set the beam weights for one beam.
 
         Parameters
@@ -599,9 +600,9 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             A 1D array containing real-valued weights (per input).
         """
         self._weights[stream_id] = weights
-        self._weights_updated()
+        return self._weights_updated()
 
-    def set_quant_gain(self, stream_id: int, gain: float) -> None:
+    def set_quant_gain(self, stream_id: int, gain: float) -> int:
         """Set the quantisation gain for one beam.
 
         Parameters
@@ -612,9 +613,9 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             Real-valued quantisation gain.
         """
         self._quant_gains[stream_id] = gain
-        self._weights_updated()
+        return self._weights_updated()
 
-    def set_delays(self, stream_id: int, delays: np.ndarray) -> None:
+    def set_delays(self, stream_id: int, delays: np.ndarray) -> int:
         """Set the beam steering delays for one beam.
 
         Parameters
@@ -628,7 +629,7 @@ class BPipeline(Pipeline[BOutput, BOutQueueItem]):
             channel-independent phase rotation in radians.
         """
         self._delays[stream_id] = delays
-        self._weights_updated()
+        return self._weights_updated()
 
 
 class XPipeline(Pipeline[XOutput, XOutQueueItem]):
@@ -1384,8 +1385,8 @@ class XBEngine(DeviceServer):
         pipeline, stream_id = self._request_bpipeline(stream_name)
         if len(weights) != self.n_ants:
             raise aiokatcp.FailReply(f"Incorrect number of weights (expected {self.n_ants}, received {len(weights)})")
-        pipeline.set_weights(stream_id, np.array(weights))
-        self.sensors[f"{stream_name}.weight"].set_value(str(list(weights)))
+        steady_state_timestamp = pipeline.set_weights(stream_id, np.array(weights))
+        self.sensors[f"{stream_name}.weight"].set_value(str(list(weights)), timestamp=steady_state_timestamp)
 
     async def request_beam_delays(self, ctx, stream_name: str, *delays: str) -> None:
         """Set the delays for all inputs of a given beam and update the sensor.
@@ -1408,9 +1409,12 @@ class XBEngine(DeviceServer):
             delay_str, phase_str = entry.split(":")
             new_delays[i, 0] = float(delay_str)
             new_delays[i, 1] = float(phase_str)
-        pipeline.set_delays(stream_id, new_delays)
+        steady_state_timestamp = pipeline.set_delays(stream_id, new_delays)
         delays_formatted_str = ", ".join(str(value) for value in new_delays.flatten())
-        self.sensors[f"{stream_name}.delay"].set_value(f"({pipeline._weights_steady}, {delays_formatted_str})")
+        self.sensors[f"{stream_name}.delay"].set_value(
+            f"({steady_state_timestamp}, {delays_formatted_str})",
+            timestamp=steady_state_timestamp,
+        )
 
     async def request_beam_quant_gains(self, ctx, stream_name: str, gain: float) -> None:
         """Set the quantisation gain for a beam and update the sensor.
@@ -1423,8 +1427,8 @@ class XBEngine(DeviceServer):
             The new gain to apply.
         """
         pipeline, stream_id = self._request_bpipeline(stream_name)
-        pipeline.set_quant_gain(stream_id, gain)
-        self.sensors[f"{stream_name}.quantiser-gain"].set_value(gain)
+        steady_state_timestamp = pipeline.set_quant_gain(stream_id, gain)
+        self.sensors[f"{stream_name}.quantiser-gain"].set_value(gain, timestamp=steady_state_timestamp)
 
     async def start(self, descriptor_interval_s: float = SPEAD_DESCRIPTOR_INTERVAL_S) -> None:
         """
--- a/test/xbgpu/test_engine.py
+++ b/test/xbgpu/test_engine.py
@@ -60,6 +60,9 @@ SEND_RATE_FACTOR: Final[float] = 1.1
 SAMPLE_BITWIDTH: Final[int] = 8
 N_TOTAL_XB_HEAPS: Final[int] = 70
 N_BEAM_REQUESTS: Final[int] = 2
+# The latest chunk index at which updated beam parameters can be updated
+# in order to be properly tested.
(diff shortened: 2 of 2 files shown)


src/katgpucbf/xbgpu/engine.py
```diff
@@ -583,12 +583,13 @@ async def sender_loop(self) -> None:  # noqa: D102
     def capture_enable(self, *, stream_id: int, enable: bool = True, timestamp: int = 0) -> None:  # noqa: D102
         self.send_stream.enable_beam(beam_id=stream_id, enable=enable, timestamp=timestamp)
 
-    def _weights_updated(self) -> None:
+    def _weights_updated(self) -> int:
```
bmerry, line 586: Update the docstring to explain what value is returned. Similarly for the other functions that are now returning a timestamp.

test/xbgpu/test_engine.py
```diff
@@ -1153,13 +1224,24 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
 
     def _patch_get_in_item(
-        self, monkeypatch: pytest.MonkeyPatch, count: int, client: aiokatcp.Client, *request
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        count: int,
+        client: aiokatcp.Client,
+        requests: list[tuple[str, ...]],
     ) -> list[int]:
         """Patch :meth:`~.BPipeline._get_in_item` to make a request partway through the stream.
```
bmerry, line 1233: This docstring needs to be updated to use *requests* (plural).

test/xbgpu/test_engine.py
```diff
@@ -1153,13 +1224,24 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
 
         The returned list will be populated with the value of the
         ``steady-state-timestamp`` sensor immediately after executing the
         request.
+
+        Parameters
+        ----------
+        count
+            Count point at which to execute katcp requests in `requests`.
```
bmerry, line 1242: "Count point" isn't a clear term. How about
```suggestion
            Index (counting from 1) of the call to `_get_in_item` prior to which the `requests` will be made.
```

test/xbgpu/test_engine.py
```diff
@@ -1153,13 +1224,24 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
             assert xbengine.sensors[f"{corrprod_output.name}.xeng-clip-cnt"].value == n_vis
 
     def _patch_get_in_item(
-        self, monkeypatch: pytest.MonkeyPatch, count: int, client: aiokatcp.Client, *request
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        count: int,
+        client: aiokatcp.Client,
+        requests: list[tuple[str, ...]],
```
bmerry, line 1231: From the implementation, it looks like it could more generically be an iterable of iterables.

test/xbgpu/test_engine.py
```diff
@@ -429,6 +431,63 @@ def verify_corrprod_sensors(
...
+        channel_spacing=channel_spacing,
+        centre_channel=centre_channel,
+    )
+    # assert_allclose converts to float, which bloats memory usage.
+    # To keep it manageable, compare a batch at a time.
+    for i in range(len(beam_outputs)):
+        for j in range(len(batch_indices)):
+            np.testing.assert_allclose(expected_beams[i, j], beam_results[i, j], atol=1)
+
+    return expected_beam_saturated_low, expected_beam_saturated_high
```
bmerry, line 488: Should probably document these return values.

test/xbgpu/test_engine.py
```diff
@@ -490,29 +549,39 @@ def verify_beam_sensors(
         assert saturated_low[i] <= stream_diff.diff("output_b_clipped_samples_total") <= saturated_high[i]
 
         # Check that sensor value matches Prometheus
+        # NOTE: Verifying the timestamp on saturation count updates is not as predictable
+        # as saturation relies on dithering. As a result, the timestamp field here is mocked
+        # out, and other sensor updates are verified more completely.
```
bmerry, line 554: I wouldn't use the term "mocked out" here - that normally implies replacing functional behaviour. I'd say the timestamp field is "ignored".

test/xbgpu/test_engine.py
```diff
@@ -984,6 +1053,22 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
         delays[..., 0] = rng.uniform(-1e-9, 1e-9, size=(len(beam_outputs), n_ants))
         # Phase is in radians
         delays[..., 1] = rng.uniform(-2 * np.pi, 2 * np.pi, size=(len(beam_outputs), n_ants))
+
+        katcp_requests: dict[str, list[tuple[str, ...]]] = {
```
bmerry, line 1057: The type hint `tuple[str, ...]` is wrong: it means a tuple containing only strings, whereas `weights` and `quant_gains` contain floats.

test/xbgpu/test_engine.py
```diff
@@ -984,6 +1053,22 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
+            output.name: [
+                ("beam-weights", output.name, *weights[i]),
+                ("beam-quant-gains", output.name, quant_gains[i]),
+                ("beam-delays", output.name, *[f"{d[0]}:{d[1]}" for d in delays[i]]),
+            ]
+            for i, output in enumerate(beam_outputs)
+        }
+        flattened_requests = list(chain.from_iterable(katcp_requests.values()))
```
bmerry, line 1065: Since you're only using the values of `katcp_requests` (and not the keys), why not make it a list of lists rather than a dict of lists?

test/xbgpu/test_engine.py
```diff
@@ -1086,7 +1157,7 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
             prom_diff=prom_diff,
             actual_sensor_updates=actual_sensor_updates,
             first_timestamp=first_timestamp,
-            last_timestamp=last_timestamp,
+            last_timestamp=steady_state_timestamps[-1],
```
bmerry, line 1160: This no longer matches the documentation for this parameter of `verify_beam_sensors`. It looks like you've hijacked this parameter and used it for a different purpose. You should probably change the parameter names and the documentation of them.

bmerry (changes requested):
Looks like a good start. The next step will be to have the mid-stream requests actually *change* the weights etc, and then update generate_expected_beams to take the changes into account. That will verify that the sensor timestamps actually match the data path, rather than just checking that sensor timestamps are consistent with the steady-state-timestamp sensor.

test/xbgpu/test_engine.py
amishatishpatel: Aha, I was under the impression the `...` type hint is similar to numpy indexing
* That is, "I'm not concerned with these dimensions"
* Slash, "They can be anything, as long as the first item/index is a string"

I'll figure something out. Was hesitant to use `typing.Any` as the requests don't all have the same number of args.

bmerry: I don't think the type system supports the idea of "first element is a string, and the remaining element can be anything". I think the best you'll manage is plain `tuple` with no specification of what's in the tuple.

amishatishpatel:
@bmerry I thought I could make some headway on a working example of what I was trying to discuss this morning; running out of mental steam on it. I was trying to ask this morning Re:
> The next step will be to have the mid-stream requests actually change the weights etc, and then update generate_expected_beams to take the changes into account.

I've currently implemented something that issues **two** sets of `?beam` requests during the test run
* Once at the start, 
* Again a bit later on in the data processing

In verifying the beam data
* I do a manual check `if current heap_id == beam-param change id` then scroll over to the next set of `?beam` params
* Should I expand that to allow for more `?beam` requests to be tested?

Let me know if that's unclear, happy to chat via Zoom. I'll continue at it for today and request another round of reviewing from you either way by the morning.

bmerry:
> Should I expand that to allow for more ?beam requests to be tested?

Nah, what you've described sounds like it'll verify that the sensors and the data agree.

amishatishpatel:
Jenkins build failures were due to documentation (again).

test/xbgpu/test_engine.py
```diff
@@ -244,7 +250,11 @@ def generate_expected_beams(
     accum = np.zeros((len(beam_pols), len(batch_indices), n_channels), np.complex64)
     sample = np.empty((N_POLS, COMPLEX), np.int8)
     sample_fp = np.empty(N_POLS, np.complex64)
+
+    beam_params_id = 0
     for batch_id, batch_index in enumerate(batch_indices):
+        if beam_params_change_batch_index == batch_id:
```
bmerry, line 256: Should `beam_params_change_batch_index` be renamed to `beam_params_change_batch_id`? It's a little confusing that it's being compared to `batch_id` and not `batch_index` given the name.

test/xbgpu/test_engine.py
```diff
@@ -457,22 +468,33 @@ def verify_beam_data(
...
+    beam_params_change_batch_index
+        Batch index in `batch_indices` where the beam weights, delays and
+        quant-gains changed.
     weights, quant_gains, delays
         The beam weights, quantiser-gains and delays applied to each input of
         the beam data product. These are real floating-point values generated
-        for the unit test.
+        for the unit test. There are `N_BEAM_REQUESTS` distinct sets of request
+        parameters in each ndarray. Starting at index 0, the next subsequent set
+        is applied at the indices specified in `beam_request_index`.
```
bmerry, line 479: This is the only reference to `beam_request_index`.

test/xbgpu/test_engine.py
```diff
@@ -523,9 +544,9 @@ def verify_beam_sensors(
         Dictionary of lists of sensor updates. They dictionary keys are sensor
         names, the values are a list of readings for each sensor update captured
         via the callback attached to :class:`BPipeline` sensors.
-    first_timestamp, last_timestamp
-        Two timestamps indicating the start and end of data processing
-        by the :class:`BPipeline`.
+    runtime_timestamps
```
bmerry, line 547: What does "runtime" mean in this context?

test/xbgpu/test_engine.py
```diff
@@ -1046,28 +1079,41 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
-        delays[..., 1] = rng.uniform(-2 * np.pi, 2 * np.pi, size=(len(beam_outputs), n_ants))
-
-        katcp_requests: dict[str, list[tuple[str, ...]]] = {
-            output.name: [
-                ("beam-weights", output.name, *weights[i]),
-                ("beam-quant-gains", output.name, quant_gains[i]),
-                ("beam-delays", output.name, *[f"{d[0]}:{d[1]}" for d in delays[i]]),
+        delays[..., 1] = rng.uniform(-2 * np.pi, 2 * np.pi, size=(N_BEAM_REQUESTS, len(beam_outputs), n_ants))
+
+        # NOTE: This list of lists is done in an order that has a set of requests for alternating
```
bmerry, line 1090: "done" is normally associated with a verb, but "list of lists" is a noun. Do you mean
```suggestion
        # NOTE: This list of lists is constructed in an order that has a set of requests for alternating
```

test/xbgpu/test_engine.py
```diff
@@ -1046,28 +1079,41 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
+        # triggered during the BPipeline's processing of data. We use the 'lower' subset for
+        # `?beam` requests before it starts processing data so as to make data verification a
+        # bit simpler.
         steady_state_timestamps = self._patch_get_in_item(
             monkeypatch,
-            count=10,
+            count=beam_params_change_index,
             client=client,
-            requests=flattened_requests,
+            requests=list(chain.from_iterable(katcp_requests[N_BEAM_REQUESTS:])),
```
bmerry, line 1116: This indexing doesn't look right - I think it should be `[len(beam_outputs):]`. But you'll probably make the code clearer if you add another level to katcp_requests, so that it's a list of list of lists of shape (N_BEAM_REQUESTS, len(beam_outputs), 3).

test/xbgpu/test_engine.py
```diff
@@ -1086,8 +1132,14 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
                 present[40:50, missing_antenna] = False  # Covers some complete accumulations
                 present[60, missing_antenna] = False  # Just one heap in an accumulation
 
-            for beam_request in flattened_requests:
+            for beam_request in list(chain.from_iterable(katcp_requests[:N_BEAM_REQUESTS])):
```
bmerry, line 1135:
```suggestion
            for beam_request in chain.from_iterable(katcp_requests[:N_BEAM_REQUESTS]):
```
You can iterate directly over `from_iterable` without collecting it into a list first.

bmerry: Again, the indexing looks fishy.

bmerry (changes requested):
Apart from the line-by-line changes, something I'd like to see (and maybe it's already there and I missed it) is a check that the change point is strictly before the end of the data, so that you're testing a non-zero number of heaps after the change.

test/xbgpu/test_engine.py
amishatishpatel: You're quite right, I was thinking the same and forgot to update before I pushed.

test/xbgpu/test_engine.py
amishatishpatel: Yeah I was trying to think of a better adjective because it's
* the timestamps 'during the test', and 
* not the timestamp at the instant the test begins

Perhaps I'll rename to `beam_request_timestamps` (?)

test/xbgpu/test_engine.py
amishatishpatel: Quite right once again.

amishatishpatel:
> [...] something I'd like to see [...] is a check that the change point is strictly before the end of the data, so that you're testing a non-zero number of heaps after the change.

You're quite right. I do outline that in a `NOTE` before `beam_params_change_index`, but an explicit assertion always helps.

amishatishpatel:
> > [...] something I'd like to see [...] is a check that the change point is strictly before the end of the data, so that you're testing a non-zero number of heaps after the change.
> 
> You're quite right. I do outline that in a `NOTE` before `beam_params_change_index`, but an explicit assertion always helps.

Rather interestingly, in testing that a `beam_params_change_index` happens 'too late' in the data
* It actually manifests as an `IndexError` in accessing `steady_state_timestamps` returned from `_patch_get_in_item`
  * That is, the `counter == count` branch never gets triggered, and
  * `timestamp` never gets appended with a new value (really, any value)

I've arbitrarily chosen the 'latest possible beam param change' as one chunk from the end (so there are at least `HEAPS_PER_FENGINE_PER_CHUNK` to verify newer values). You'll let me know if that needs to be brought forward during the next round of reviewing.

bmerry (approved):

amishatishpatel:
@bmerry I realised I should've updated the end-to-end test's docstring earlier. Pushed an update now, I'll leave it until the morning to merge in case you want to review that last commit.

test/xbgpu/test_engine.py
```diff
@@ -1028,6 +1028,10 @@ async def test_engine_end_to_end(
         non-contiguous batch indices (i.e. missing Chunks in the receiver), there
         is an artificial gap introduced in the data sent to/processed by the
         XBEngine.
+
+        Lastly, more complete testing of `?beam` requests is carried out by
```
bmerry, line 1032: "More" complete relative to what?

amishatishpatel: Relative to how it was being tested before. Although I see how looking at the test as-is doesn't really give context of how it was (not the point of it either).

I'll reword.

bmerry (approved):

amishatishpatel:
For the record, recent Jenkins build failures were (once again) in the documentation arm.

PR #875, opened 2024-10-21
dsim: speed up multicw

bmerry:
Implement it using an FFT (actually a DCT), instead of manually adding all the signals together. This significant reduces the time when there are a large number of CWs to add.

Relates to NGC-1464.


james-smith-za (approved):
Logic makes sense to me.

PR #879, opened 2024-10-23
Qual: increase timeout for tacv steady_state tests

bmerry:
The timeout was not quite long enough for the lowest-bandwidth configurations, where the pipeline becomes very deep (in seconds).

Closes NGC-1462


bmerry:
I haven't actually run this, but I figure it's safe enough that it can get run in a nightly test after merging.

PR #883, opened 2024-11-08
Add unit test to capture-stop some streams

amishatishpatel:
Randomly select data streams in the XB-engine to issue a `?capture-stop` request to at
some point during data processing. Verify that those streams stopped only have data
up to the `capture-stop` point, and other streams have processed and transmitted data
uninterrupted.

Closes NGC-1155.


test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1465,99 @@ def get_stream_status(stream_name: str) -> bool:
...
+        check the corresponding streams only have partial data transmission.
+        Also ensure that data is completely received for streams that did not
+        receive a `?capture-stop` request.
+        """
+        n_batches = heap_accumulation_threshold[0]
+        # Naive list concatention using '+' not used below as mypy was not happy
+        # lists of two different types could be added together.
+        all_stream_names = list(map(lambda output: output.name, chain(corrprod_outputs, beam_outputs)))
+        rng = np.random.default_rng(seed=1)
+        # Select random subset of output data products to issue `?capture-stop` to
```
james-smith-za, line 1497: The last part of the comment seems to have gotten lost somewhere...

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1465,99 @@ def get_stream_status(stream_name: str) -> bool:
...
+        # therefore still be zero for indices after the `?capture-stop` request
+        # was issued. That is, completely zero data is akin to no data received.
+        for stopped_stream in stopped_stream_names:
+            if stopped_stream in corrprod_results.keys():
+                # NOTE: As per the `DEFAULT_PARAMETERS`, both corrprod_outputs have the same
+                # heap_accumulation_threshold, which is why we can use either one to calculate
+                # the accumulation index where data stopped transmitting.
+                np.testing.assert_equal(
+                    corrprod_results[stopped_stream][
+                        capture_stop_chunk_id * HEAPS_PER_FENGINE_PER_CHUNK // heap_accumulation_threshold[0] :
```
bmerry, line 1540: This line wrapping makes this whole statement hard to read. I'd suggest pulling the calculation into a separate statement (assigning it to a variable).

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1465,99 @@ def get_stream_status(stream_name: str) -> bool:
...
+                # ceased transmitting for a stopped stream.
+                np.testing.assert_equal(
+                    beam_results[stopped_stream][capture_stop_chunk_id * HEAPS_PER_FENGINE_PER_CHUNK :], 0
+                )
+            # Remove this stopped-stream from the master list of data streams
+            all_stream_names.remove(stopped_stream)
+
+        for captured_stream in all_stream_names:
+            if captured_stream in corrprod_results.keys():
+                # NOTE: The `heap_factory` only sends real data. The results buffer has
```
bmerry, line 1557: Um, it looks to me like it's sending 10+10j. Perhaps you mean that the correlations that result are purely real?

test/xbgpu/test_engine.py
```diff
@@ -821,77 +818,91 @@ async def _send_data(
...
+                stream.add_inproc_reader(mock_send_stream[i + len(corrprod_outputs)])
+                # It is expected that the first packet will be a descriptor.
+                ig_recv = spead2.ItemGroup()
                 heap = await stream.get()
-                while (updated_items := set(ig_recv.update(heap))) == set():
-                    # Test has gone on long enough that we've received another descriptor
-                    heap = await stream.get()
+                items = ig_recv.update(heap)
+                assert (
+                    len(list(items.values())) == 0
```
bmerry, line 890: I realise you haven't actually changed this code, just reflowed it, but I think this could just be
```suggestion
                    len(items) == 0
```
That would probably allow it to go back to being a single line instead of awkwardly wrapped.

test/xbgpu/test_engine.py
```diff
@@ -1201,7 +1212,7 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
 
         expected_beam_saturated_low, expected_beam_saturated_high = verify_beam_data(
             beam_outputs=beam_outputs,
-            beam_results=beam_results,
+            beam_results=list(beam_results.values()),
```
bmerry, line 1215: This makes me slightly nervous: it depends on the iteration order of the beam_results dictionary matching the order of `beam_outputs`. That is guaranteed with the way you've written the code, but it's an implicit assumption that someone might break then spend time debugging. Some alternative ideas:

1. 
```suggestion
            beam_results=[beam_results[output] for output in beam_outputs],
```

2. Change the parameter in verify_beam_data to a dictionary.
3. Add a comment where `beam_results` is defined to indicate that the iteration order is significant.

Some comment applies to the verify_corrprod_data call.

test/xbgpu/test_engine.py
```diff
@@ -821,77 +818,91 @@ async def _send_data(
...
+                    assert ig_recv["frequency"].value == frequency, (
+                        "Output channel offset not correct. "
+                        f"Expected: {frequency}, "
+                        f"actual: {ig_recv['frequency'].value}."
+                    )
 
-                corrprod_results[i][j] = ig_recv["xeng_raw"].value
+                    corrprod_results[corrprod_output.name][j] = ig_recv["xeng_raw"].value
+            except spead2.Stopped:
+                pass
```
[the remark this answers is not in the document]
bmerry: I'm also not wild about it. My main worry is that if something causes a Stopped exception in one of the other tests, you're going to have zeros in the results and you'll get a data mismatch, which will be misleading and make the failure harder to debug.

My inclination would be to pass extra information into send_data to indicate when it should expect to run out of data. It can then directly verify that this is actually when data stops flowing: after receiving all the expected data, it could try to receive another heap inside a `pytest.raises` context manager (which might still need to handle descriptors).

It might also be nice to then size the results array to the amount of data actually expected, instead of returning arrays with trailing zeros. That's actually what I thought you were going to do when I saw you'd replaced a big array with a dictionary of arrays. I realise this could have knock-on implications for all the verification functions though, so if it's going to be big pain then don't bother.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1465,99 @@ def get_stream_status(stream_name: str) -> bool:
...
+        timestamp_step = n_samples_between_spectra * n_spectra_per_heap
+
+        def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.HeapReference]:
+            timestamp = batch_index * timestamp_step
+            data = np.full((n_channels_per_substream, n_spectra_per_heap, N_POLS, COMPLEX), 10, np.int8)
+            return [
+                spead2.send.HeapReference(gen_heap(timestamp, ant_index, frequency, data))
+                for ant_index in range(n_ants)
+                if present[ant_index]
+            ]
```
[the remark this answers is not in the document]
amishatishpatel: Ah yes, this was on my list of `TODO`s for this PR.

test/xbgpu/test_engine.py
amishatishpatel: Yes, apologies, I'm referring to the results. I noticed that when inspecting the `corrprod_results` ndarray.

test/xbgpu/test_engine.py
amishatishpatel: Noted all around and appreciate you flagging this. Must admit, I had similar concerns, but hadn't considered this particular case.

Regarding
> 2. Change the parameter in verify_beam_data to a dictionary

I'm not quite seeing why that needs to be done if you've suggested list comprehension for the `beam_results` values?
* That is, `[my_dict[this_key] for this_key in my_keys]` just results in a list of values (at least in how I've tested the logic in an ipython session)

bmerry: I meant do one of those things, not all of them.

test/xbgpu/test_engine.py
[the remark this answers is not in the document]
amishatishpatel: > I realise this could have knock-on implications for all the verification functions though, so if it's going to be big pain then don't bother.

Your proposed suggestion is exactly what I intended on doing. Stopped short of that because I didn't want my changes to be too wholesale. Still, I see it's tending that way anyway even with just a "expect data to stop here" parameter. Might as well go all in.

test/xbgpu/test_engine.py
```diff
@@ -770,16 +792,19 @@ async def _send_data(
...
-            corrprod_output, where each array has shape
+            Dictionary of arrays of all XPipeline output. Each key is the
+            corrprod_output name. Each value is an array with shape
             (n_accumulations, n_channels_per_substream, n_baselines, COMPLEX).
         beam_results
-            Beamformer output, with shape (n_beams, n_batches,
-            n_channels_per_substream, n_spectra_per_heap, COMPLEX).
+            Dictionary of arrays of all BPipeline output. Each key is the
+            beam_output name. Each value is an array with shape
+            (n_batches, n_channels_per_substream, n_spectra_per_heap, COMPLEX).
```
amishatishpatel, line 807: I suppose it's not entirely necessary for **both** of these returned results to be dictionaries anymore - now that we are only `capture-stop`'ing the beam data streams.

I've left it as-is just to keep it homogeneous. It will likely cause confusion, and generally just look odd, if the results are different data types/orientations.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1490,125 @@ def get_stream_status(stream_name: str) -> bool:
...
+        # were transmitted before the `?capture-stop` request.
+        for beam_output in beam_outputs:
+            np.testing.assert_equal(beam_results[beam_output.name][:beam_capture_stop_heap_index] != 0, True)
+
+        for corrprod_output in corrprod_outputs:
+            # NOTE: The X-engine output is entirely real (no imag component).
+            # The results buffer has shape (n_accumulations,
+            # n_channels_per_substream, n_baselines, COMPLEX). As a result, the
+            # final dimension only has data populated for the 'real' (first) index.
+            np.testing.assert_equal(corrprod_results[corrprod_output.name][..., 0] != 0, True)
```
amishatishpatel, line 1614: Although looking at this again, I suppose there is scope for me to roll the `corrprod_results` and `beam_results` back to simply numpy `ndarray`s.

Let me know what you think, happy to do so (but will quietly grumble from time to time).

test/xbgpu/test_engine.py
```diff
@@ -1482,82 +1547,68 @@ async def test_capture_stop_some_streams(
...
         Issue a `?capture-stop` request at some point into data processing and
         check the corresponding streams only have partial data transmission.
-        Also ensure that data is completely received for streams that did not
-        receive a `?capture-stop` request.
+        Also ensure that data is completely received for X-engine streams that
+        did not receive a `?capture-stop` request.
+
+        This is only tested of B-engine data streams so as to deterministic in
+        measuring the strength of the test. It has been tested successfully when
+        issuing capture-stop to X-engine data streams as well.
```
bmerry, line 1559: I don't follow this last paragraph at all (possibly because the first line has typos: "tested of" and "as to deterministic").

test/xbgpu/test_engine.py
```diff
@@ -891,7 +923,12 @@ async def _send_data(
                 assert ig_recv["timestamp"].value == index * timestamp_step
                 assert ig_recv["frequency"].value == frequency
                 assert ig_recv["beam_ants"].value == np.sum(present[index])
-                beam_results[i, j, ...] = ig_recv["bf_raw"].value
+                beam_results[beam_output.name][j, ...] = ig_recv["bf_raw"].value
+
+            if beam_capture_stop_heap_index:
```
bmerry, line 928: This `if` can be removed - we might as well test this for all the unit tests, to ensure we've calculated batch_indices correctly.

test/xbgpu/test_engine.py
```diff
@@ -891,7 +923,12 @@ async def _send_data(
...
+                # Confirm that there are no more heaps to receive
+                with pytest.raises(spead2.Stopped):
+                    heap = await stream.get()
```
bmerry, line 931: I think this could fail to raise the Stopped exception if there is a descriptor heap that got sent just after the last data heap. If it is a descriptor heap, it should keep trying.

What I'd suggest is to take the logic from lines 867-872 (which sets and checks updated_items in a loop) and move it into a helper function, which can then be called here to (attempt to) get another data heap.

bmerry (changes requested):
I've made some low-level comments, but the more major concern I have is that this is stopping all of the B streams and none of the X streams. That means that we've got no coverage for the stop logic in XSend, and for BSend we've not checking that the stop is being directed to the correct stream.

From what I see, the main changes that would be needed to stop a subset of the B streams would be to replace some singletons (like beam_capture_stop_heap_index and beam_results_shape) with arrays/dicts to make them per-stream.

test/xbgpu/test_engine.py
amishatishpatel: Ah, my apologies. Quite right, I think I'd started writing a different sentence, then started inserting better wording without removing the old contents. It's meant to read more like
> This is only tested ^on B-engine data streams so as to ^be deterministic in measuring the strength of the test.

That being said, once I implement your suggested change of disabling a subset of **both** data streams then I can remove this docstring addition.

amishatishpatel:
> I've made some low-level comments, but the more major concern I have is that this is stopping all of the B streams and none of the X streams. That means that we've got no coverage for the stop logic in XSend, and for BSend we've not checking that the stop is being directed to the correct stream.
> 
> From what I see, the main changes that would be needed to stop a subset of the B streams would be to replace some singletons (like beam_capture_stop_heap_index and beam_results_shape) with arrays/dicts to make them per-stream.

As always, fair points all around. Seems I misunderstood [your comment](https://github.com/ska-sa/katgpucbf/pull/883#discussion_r1838127281) Re:
> I'd rather be deterministic here so that you have a better idea of how strong the test is. For example, might this allow it to stop all the B-streams and leave all the X-streams running?

Perhaps I'll update to include a test parameter for dictating how many of each type of stream is `capture-stop`'d. That is, for each iteration stop
* Only X streams,
* Some X streams, some B streams
* Only B streams.

bmerry:
> As always, fair points all around. Seems I misunderstood https://github.com/ska-sa/katgpucbf/pull/883#discussion_r1838127281 Re:

> >    I'd rather be deterministic here so that you have a better idea of how strong the test is. For example, might this allow it to stop all the B-streams and leave all the X-streams running?

Indeed - what I meant was that such a test would be bad (weak) and should be avoided, but could potentially arise from randomly choices.

test/xbgpu/test_engine.py
```diff
@@ -866,32 +917,69 @@ async def _send_data(
...
+                    )
+        else:
+            beam_results.update(
+                {
+                    beam_output.name: np.zeros(
+                        (len(batch_indices), n_channels_per_substream, n_spectra_per_heap, COMPLEX), bsend.SEND_DTYPE
+                    )
+                    for beam_output in beam_outputs
+                }
+            )
```
amishatishpatel, line 962: Apologies for the formatting. `black` chose this hill to die on. I can attempt to roll back/condense if you'd like.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+            capture_stop_corrprods = []
+            for stopped_corrprod in stopped_corrprods:
+                stream_index = corrprod_outputs.index(stopped_corrprod)
+                # NOTE: Data is received at the heap level. As a result, we multiply the chunk
+                # ID at which `?capture-stop` was issued by `HEAPS_PER_FENGINE_PER_CHUNK` to
+                # get the specific heap ID at which data ceased transmitting for a stopped
+                # stream.
+                corrprod_capture_stop_heap_indices[stream_index] = (
+                    capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
+                )
```
amishatishpatel, line 1635: Again, `black` doing its thing.

amishatishpatel:
PR build failed on building documentation. Phew.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -480,13 +487,6 @@ def _populate_sensors(self) -> None:
                 )
             )
 
-    async def _get_in_item(self) -> InQueueItem | None:
-        """Get the next :class:`InQueueItem`.
-
-        This is wrapped in a method so that it can be mocked.
-        """
-        return await self._in_queue.get()
-
```
bmerry, line 489: Why did this get moved?

test/xbgpu/test_engine.py
```diff
@@ -729,6 +755,26 @@ def _make_feng(
         )
         return spead2.send.asyncio.InprocStream(spead2.ThreadPool(), queues, feng_stream_config)
 
+    @staticmethod
+    async def get_data_heap(stream: spead2.recv.asyncio.Stream, ig_recv: spead2.ItemGroup) -> set[str]:
+        r"""Continue to receive heaps from a stream until a non-descriptor heap arrives.
```
bmerry, line 760:
```suggestion
        r"""Receive heaps from a stream until a non-descriptor heap arrives.
```

test/xbgpu/test_engine.py
```diff
@@ -770,16 +814,19 @@ async def _send_data(
             Timestamp step between each received heap processed.
         n_channels_per_substream, n_spectra_per_heap, frequency
             See :meth:`_create_heaps` for more info.
+        beam_capture_stop_heap_index
+            Heap count at which a `?capture-stop` was issued to beam streams.
```
bmerry, line 818: Out of sync with the function signature.

test/xbgpu/test_engine.py
```diff
@@ -866,32 +917,69 @@ async def _send_data(
...
+        if beam_capture_stop_heap_indices is not None:
+            # We adjust the corresponding beam stream to only receive the required
+            # amount of data (up until the capture-stop point).
+            for beam_output, capture_stop_heap_index in zip(beam_outputs, beam_capture_stop_heap_indices):
+                if capture_stop_heap_index > 0:
+                    # Obtaining a subset using zero (0) would result in an empty array.
+                    # We also don't want to use negative indices to work from the end
+                    # of the list of indices. Positive values are desired and more
+                    # explicit in this case.
+                    beam_results.update(
```
bmerry, line 932: Using dict.update with a single dictionary is an unusually complicated way of assigning to a single element.

test/xbgpu/test_engine.py
bmerry, line 962: This involves some repetition. It might be simpler to handle this case by instead starting with something like
```python
if beam_capture_stop_heap_indices is None:
    beam_capture_stop_heap_indices = [-1] * len(beam_outputs)
```
and then handling it with the same codepath as if it weren't None.

test/xbgpu/test_engine.py
```diff
@@ -866,32 +917,69 @@ async def _send_data(
...
 
-                corrprod_results[i][j] = ig_recv["xeng_raw"].value
+                corrprod_results[corrprod_output.name][j] = ig_recv["xeng_raw"].value
+
+        beam_results: dict[str, np.ndarray] = {}
+        if beam_capture_stop_heap_indices is not None:
+            # We adjust the corresponding beam stream to only receive the required
+            # amount of data (up until the capture-stop point).
+            for beam_output, capture_stop_heap_index in zip(beam_outputs, beam_capture_stop_heap_indices):
+                if capture_stop_heap_index > 0:
```
bmerry, line 927: It sounds like you're treating non-positive values as magic value meaning "let it run to the end". It would be more explicit None for this purpose, and declare the arguments as `list[int | None]`.

test/xbgpu/test_engine.py
```diff
@@ -866,32 +917,69 @@ async def _send_data(
...
+                    # Obtaining a subset using zero (0) would result in an empty array.
+                    # We also don't want to use negative indices to work from the end
+                    # of the list of indices. Positive values are desired and more
+                    # explicit in this case.
+                    beam_results.update(
+                        {
+                            beam_output.name: np.zeros(
+                                (
+                                    len(batch_indices[:capture_stop_heap_index]),
```
bmerry, line 936:
```suggestion
                                    capture_stop_heap_index,
```
unless you are expecting cases where capture_stop_heap_index exceeds the size of batch_indices.

test/xbgpu/test_engine.py
```diff
@@ -1289,42 +1379,65 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
             assert stream_diff.diff("output_x_clipped_visibilities_total") == n_vis
             assert xbengine.sensors[f"{corrprod_output.name}.xeng-clip-cnt"].value == n_vis
 
-    def _patch_get_in_item(
+    def _patch_method(
         self,
         monkeypatch: pytest.MonkeyPatch,
+        method_name: str,
+        class_type: type[_T],
```
bmerry, line 1386: Calls would read a little easier if these parameters were swapped, so that a patch to Class.method takes Class, method.

test/xbgpu/test_engine.py
```diff
@@ -1289,42 +1379,65 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
         ----------
+        method_name
+            The method intended to be patched.
+        class_type
+            The class which has an attribute `method_name`.
         count
-            Counting from 1, the index of the call to `_get_in_item` to trigger
+            Counting from 1, the index of the call to `method_name` to trigger
             `requests`. `requests` are made prior to the n-th call to
             `_get_in_item`.
```
bmerry, line 1406: Lingering reference to `_get_in_item`.

test/xbgpu/test_engine.py
```diff
@@ -1289,42 +1379,65 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
+
+            NGC-1568: Update the count logic to count from zero instead of one.
         """
         counter = 0
         timestamp = []
-        orig_get_in_item = BPipeline._get_in_item
+        try:
+            orig_method = getattr(class_type, method_name)
+        except AttributeError:
+            raise AttributeError(f"Class {class_type} does not have method {method_name}")
```
bmerry, line 1429: Is it really necessary to catch an AttributeError only to raise another AttributeError?

If for some reason it is (e.g. if the original exception message is unhelpful), then append `from None` here so that you don't get the message that an exception was raised during exception handling along with both exception tracebacks.

test/xbgpu/test_engine.py
```diff
@@ -1358,24 +1471,32 @@ async def test_steady_state(
...
             corrprod_outputs,
             beam_outputs,
-            heap_factory=heap_factory,
+            heap_factory=lambda batch_index, present: self._default_heap_factory(
+                timestamp_step=timestamp_step,
+                n_ants=n_ants,
+                n_channels_per_substream=n_channels_per_substream,
+                n_spectra_per_heap=n_spectra_per_heap,
+                frequency=frequency,
+                data_value=10,
```
bmerry, line 1496: Is it necessary to repeat the default here?

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
 
             await client.request("capture-start", output.name)
             assert get_stream_status(output.name) is True, f"Stream {output.name} is still disabled"
+
+    @DEFAULT_PARAMETERS
+    @pytest.mark.parametrize("n_x_streams_to_stop, n_b_streams_to_stop", [(2, 0), (1, 1), (0, 2)])
```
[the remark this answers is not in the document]
bmerry: This seems fine - it's not as if we're expecting the lengths to change between tests, right?

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+        frequency: int,
+        n_samples_between_spectra: int,
+        n_spectra_per_heap: int,
+        heap_accumulation_threshold: list[int],
+        corrprod_outputs: list[XOutput],
+        beam_outputs: list[BOutput],
+        n_x_streams_to_stop: int,
+        n_b_streams_to_stop: int,
+    ) -> None:
+        """Test capture-stop request on only B-engine data streams.
```
bmerry, line 1596: Docstring seems out of date?

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+        both X- and BPipelines.
+        """
+        assert n_x_streams_to_stop <= len(corrprod_outputs)
+        assert n_b_streams_to_stop <= len(beam_outputs)
+        n_batches = heap_accumulation_threshold[0]
+        # NOTE This value is arbitrarily chosen, but must be less than
+        # `n_batches` / `HEAPS_PER_FENGINE_PER_CHUNK` to actually take effect.
+        # A stop index <= 0 (zero) is treated as invalid/an indication to proceed
+        # as normal for the following reasons:
+        # - Using index zero (0) to obtain a subset of results yields an empty array.
```
bmerry, line 1613: It's probably not a particularly important case to test, but why is that an issue? numpy supports empty arrays.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+                "_get_in_item",
+                XPipeline,
+                capture_stop_chunk_index,
+                client,
+                capture_stop_corrprods,
+            )
+
+        beam_capture_stop_heap_indices = [0] * len(beam_outputs)
+        if n_b_streams_to_stop > 0:
+            stopped_beams = [beam_output for beam_output in beam_outputs[:n_b_streams_to_stop]]
```
bmerry, line 1648: This is a case where randomness might not be a bad idea (take a random selection of outputs instead of a prefix).

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+                # ID at which `?capture-stop` was issued by `HEAPS_PER_FENGINE_PER_CHUNK` to
+                # get the specific heap ID at which data ceased transmitting for a stopped
+                # stream.
+                corrprod_capture_stop_heap_indices[stream_index] = (
+                    capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
+                )
+                capture_stop_corrprods.append(("capture-stop", stopped_corrprod.name))
+            self._patch_method(
+                monkeypatch,
+                "_get_in_item",
```
bmerry, line 1639: Isn't this going to lead to the same sort of race conditions you were running into with the BPipeline? Maybe you just haven't encountered it yet because XPipeline outputs are much chunkier, but it seems risky.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+
+        beam_capture_stop_heap_indices = [0] * len(beam_outputs)
+        if n_b_streams_to_stop > 0:
+            stopped_beams = [beam_output for beam_output in beam_outputs[:n_b_streams_to_stop]]
+            capture_stop_beams = []
+            for stopped_beam in stopped_beams:
+                stream_index = beam_outputs.index(stopped_beam)
+                beam_capture_stop_heap_indices[stream_index] = capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
+                capture_stop_beams.append(("capture-stop", stopped_beam.name))
+            # NOTE: `get_free_chunk` is given `capture_stop_index` + 1 because
```
bmerry, line 1654: More simply, I'd say it's because patch_method counts from 1 instead of from 0.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+
+        # NOTE: During receiving beam data, only data up to `beam_capture_stop_index`
+        # is stored. It is also verified that there is no more data in the stream.
+        # This is largely a sanity check that there is non-zero data for heaps that
+        # were transmitted before the `?capture-stop` request.
+        for beam_output, beam_capture_stop_heap_index in zip(beam_outputs, beam_capture_stop_heap_indices):
+            if beam_capture_stop_heap_index > 0:
+                np.testing.assert_equal(beam_results[beam_output.name][:beam_capture_stop_heap_index] != 0, True)
+            else:
+                np.testing.assert_equal(beam_results[beam_output.name] != 0, True)
```
bmerry, line 1704: Since beam_results is now sized properly, can't this just be
```suggestion
            np.testing.assert_equal(beam_results[beam_output.name] != 0, True)
```

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1573,150 @@ def get_stream_status(stream_name: str) -> bool:
...
+            corrprod_outputs, corrprod_capture_stop_heap_indices
+        ):
+            if corrprod_capture_stop_heap_index > 0:
+                accum_index = corrprod_capture_stop_heap_index // corrprod_output.heap_accumulation_threshold
+                if accum_index == 0:
+                    # The XPipeline never transmitted any accumulations
+                    # Make sure the ndarray is empty
+                    assert corrprod_results[corrprod_output.name].size == 0
+                else:
+                    np.testing.assert_equal(corrprod_results[corrprod_output.name][:accum_index, ..., 0] != 0, True)
```
bmerry, line 1720: Do you need `:accum_index` since that should be the size of the array anyway?

bmerry (changes requested):

test/xbgpu/test_engine.py
```diff
@@ -802,6 +849,17 @@ async def _send_data(
...
+        if corrprod_capture_stop_heap_indices is not None:
+            for i, (corrprod_output, capture_stop_heap_index) in enumerate(
+                zip(corrprod_outputs, corrprod_capture_stop_heap_indices)
+            ):
+                if capture_stop_heap_index > 0:
+                    # Find out which accumulation the heap belongs to
+                    acc_index = capture_stop_heap_index // corrprod_output.heap_accumulation_threshold
+                    # Again, if data goes missing in an accumulation,
+                    # the accumulation is not transmitted.
+                    acc_counts[i][acc_index] = 0
```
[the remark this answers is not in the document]
amishatishpatel: Quite right. Updating now.

test/xbgpu/test_engine.py
amishatishpatel: I was subconsciously hanging on to it because I'd been to a comprehension inside the `.update(...)` previously.

Quite right, I can just `beam_results["beam_output.name"] = np.zeros(...)`.

test/xbgpu/test_engine.py
amishatishpatel: Of course, you are too right. I was thinking about reordering as well, not sure why I left it in this "unintuitive" way.

test/xbgpu/test_engine.py
```diff
@@ -853,12 +855,15 @@ async def _send_data(
...
+                if capture_stop_heap_index is not None and capture_stop_heap_index > 0:
                     # Find out which accumulation the heap belongs to
-                    acc_index = capture_stop_heap_index // corrprod_output.heap_accumulation_threshold
+                    stop_acc_index = capture_stop_heap_index // corrprod_output.heap_accumulation_threshold
                     # Again, if data goes missing in an accumulation,
-                    # the accumulation is not transmitted.
-                    acc_counts[i][acc_index] = 0
+                    # this accumulation and all subsequent accumulations
+                    # are not transmitted.
+                    acc_indices_affected = filter(lambda acc_index: acc_index >= stop_acc_index, acc_counts[i])
```
bmerry, line 864: List comprehensions are preferred Python syntax over `filter` and `map`.

test/xbgpu/test_engine.py
```diff
@@ -853,12 +855,15 @@ async def _send_data(
             for i, (corrprod_output, capture_stop_heap_index) in enumerate(
                 zip(corrprod_outputs, corrprod_capture_stop_heap_indices)
             ):
-                if capture_stop_heap_index > 0:
+                if capture_stop_heap_index is not None and capture_stop_heap_index > 0:
```
bmerry, line 858: It looks like you've introduced None to mean "don't stop" but left behind non-positive numbers to also mean "don't stop". Just eliminate the latter.

bmerry (changes requested):

src/katgpucbf/xbgpu/engine.py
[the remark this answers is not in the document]
amishatishpatel: Oh dear, you are quite right. Not sure how I missed that, must've been on autopilot for a bit.
* Will fix that now.

I was meant to comment on my `@abstractmethod` addition for `_get_in_item`
* Basically saying this doesn't feel entirely correct to "let the developer know this is here if they need it"
* It doesn't do anything, and it's up to the developer to add wrappers as they need.

In principle, are you still ok with both BPipeline and XPipeline having the same helper method?
* Just part of their respective classes to avoid any race conditions.

test/xbgpu/test_engine.py
amishatishpatel: Argh, I had originally typed up a list comprehension. I think I was trying to do too much in it, which is why I opted for this.
* Will update accordingly.

test/xbgpu/test_engine.py
bmerry: Still to be addressed.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -420,6 +420,13 @@ def __init__(
 
         self._populate_sensors()
 
+    async def _get_in_item(self) -> InQueueItem | None:
+        """Get the next :class:`InQueueItem`.
+
+        This is wrapped in a method so it can be mocked.
+        """
+        return await self._in_queue.get()
+
```
amishatishpatel, line 429: Argh, apologies. I thought I put it back where I found it. Will move it once more.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+            capture_stop_accum_index = 3
+            assert capture_stop_accum_index < n_total_accumulations
+            if n_x_streams_to_stop == 1:
+                # Randomly select one of the corrprod outputs to stop
+                stream_id_to_stop = rng.integers(0, len(corrprod_outputs))
+                stopped_corrprods = [corrprod_outputs[stream_id_to_stop]]
+                corrprod_capture_stop_accum_indices[stream_id_to_stop] = capture_stop_accum_index
+            else:
+                stopped_corrprods = corrprod_outputs
+                corrprod_capture_stop_accum_indices = [capture_stop_accum_index] * len(corrprod_outputs)
```
amishatishpatel, line 1613: Apologies if this approach seems a bit... lazy. Brain just couldn't think of something nicer at the moment.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -483,7 +483,7 @@ def _populate_sensors(self) -> None:
     async def _get_in_item(self) -> InQueueItem | None:
         """Get the next :class:`InQueueItem`.
 
-        This is wrapped in a method so that it can be mocked.
+        This is wrapped in a method so it can be mocked.
```
bmerry, line 486: Why this change? The grammar read better before (to me).

test/xbgpu/test_engine.py
```diff
@@ -729,6 +754,26 @@ def _make_feng(
...
+    async def get_data_heap(stream: spead2.recv.asyncio.Stream, ig_recv: spead2.ItemGroup) -> set[str]:
+        r"""Receive heaps from a stream until a non-descriptor heap arrives.
+
+        This modifies `ig_recv` in place in order to obtain an updated set of
+        :class:`spead2.Item`\ s.
+
+        Parameters
+        ----------
+        stream
+            Streams that received XBEngine
```
bmerry, line 767: The parameter list is incomplete. The return value is also not documented (and probably needs it more than the `stream` parameter).

test/xbgpu/test_engine.py
```diff
@@ -866,32 +919,58 @@ async def _send_data(
...
+            # Rather be explicit for each `beam_output` to simplify the logic below
+            beam_capture_stop_heap_indices = [None] * len(beam_outputs)
+        # We adjust the corresponding beam stream to only receive the required
+        # amount of data (up until the capture-stop point).
+        for beam_output, capture_stop_heap_index in zip(beam_outputs, beam_capture_stop_heap_indices):
+            if capture_stop_heap_index is not None:
+                # Obtaining a subset using zero (0) would result in an empty array.
+                # We also don't want to use negative indices to work from the end
+                # of the list of indices. Positive values are desired and more
+                # explicit in this case.
```
bmerry, line 939:
```suggestion
```
Comment is out of date.

test/xbgpu/test_engine.py
```diff
@@ -866,32 +919,58 @@ async def _send_data(
...
+                        n_spectra_per_heap,
+                        COMPLEX,
+                    ),
+                    bsend.SEND_DTYPE,
+                )
+            else:
+                beam_results[beam_output.name] = np.zeros(
+                    (len(batch_indices), n_channels_per_substream, n_spectra_per_heap, COMPLEX),
+                    bsend.SEND_DTYPE,
+                )
```
bmerry, line 953: This code is mostly a copy of the `if` branch. I'd suggest you instead do something like
```python
n_batches = len(batch_indices) if capture_stop_heap_index is None else capture_stop_heap_index
beam_results[beam_output.name] = np.zeros(
    (n_batches, n_channels_per_substream, n_spectra_per_heap, COMPLEX),
    bsend.SEND_DTYPE,
)
```
(untested)

test/xbgpu/test_engine.py
```diff
@@ -1289,42 +1370,57 @@ def heap_factory(batch_index: int, present: np.ndarray) -> list[spead2.send.Heap
...
         self,
         monkeypatch: pytest.MonkeyPatch,
+        class_or_instance: object,
+        method_name: str,
         count: int,
         client: aiokatcp.Client,
         requests: Iterable[Iterable],
     ) -> list[int]:
-        """Patch :meth:`~.BPipeline._get_in_item` to make requests partway through the stream.
+        """Patch `method_name` of class `class_type` to make requests during operation.
```
bmerry, line 1382: Lingering reference to `class_type`.

`class_or_instance` is a bit of a mouthful. I'd just call it `obj` (since *any* object is a class or instance), and `obj` is the [parameter name](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#how-to-monkeypatch-mock-modules-and-environments) used by monkeypatch.

It is probably worth noting in the body of the docstring that it can be used to patch both classes and instances though.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+        n_channels_per_substream: int,
+        frequency: int,
+        n_samples_between_spectra: int,
+        n_spectra_per_heap: int,
+        heap_accumulation_threshold: list[int],
+        corrprod_outputs: list[XOutput],
+        beam_outputs: list[BOutput],
+        n_x_streams_to_stop: int,
+        n_b_streams_to_stop: int,
+        xbengine: XBEngine,
```
bmerry, line 1580: Can you file a bug on the ordering dependency that breaks the test if `xbengine` is earlier in the parameter order? pytest doesn't guarantee anything about how order of fixtures in the function signature affects fixture initialisation order, so it's quite possible that this will come back to bite us in a future pytest version.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+        both X- and BPipelines.
+        """
+        rng = np.random.default_rng(seed=1)
+        # NOTE: This value is arbitrarily chosen to have a few whole accumulations
+        # for each corrprod_output. This allows us to issue a capture-stop request
+        # partway in its processing.
+        n_total_accumulations = 5
+        n_batches = heap_accumulation_threshold[0] * n_total_accumulations
+        corrprod_capture_stop_accum_indices: list[int | None] = [None] * len(corrprod_outputs)
+        if n_x_streams_to_stop is not None:
```
bmerry, line 1599: Wouldn't it be simpler to just 0 instead of None, and then just make the code unconditional?

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+            assert capture_stop_accum_index < n_total_accumulations
+            if n_x_streams_to_stop == 1:
+                # Randomly select one of the corrprod outputs to stop
+                stream_id_to_stop = rng.integers(0, len(corrprod_outputs))
+                stopped_corrprods = [corrprod_outputs[stream_id_to_stop]]
+                corrprod_capture_stop_accum_indices[stream_id_to_stop] = capture_stop_accum_index
+            else:
+                stopped_corrprods = corrprod_outputs
+                corrprod_capture_stop_accum_indices = [capture_stop_accum_index] * len(corrprod_outputs)
+            for stopped_corrprod in stopped_corrprods:
```
bmerry, line 1614: Try this (totally untested):
```suggestion
            stream_ids_to_stop = sorted(rng.choice(len(corrprod_outputs), n_x_streams_to_stop, replace=False))
            stopped_corrprods = [corrprod_outputs[stream_id] for id in stream_id in stream_ids_to_stop]
            for stream_id in stream_ids_to_stop:
                corrprod_capture_stop_accum_indices[stream_id] = capture_stop_accum_index
            for stopped_corrprod in stopped_corrprods:
```

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+                stream_id_to_stop = rng.integers(0, len(beam_outputs))
+                stopped_beams = [beam_outputs[stream_id_to_stop]]
+                beam_capture_stop_heap_indices[stream_id_to_stop] = (
+                    capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
+                )
+            else:
+                stopped_beams = beam_outputs
+                beam_capture_stop_heap_indices = [capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK] * len(
+                    beam_outputs
+                )
```
bmerry, line 1646: Suggestion for corrprod outputs also applies here.

test/xbgpu/test_engine.py
```diff
@@ -1452,3 +1555,153 @@ def get_stream_status(stream_name: str) -> bool:
...
+        # n_baselines, COMPLEX). As a result, the final dimension only has data
+        # populated for the 'real' (first) index.
+        for corrprod_output, corrprod_capture_stop_accum_index in zip(
+            corrprod_outputs, corrprod_capture_stop_accum_indices
+        ):
+            if corrprod_capture_stop_accum_index is not None:
+                if corrprod_capture_stop_accum_index == 0:
+                    # The XPipeline never transmitted any accumulations
+                    # Make sure the ndarray is empty
+                    assert corrprod_results[corrprod_output.name].size == 0
```
bmerry, line 1703: Is this code even reachable, given that you've explicitly set `capture_stop_accum_index` to 3?

bmerry (changes requested):

src/katgpucbf/xbgpu/engine.py
amishatishpatel: I'd tweaked in when I duplicated the method for each Pipeline. Will roll it back to how it was now-now.

test/xbgpu/test_engine.py
amishatishpatel: Will do.

test/xbgpu/test_engine.py
amishatishpatel: Fair enough. I updated the zeroes to `None`s in a naive attempt to be consistent with the other logical tweak you suggested.

test/xbgpu/test_engine.py
amishatishpatel: Fair enough. I'll remove.

test/xbgpu/test_engine.py
amishatishpatel: Ah, fair enough. I had a peek at the type hint monkeypatch uses. Didn't go deep enough for the parameter name.

test/xbgpu/test_engine.py
bmerry: The difference is that in the other case zero didn't actually mean zero (as in stop after zero chunks) - it triggered a behaviour qualitatively different from other integers. Here zero would really mean zero streams.

test/xbgpu/test_engine.py
```diff
@@ -1596,68 +1595,57 @@ async def test_capture_stop_some_streams(
...
+        # `n_total_accumulations` in order to accurately test stopping of
+        # corrprod data streams.
+        capture_stop_accum_index = 3
+        assert capture_stop_accum_index < n_total_accumulations
+        stream_ids_to_stop = sorted(rng.choice(len(corrprod_outputs), n_x_streams_to_stop, replace=False))
+        stopped_corrprods: list[XOutput] = []
+        for stream_id in stream_ids_to_stop:
+            stopped_corrprods.append(corrprod_outputs[stream_id])
+            corrprod_capture_stop_accum_indices[stream_id] = capture_stop_accum_index
+        for stopped_corrprod in stopped_corrprods:
```
bmerry, line 1609: I think this can be simplified a little now, since you're building up `stopped_corrprods` just to loop over it immediately after:
```suggestion
        for stream_id in stream_ids_to_stop:
            stopped_corrprod = corrprod_outputs[stream_id]
            corrprod_capture_stop_accum_indices[stream_id] = capture_stop_accum_index
```

test/xbgpu/test_engine.py
```diff
@@ -1596,68 +1595,57 @@ async def test_capture_stop_some_streams(
...
+        capture_stop_chunk_index = 10
+        assert capture_stop_chunk_index < n_batches // HEAPS_PER_FENGINE_PER_CHUNK
+        stream_ids_to_stop = sorted(rng.choice(len(beam_outputs), n_b_streams_to_stop, replace=False))
+        stopped_beams: list[BOutput] = []
+        for stream_id in stream_ids_to_stop:
+            stopped_beams.append(beam_outputs[stream_id])
+            beam_capture_stop_heap_indices[stream_id] = capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
+        capture_stop_beams = []
+        for stopped_beam in stopped_beams:
+            capture_stop_beams.append(("capture-stop", stopped_beam.name))
```
bmerry, line 1637: I think this can similarly be simplified:
```suggestion
        capture_stop_beams = []
        for stream_id in stream_ids_to_stop:
            beam_capture_stop_heap_indices[stream_id] = capture_stop_chunk_index * HEAPS_PER_FENGINE_PER_CHUNK
            capture_stop_beams.append(("capture-stop", beam_outputs[stream_id].name))
```

bmerry (changes requested):
Looking good. Just a couple of suggestions to do simplifications that are possible now that you've applied other changes.

test/xbgpu/test_engine.py
amishatishpatel: Ah, too right. My apologies, I was still holding on to the idea that I was using `stopped_corrprods` later on when verifying data. Good spot.

bmerry (approved):

PR #885, opened 2024-11-13
Compare like pols in test_gain

james-smith-za:
The qualification/tied_array_channelised_voltage/test_gain test occasionally shows a difference one greater expected for a single value in the test every now and then. It's very intermittent. To a first approximation this shouldn't be possible because the tolerance for each test should be taken into account and there's no way for dithering to cause a bigger difference assuming that the incoming signal is the same.

We think that this is caused by separate dithering for each pol in the F-engine which means that despite having the same d-sim input, the F-engine output for pol0 and pol1 may be ever so slightly different from each other, just different to cause the error.

In this commit I adjust the test to using the same pol in three different beams for comparison, rather than three half-beams which would have been pol0, pol1, pol0.

This appears to solve the problem.

Closes NGC-1473.

- [ ] If qualification tests are changed: attach a sample qualification report **No significant change in report.**

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_gain.py
+++ b/qualification/tied_array_channelised_voltage/test_gain.py
@@ -79,9 +79,9 @@ async def test_gain(
     pdf_report.step("Compare beams to each other.")
     # Compare like pols because the dithering is independent between pols and may rarely cause
     # differences greater than expected.
-    for i in [2, 4]:
-        scale = gains[i] / gains[0]
-        expected = data[0] * scale
+    for i in range(2, len(gains)):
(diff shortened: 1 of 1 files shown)


bmerry (changes requested):
It looks like you're capturing 6 half-beams, but then ignoring 3 of them. I'd suggest either
(a) set stream_names to grab only the 3 half-beams you're interested in; or
(b) update the comparison code to test the odd-indexed polarisations against each other.

I'd lean towards option (b) because it's a stronger test (ensures we haven't mucked up gain-setting such that it only applies to one polarisation).

bmerry (approved):

PR #886, opened 2024-11-21
Move pip to dockerfile

james-smith-za:
to benefit from the caching of docker builds.

Also, chown the venv in the jenkins build to 1000:1000 so that jenkins-test can run it without any hackery of who is running what.

- [x] (N/A)  If modules are added/removed: use `sphinx-apidoc -efo doc/ src/` to update files in `doc/`
- [x] (N/A)  If qualification tests are changed: attach a sample qualification report

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,6 +78,7 @@ RUN pip install pip==24.2 setuptools==72.1.0 wheel==0.44.0
 # Jenkins image has a clean environment to start from).
 WORKDIR /tmp/katgpucbf
 COPY requirements.txt .
+COPY requirements-dev.txt .
 RUN pip install --no-deps -c /tmp/katgpucbf/requirements.txt pycuda && \
     pip uninstall -y pycuda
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -78,6 +78,7 @@ RUN pip install pip==24.2 setuptools==72.1.0 wheel==0.44.0
 # Jenkins image has a clean environment to start from).
 WORKDIR /tmp/katgpucbf
 COPY requirements.txt .
+COPY requirements-dev.txt .
```
bmerry, line 81: If we don't need this in the base, let's rather move it to the jenkins image section, so that changes to requirements-dev.txt don't invalidate the caches of other images.

Dockerfile
```diff
@@ -101,6 +102,12 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
     texlive-latex-recommended \
     texlive-science
 
+# Install required packages for testing to be able to run
+RUN pip install -r requirements.txt -r requirements-dev.txt
+
+# Jenkins runs the containers with the `-u 1000:1000` option, not as root.
+RUN chown -Rf +1000:+1000 /venv
```
bmerry, line 109: `-f` is documented as "suppresses most error messages". Don't we want to know if something has gone wrong?

I'd never seen the leading +'s before, but I see the documentation does allow it and it ensures it's interpreted numerically even if there is a user called '1000'.

bmerry (changes requested):

Dockerfile
james-smith-za: Fair point about the `-f`.

I read through the `man` page for `chown` and couldn't see how to do it in a good way, so I will admit I asked an LLM, and it gave me the + notation. I did check it before committing though.

Dockerfile
james-smith-za: Fair point

bmerry (approved):

PR #888, opened 2024-11-27
Modify qualification Jenkinsfile for saner agents

james-smith-za:
In the Jenkins setup that our former colleague put together, we had "clouds" rather than Nodes, which ran their containers as root. It was a rather complicated setup and we've simplified things, but in the new regime containers don't get run as root anymore so they need slightly different capabilities.

This commit adds spead2_net_raw into the Jenkins image so that the agent that runs qualification can use its magic to enable ibverbs. In order for that to work, we need to install a couple more libraries in the `jenkins` docker image. We also add usage of the spead2_net_raw executable to endow pytest with ibverbs.

Contribures to NGC-1447.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,6 +93,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
     docker-buildx \
     fonts-liberation2 \
     latexmk \
+    libcap2-dev \
+    libcap2-bin \
     lmodern \
     pdf2svg \
     tex-gyre \
@@ -100,6 +102,12 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
     texlive-latex-extra \
     texlive-latex-recommended \
     texlive-science
+
+# spead2_net_raw is needed so that qualification tests can use ibverbs
+RUN wget https://raw.githubusercontent.com/ska-sa/spead2/master/src/spead2_net_raw.cpp && \
+    gcc -Wall -O2 -o spead2_net_raw spead2_net_raw.cpp -lcap && \
+    mv spead2_net_raw /usr/local/bin && \
+    setcap cap_net_raw+p /usr/local/bin/spead2_net_raw
 
 # Install required packages for testing to be able to run
 COPY requirements-dev.txt .
--- a/Dockerfile
+++ b/Dockerfile
@@ -113,11 +113,10 @@ RUN chown -R +1000:+1000 /venv
 # spead2_net_raw is needed so that qualification tests can use ibverbs
 RUN SPEAD2_VERSION=$(grep ^spead2== requirements.txt | cut -d= -f3) && \
     cd /tmp && \
-    wget https://github.com/ska-sa/spead2/releases/download/v$SPEAD2_VERSION/spead2-$SPEAD2_VERSION.tar.gz && \
-    tar -zxf "spead2-$SPEAD2_VERSION.tar.gz" && \
-    cd "spead2-$SPEAD2_VERSION/src" %&& \
+    wget https://raw.githubusercontent.com/ska-sa/spead2/refs/tags/v$SPEAD2_VERSION/src/spead2_net_raw.cpp && \
     gcc -Wall -O2 -o /usr/local/bin/spead2_net_raw spead2_net_raw.cpp -lcap && \
-    setcap cap_net_raw+p /usr/local/bin/spead2_net_raw
+    setcap cap_net_raw+p /usr/local/bin/spead2_net_raw && \
+    rm spead2_net_raw.cpp
 
 #######################################################################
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -101,6 +103,12 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
     texlive-latex-recommended \
     texlive-science
 
+# spead2_net_raw is needed so that qualification tests can use ibverbs
+RUN wget https://raw.githubusercontent.com/ska-sa/spead2/master/src/spead2_net_raw.cpp && \
```
bmerry, line 107: I'd suggest specifying the output path. At the moment it's wherever a previous build stage happens to leave WORKDIR, which could cause some spooky action at a distance if we make changes elsewhere in the file.

Dockerfile
```diff
@@ -101,6 +103,12 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
...
+    gcc -Wall -O2 -o spead2_net_raw spead2_net_raw.cpp -lcap && \
+    mv spead2_net_raw /usr/local/bin && \
```
bmerry, line 109: Why not just
```suggestion
    gcc -Wall -O2 -o /usr/local/bin/spead2_net_raw spead2_net_raw.cpp -lcap && \
```

Dockerfile
bmerry, line 107: Using master is not ideal - it means we're not isolated from changes on spead2. https://github.com/ska-sa/katgpucbf/commit/d842d297a0fd509d4b8fcbc42914cd82e891ec74 shows how an older version of the unit test Jenkinsfile used to extract the pinned version of spead2.

qualification/Jenkinsfile
```diff
@@ -86,7 +86,7 @@ pipeline {
           * tests and mark the build as unstable.
           */
          sh '''
-           pytest -v -c qualification/pytest-jenkins.ini qualification\
+           spead2_net_raw pytest -v -c qualification/pytest-jenkins.ini qualification\
```
bmerry, line 89: Not really related to your PR, but for my OCD:
```suggestion
           spead2_net_raw pytest -v -c qualification/pytest-jenkins.ini qualification \
```

bmerry (changes requested):

Dockerfile
james-smith-za: Makes sense.

james-smith-za: Thanks for the tip.

Dockerfile
```diff
@@ -108,6 +110,15 @@ RUN pip install -r requirements.txt -r requirements-dev.txt
 # Jenkins runs the containers with the `-u 1000:1000` option, not as root.
 RUN chown -R +1000:+1000 /venv
 
+# spead2_net_raw is needed so that qualification tests can use ibverbs
+RUN SPEAD2_VERSION=$(grep ^spead2== requirements.txt | cut -d= -f3) && \
+    cd /tmp && \
+    wget https://github.com/ska-sa/spead2/releases/download/v$SPEAD2_VERSION/spead2-$SPEAD2_VERSION.tar.gz && \
```
bmerry, line 116: I'd keep what you have previously to just pull the file directly from Github (via the tag) rather than downloading a release (should be lighter-weight than downloading a whole tarball, 99% of which you don't need); the snippet I directed you to was intended just to show you how to set SPEAD2_RELEASE.

Dockerfile
```diff
@@ -108,6 +110,15 @@ RUN pip install -r requirements.txt -r requirements-dev.txt
...
+    tar -zxf "spead2-$SPEAD2_VERSION.tar.gz" && \
+    cd "spead2-$SPEAD2_VERSION/src" %&& \
```
bmerry, line 118: What is the `%` for? When I try it in bash I get an error about too many arguments.

Dockerfile
```diff
@@ -108,6 +110,15 @@ RUN pip install -r requirements.txt -r requirements-dev.txt
...
+    gcc -Wall -O2 -o /usr/local/bin/spead2_net_raw spead2_net_raw.cpp -lcap && \
+    setcap cap_net_raw+p /usr/local/bin/spead2_net_raw
+
```
bmerry, line 121: For tidyness, I'd also suggest removing the source code at the end of this step, so that it doesn't leave a random file lying around in the container.

bmerry (changes requested):

Dockerfile
james-smith-za: I think that's finger trouble. Good catch. I wonder why the thing built... 🤔

Dockerfile
james-smith-za: I'd kind of hoped that by virtue of being in `/tmp`, that would be taken care of.

bmerry (approved):

PR #890, opened 2024-11-29
Simplify treatment of vkgdr in benchmark script

bmerry:
Remove special handling from benchmark_fgpu when running with --use-vkgdr. Previously this was needed because we only had a 256 MiB BAR, but with a BIOS update we get a full-sized BAR and no longer need these options.

Relates to NGC-1477

### DIFF
--- a/scratch/benchmarks/benchmark_fgpu.py
+++ b/scratch/benchmarks/benchmark_fgpu.py
@@ -120,12 +120,8 @@ def fgpu_factory(
     hstep = step // 2
     qstep = step // 4
     my_cpus = server_info.cpus[index * step : (index + 1) * step]
-    if args.use_vkgdr:
-        recv_chunk_samples = 2**24 // scaling_n
-        send_chunk_jones = recv_chunk_samples // 2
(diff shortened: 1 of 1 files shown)


scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -172,7 +168,7 @@ def fgpu_factory(
         f"--prometheus-port={prometheus_port} "
         f"--sync-time={sync_time} "
         f"--feng-id={index} "
-        f"{'--use-vkgdr --max-delay-diff=65536' if args.use_vkgdr else ''} "
+        f"{'--use-vkgdr' if args.use_vkgdr else ''} "
```
amishatishpatel, line 171: You don't need `--max-delay-diff` anymore?

amishatishpatel (approved):
Great.

scratch/benchmarks/benchmark_fgpu.py
bmerry: That was to save memory.

PR #892, opened 2024-12-02
Disable concurrent builds in Jenkinsfile

james-smith-za:
The intention is to check whether we can concurrently run separate branches on an agent with two executors.

On test-Jenkins it appears to work, different branches run happily side-by-side, the same branch however sits and waits for its predecessor to complete.

- [x] (N/A) f dependencies are added/removed: update `setup.cfg` and `.pre-commit-config.yaml`

Contributes to NGC-1447.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,7 @@ pipeline {
   agent { label 'katgpucbf' }
   options {
     timeout(time: 2, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


bmerry (approved):
Looks good for now. At some stage we might want to consider `disableConcurrentBuilds(abortPrevious: true)` but that's a separate issue.

PR #893, opened 2024-12-03
Add copy-mk script

james-smith-za:
This script connects to a running MK CMC, figures out some particulars from it, and leverages sim_correlator.py in order to fire up a similar correlator on MK+.

Relates to NGC-1551.

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -127,7 +127,7 @@ async def async_main(args) -> int:
         "band": band,
     }
 
-    out_cmd = [str(os.path.join(os.path.dirname(__file__), "sim_correlator.py"))]
+    out_cmd = [os.path.join(os.path.dirname(__file__), "sim_correlator.py")]
 
     for k, v in out_kwargs.items():
         out_cmd.append(f"--{k}={v}")
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -45,7 +45,7 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     """Parse command-line arguments."""
     parser = argparse.ArgumentParser()
     parser.add_argument("cmc_host", help="Hostname / IP address of the CMC.")
-    parser.add_argument("--cmc_port", type=int, default=7147, help="KATCP port of the CMC. [%(default)s]")
+    parser.add_argument("--cmc-port", type=int, default=7147, help="KATCP port of the CMC. [%(default)s]")
     parser.add_argument("--target", type=str, required=True, help="Which subordinate to copy.")
     parser.add_argument(
         "--develop",
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 ################################################################################
-# Copyright (c) 2024, National Research Foundation (SARAO)
+# Copyright (c) 2024-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -85,7 +85,7 @@ async def async_main(args) -> int:
         n_antennas = len(mcast_groups) // 2
         for n, grp in enumerate(mcast_groups[1:]):
             if grp == mcast_groups[0]:
-                # MK is not using a full power-of-2 size array, there are dummies from here onwards.
+                # MK is not using a full power-of-2 size array. There are dummies from here onwards.
                 print("Dummy antenna found starting at ", (n + 1) // 2)
                 n_antennas = (n + 1) // 2
                 break
@@ -142,7 +142,7 @@ async def async_main(args) -> int:
 
     out_cmd.append("cbf-mc.cbf.mkat.karoo.kat.ac.za")
 
-    print(f"Executing: {out_cmd}", "\n")
+    print(f"Executing: {out_cmd}\n")
     execv(out_cmd[0], out_cmd[1:])
 
 
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -104,11 +104,11 @@ async def async_main(args) -> int:
     try:
         target = subordinates[args.target]
     except KeyError:
-        print(f"{args.target} is not among the subordinates currently running. Options are: {subordinates.keys()}")
+        print(f"{args.target} is not among the subordinates currently running. Options are: {subordinates.keys()}", file=sys.stderr)
         return 1
 
     # connect to the corr2_servlet itself to see the sync time, n_chans and bandwidth
-    corr2_client = await aiokatcp.Client.connect(args.cmc_host, control_port)
+    corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
     sync_time = await get_sensor_value(corr2_client, "sync-time", float)
     channels = await get_sensor_value(corr2_client, "antenna-channelised-voltage-n-chans", int)
     bandwidth = await get_sensor_value(corr2_client, "antenna-channelised-voltage-bandwidth", float)
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -24,12 +24,14 @@ instead of a simulated one, it gets the parameters from a live MK correlator.
 
 import argparse
 import asyncio
-import subprocess
 import sys
 from collections.abc import Sequence
 from dataclasses import dataclass
+from os import execv
 
 import aiokatcp
+
+from katgpucbf.meerkat import BANDS
 
 
 @dataclass
@@ -101,16 +103,19 @@ async def async_main(args) -> int:
     corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
     sync_time = await corr2_client.sensor_value("sync-time", float)
     channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
-    bandwidth = await corr2_client.sensor_value("antenna-channelised-voltage-bandwidth", float)
-    # TODO: this could probably be better, and doesn't support S-band yet.
-    match bandwidth:
-        case 544000000.0:
-            band = "u"
-        case 856000000.0:
-            band = "l"
-        case _:
-            print(f"Unable to determine which band we are in. Bandwidth reported as {bandwidth}", file=sys.stderr)
-            return 1
+    sample_rate = await corr2_client.sensor_value("antenna-channelised-voltage-bandwidth", float) * 2
+
+    # This won't distinguish between the different kinds of S-band. It'll be wrong. I'm not quite
+    # sure at this point how to match up the centre_frequency values in katgpucbf.meerkat.BANDS with
+    # what MK actually reports, even in the L and UHF cases, so for the purposes of getting
+    # something producing a result, I am leaving it as-is for the time being.
+    band = next((k for k, v in BANDS.items() if v.adc_sample_rate == sample_rate), None)
+    if not band:
+        print(
+            f"Unable to determine which band. Sample rate reported as {sample_rate}",
+            file=sys.stderr,
+        )
+        return 1
 
     out_kwargs = {
         "name": target.name.replace(".", "_"),
@@ -122,9 +127,7 @@ async def async_main(args) -> int:
     }
 
     out_cmd = [
-        "python",
-        "sim_correlator.py",
-        "cbf-mc.cbf.mkat.karoo.kat.ac.za",
+        "./sim_correlator.py",
     ]
 
     for k, v in out_kwargs.items():
@@ -137,10 +140,10 @@ async def async_main(args) -> int:
         for override in args.image_override:
             out_cmd.append(f"--image-override={override}")
 
-    output = subprocess.run(out_cmd, capture_output=True)
-    print(output)
+    out_cmd.append("cbf-mc.cbf.mkat.karoo.kat.ac.za")
 
-    return output.returncode
+    print(f"Executing: {out_cmd}", "\n")
+    execv(out_cmd[0], out_cmd[1:])
 
 
 def main(argv: Sequence[str] | None = None) -> int:
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -26,7 +26,7 @@ import argparse
 import asyncio
 import subprocess
 import sys
-from collections.abc import Callable, Sequence
+from collections.abc import Sequence
 from dataclasses import dataclass
 
 import aiokatcp
@@ -56,16 +56,6 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     parser.add_argument("--image-override", action="append", metavar="NAME:IMAGE:TAG", help="Override a single image")
     args = parser.parse_args(argv)
     return args
-
-
-async def get_sensor_value(client: aiokatcp.Client, sensor_name: str, sensor_type: Callable):
-    """Retrieve a sensor value from a katcp client."""
-    try:
-        _, informs = await client.request("sensor-value", sensor_name)
-    except (aiokatcp.FailReply, ConnectionError) as exc:
-        print(f"Error: {exc}", file=sys.stderr)
-        sys.exit(1)
-    return sensor_type(informs[0].arguments[4].decode())
 
 
 async def async_main(args) -> int:
@@ -109,9 +99,9 @@ async def async_main(args) -> int:
 
     # connect to the corr2_servlet itself to see the sync time, n_chans and bandwidth
     corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
-    sync_time = await get_sensor_value(corr2_client, "sync-time", float)
-    channels = await get_sensor_value(corr2_client, "antenna-channelised-voltage-n-chans", int)
-    bandwidth = await get_sensor_value(corr2_client, "antenna-channelised-voltage-bandwidth", float)
+    sync_time = await corr2_client.sensor_value("sync-time", float)
+    channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
+    bandwidth = await corr2_client.sensor_value("antenna-channelised-voltage-bandwidth", float)
     # TODO: this could probably be better, and doesn't support S-band yet.
     match bandwidth:
         case 544000000.0:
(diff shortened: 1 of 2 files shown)


james-smith-za:
Not particularly neat yet but I'm putting this here so that it can be used / examined if anyone else feels the need.

scratch/copy-mk.py
```diff
@@ -116,7 +116,7 @@ async def async_main(args) -> int:
     match bandwidth:
         case 544000000.0:
             band = "u"
-        case 85600000.0:
+        case 856000000.0:
```
bmerry, line 119: You can get this information from the katgpucbf.meerkat module.

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+    monitor_port: int
+    digitiser_address: str
+    n_antennas: int
+
+
+def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
+    """Parse command-line arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("cmc_host", help="Hostname / IP address of the CMC.")
+    parser.add_argument("--cmc_port", type=int, default=7147, help="KATCP port of the CMC. [%(default)s]")
```
bmerry, line 48:
```suggestion
    parser.add_argument("--cmc-port", type=int, default=7147, help="KATCP port of the CMC. [%(default)s]")
```

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+    parser.add_argument("--target", type=str, required=True, help="Which subordinate to copy.")
```
bmerry, line 49: I'm less familiar with MK nomenclature, but is this the same as the "instrument"?

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+async def get_sensor_value(client: aiokatcp.Client, sensor_name: str, sensor_type: Callable):
+    """Retrieve a sensor value from a katcp client."""
+    try:
+        _, informs = await client.request("sensor-value", sensor_name)
+    except (aiokatcp.FailReply, ConnectionError) as exc:
+        print(f"Error: {exc}", file=sys.stderr)
+        sys.exit(1)
+    return sensor_type(informs[0].arguments[4].decode())
```
bmerry, line 68: This functionality is built in to aiokatcp now: `client.sensor_value(sensor_name, sensor_type)`

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+async def async_main(args) -> int:
+    cmc_client = await aiokatcp.Client.connect(args.cmc_host, args.cmc_port)
+
+    try:
+        reply, informs = await cmc_client.request("subordinate-list")
```
bmerry, line 75:
```suggestion
        _, informs = await cmc_client.request("subordinate-list")
```
I think that'll make flake8 happier.

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+        control_port, monitor_port = inform.arguments[1].decode().split(",")
+        control_port = int(control_port)
+        monitor_port = int(monitor_port)
+        mcast_groups = []
+        for grp in inform.arguments[2:]:
+            mcast_groups.append(grp.decode())
+        n_antennas = len(mcast_groups) // 2
+        for n, grp in enumerate(mcast_groups[1:]):
+            if grp == mcast_groups[0]:
+                # MK is not using a full power-of-2 size array, there are dummies from here onwards.
```
bmerry, line 96: Comma splice

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+                print("Dummy antenna found starting at ", (n + 1) // 2)
+                n_antennas = (n + 1) // 2
+                break
+        digitiser_address = mcast_groups[0].split("+")[0]
```
bmerry, line 100: Not something that needs to get done now, but if you ever wanted to make this more robust/production-ready, it would be best to check that the mcast_groups are all numbered contiguously, have +7 etc. Probably the simplest way to do that would be to reconstruct what sim_correlator is going to think the multicast groups are and check that this matches.

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+
+        subordinates[name] = Subordinate(name, control_port, monitor_port, digitiser_address, n_antennas)
+
+    try:
+        target = subordinates[args.target]
+    except KeyError:
+        print(f"{args.target} is not among the subordinates currently running. Options are: {subordinates.keys()}")
```
bmerry, line 107:
```suggestion
        print(f"{args.target} is not among the subordinates currently running. Options are: {subordinates.keys()}", file=sys.stderr)
```

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+        return 1
+
+    # connect to the corr2_servlet itself to see the sync time, n_chans and bandwidth
+    corr2_client = await aiokatcp.Client.connect(args.cmc_host, control_port)
```
bmerry, line 111:
```suggestion
    corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
```

scratch/copy-mk.py
```diff
@@ -0,0 +1,162 @@
...
+
+    if args.develop is not None:
+        out_cmd.append(f"--develop={args.develop}")
+
+    if args.image_override is not None:
+        for override in args.image_override:
+            out_cmd.append(f"--image-override={override}")
+
+    output = subprocess.run(out_cmd, capture_output=True)
+    print(output)
```
bmerry, line 151: Is there a particular reason you want to capture the output, rather than letting the command just write directly to the stdout?

You could also consider using execve (forget the exact Python interface), which will basically replace this process with the out_cmd so that you don't need to worry about trying to proxy anything like output and return codes.

bmerry (changes requested):

scratch/copy-mk.py
james-smith-za: Nice!

scratch/copy-mk.py
james-smith-za: Are you referring to "subordinate"? If so, then yes. The request and sensor names all use the term "subordinate", though I'm not really sure why. Tyrone may remember.

scratch/copy-mk.py
james-smith-za: I did think about that but I don't know that it's worth being production-ready because this should in principle be CAM's function.

I've watched a few times and it only ever looks like contiguous ranges of addresses are assigned, so we'll probably get away with it this time.

james-smith-za:
```
2024-12-13 05:09:19.838 [cbfgpu001.cbf.mkat.karoo.kat.ac.za](http://cbf-mc.cbf.mkat.karoo.kat.ac.za:5601/app/logtrail) product_controller: array_1_wide  Failed to process config: {'data_timeout: 0.0': True} is not valid under any of the given schemas

Failed validating 'oneOf' in schema['properties']['config']['properties']['develop']:
    {'default': False,
     'oneOf': [{'additionalProperties': False,
                'properties': {'any_gpu': {'default': False,
                                           'type': 'boolean'},
                               'data_timeout': {'$ref': '#/definitions/nonneg_number'},
                               'disable_ibverbs': {'default': False,
                                                   'type': 'boolean'},
                               'less_resources': {'default': False,
                                                  'type': 'boolean'}},
                'type': 'object'},
               {'default': False, 'type': 'boolean'}]}

On instance['config']['develop']:
    {'data_timeout: 0.0': True}
```

Clearly *something* along the chain is not parsing the option that I passed properly.

scratch/copy-mk.py
bmerry: Should we name the option `--subordinate` then to make it clear?

bmerry:
> Clearly _something_ along the chain is not parsing the option that I passed properly.

My guess would be sim_correlator.py - from what I recall, it assumes that develop options are boolean flags.

scratch/copy-mk.py
james-smith-za: I can do that.

scratch/copy-mk.py
james-smith-za: I guess I should.
I did think of that, but in my mind I didn't really want to have the entire module installed on my laptop just to use this script (since I don't have a GPU, I have so far at least on this new laptop avoided having to try and install katgpucbf).

But I guess this is a silly objection.

james-smith-za:
> > Clearly _something_ along the chain is not parsing the option that I passed properly.
> 
> My guess would be sim_correlator.py - from what I recall, it assumes that develop options are boolean flags.

I think you are on to something.

```
 On instance['config']['develop']:
    {'data_timeout: 0.0': True}
```
It seems as though it's not properly distinguishing between the key and the value there and just taking the entire thing as the key. I'm not entirely sure how to solve that.

scratch/copy-mk.py
james-smith-za: I mean this was just a draft PR at this stage, I hadn't really thought this through completely.

But thanks for the tip!

james-smith-za:
OK belay that, I managed to get `sim_correlator.py` to parse the option correctly, but now I am getting this:
```
Error: Failed to process config: {'data_timeout': '0'} is not valid under any of the given schemas

Failed validating 'oneOf' in schema['properties']['config']['properties']['develop']:
    {'default': False,
     'oneOf': [{'additionalProperties': False,
                'properties': {'any_gpu': {'default': False,
                                           'type': 'boolean'},
                               'data_timeout': {'$ref': '#/definitions/nonneg_number'},
                               'disable_ibverbs': {'default': False,
                                                   'type': 'boolean'},
                               'less_resources': {'default': False,
                                                  'type': 'boolean'}},
                'type': 'object'},
               {'default': False, 'type': 'boolean'}]}

On instance['config']['develop']:
    {'data_timeout': '0'}
```

Which ...  doesn't make sense to me, because it looks as though `data_timeout` is there as a non-negative number, so I am not sure what it's confused about with the zero. Unless there's a type issue in the controller something and it's not converting the string to a number?

bmerry:
> Which ... doesn't make sense to me, because it looks as though data_timeout is there as a non-negative number

No, it's a string, which is a different thing from a number (in JSON). sim_correlator needs to do the str -> float conversion.

james-smith-za:
> No, it's a string, which is a different thing from a number (in JSON). sim_correlator needs to do the str -> float conversion.

Thanks for the tip. Turns out my mental model for how JSON worked was wrong.

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+    if args.develop is not None:
+        out_cmd.append(f"--develop={args.develop}")
+
+    if args.image_override is not None:
+        for override in args.image_override:
+            out_cmd.append(f"--image-override={override}")
+
+    out_cmd.append("cbf-mc.cbf.mkat.karoo.kat.ac.za")
+
+    print(f"Executing: {out_cmd}", "\n")
```
bmerry, line 145:
```suggestion
    print(f"Executing: {out_cmd}\n")
```

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+        out_cmd.append(f"--{k}={v}")
+
+    if args.develop is not None:
+        out_cmd.append(f"--develop={args.develop}")
+
+    if args.image_override is not None:
+        for override in args.image_override:
+            out_cmd.append(f"--image-override={override}")
+
+    out_cmd.append("cbf-mc.cbf.mkat.karoo.kat.ac.za")
```
bmerry, line 143: Since the CMC address is a command-line option, I think this should be too.

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        return 1
+
+    out_kwargs = {
+        "name": target.name.replace(".", "_"),
+        "antennas": target.n_antennas,
+        "channels": channels,
+        "digitiser-address": mcast_groups[0].split("+")[0],
+        "sync-time": sync_time,
+        "band": band,
```
bmerry, line 126: I take it --beams is being left for later?

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+    # connect to the corr2_servlet itself to see the sync time, n_chans and bandwidth
+    corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
+    sync_time = await corr2_client.sensor_value("sync-time", float)
+    channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
+    sample_rate = await corr2_client.sensor_value("antenna-channelised-voltage-bandwidth", float) * 2
+
+    # This won't distinguish between the different kinds of S-band. It'll be wrong. I'm not quite
+    # sure at this point how to match up the centre_frequency values in katgpucbf.meerkat.BANDS with
+    # what MK actually reports, even in the L and UHF cases, so for the purposes of getting
+    # something producing a result, I am leaving it as-is for the time being.
```
bmerry, line 111: I wouldn't worry about it too much for now - I'm not actually sure why we put band and centre frequency into the `dig.baseband_voltage` stream type, since we don't use them. Perhaps at the time I thought CBF could take over some of the delay manager functionality, which would require it. It's also needed if we're running Pipelines in the same subarray product (without CAM), or if we wanted to make fancier dsims that model geometric delays.

We might want to make those fields optional, at least for `dig.baseband_voltage`.

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+        "name": target.name.replace(".", "_"),
+        "antennas": target.n_antennas,
+        "channels": channels,
+        "digitiser-address": mcast_groups[0].split("+")[0],
+        "sync-time": sync_time,
+        "band": band,
+    }
+
+    out_cmd = [
+        "./sim_correlator.py",
```
bmerry, line 130: Not wild about a hardcoded path. Something like `os.path.join(os.path.dirname(__file__), "sim_correlator.py")` (untested!) would allow copy-mk.py to be run from wherever.

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env python3
+
+################################################################################
+# Copyright (c) 2024, National Research Foundation (SARAO)
```
bmerry, line 4:
```suggestion
# Copyright (c) 2024-2025, National Research Foundation (SARAO)
```

bmerry (changes requested):

scratch/copy-mk.py
```diff
@@ -0,0 +1,155 @@
...
+        target = subordinates[args.subordinate]
+    except KeyError:
+        print(f"{args.subordinate} is not among the subordinates currently running. Options are: {subordinates.keys()}")
+        return 1
+
+    # connect to the corr2_servlet itself to see the sync time, n_chans and bandwidth
+    corr2_client = await aiokatcp.Client.connect(args.cmc_host, target.control_port)
+    sync_time = await corr2_client.sensor_value("sync-time", float)
+    channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
+    sample_rate = await corr2_client.sensor_value("antenna-channelised-voltage-bandwidth", float) * 2
```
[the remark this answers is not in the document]
james-smith-za: I looked for an `adc-sample-rate` sensor previously but I couldn't find it for some reason. Now that I look again I see it's right there looking at me.

Yes, I am not supporting narrowband at this stage but it's going to be tougher because MK and MK+ have quite different models for how narrowband works. I think it probably can be done but I didn't want to bother for the sake of this exercise.

scratch/copy-mk.py
james-smith-za: I did sort of think of that but I couldn't see a straightforward way to figure out how many beams the MK correlator is making.

I guess it doesn't really matter though? But yeah for the purposes of this exercise I didn't think beams were needed yet. It was mainly at this stage to check that the routing works, and later on it's going to be used to check SDP while CAM may not yet be ready.

scratch/copy-mk.py
bmerry: Yes, exactly. I tend to run scratch scripts from the katgpucbf directory rather than inside scratch.

james-smith-za: > Not wild about a hardcoded path. Something like `os.path.join(os.path.dirname(__file__), "sim_correlator.py")` (untested!) would allow copy-mk.py to be run from wherever.

With this, I get `/home/jsmith/work/katgpucbf/scratch/./sim_correlator.py` as the path that it gives, which is perhaps not the most elegant, but it's valid so I'm going to use it.

scratch/copy-mk.py
```diff
@@ -126,9 +127,7 @@ async def async_main(args) -> int:
         "band": band,
     }
 
-    out_cmd = [
-        "./sim_correlator.py",
-    ]
+    out_cmd = [str(os.path.join(os.path.dirname(__file__), "sim_correlator.py"))]
```
bmerry, line 130: Why `str(...)`? os.path.join already returns a `str`.

scratch/copy-mk.py
```diff
@@ -56,6 +56,7 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
         help="Pass development options in the config. Use comma separation, or omit the arg to enable all.",
     )
     parser.add_argument("--image-override", action="append", metavar="NAME:IMAGE:TAG", help="Override a single image")
+    parser.add_argument("--controller", help="Hostname of the SDP master controller")
```
bmerry, line 59:
```suggestion
    parser.add_argument("--controller", help="Hostname of the CBF master controller")
```

bmerry: Either give it a default, make it a positional argument, or make it a required argument. Currently if the user doesn't pass it, you'll be trying to pass None to os.execv.

bmerry (changes requested):

scratch/copy-mk.py
james-smith-za: I ... don't remember why I did this.
I had some kind of trouble where it complained about types and it went away when I added `str` here but there may have been another issue elsewhere that has since been fixed.

Will remove.

scratch/copy-mk.py
james-smith-za: Good point. I didn't think of making it required.

I find more than one positional argument to be confusing, so I gave this one the `--` but it's a good point about the user not putting it in. I did encounter that NoneType issue but I realised what I'd done and just re-ran the script.

PR #902, opened 2025-01-22
Remove quantisation of narrowband centre frequency

bmerry:
- On the host, use Fraction instead of double precision, to eliminate
  rounding errors except where we explicitly choose to convert.
- On the device, use 64-bit fixed-point instead of 32-bit. This seems to
  have no noticeable performance impact (in some tests it seems to be
  faster!)

- [x] If qualification tests are changed: attach a sample qualification report: [report.pdf](https://github.com/user-attachments/files/18501302/report.pdf)

Closes NGC-1557.

### DIFF
--- a/doc/fgpu.design.rst
+++ b/doc/fgpu.design.rst
@@ -695,10 +695,17 @@ the fractional part. Even passing :math:`f` in single precision can lead
 to large errors.
 
 To avoid these problems, fixed-point computations are used. Phase is
-represented as a fractional number of cycles, scaled by :math:`2^{32}` and
-stored in a 32-bit integer. When performing arithmetic on values encoded this
+represented as a fractional number of cycles, scaled by :math:`2^{64}` and
+stored in a 64-bit integer. When performing arithmetic on values encoded this
 way, the values may overflow and wrap. The high bits that are lost represent
 complete cycles, and so have no effect on phase.
+
+This is sufficient to avoid any rounding errors inside the kernel, where
+:math:`t` is limited to the number of samples processed by a single kernel
+invocation. However, over much longer timescales (days) even quantising the
+frequency to 64-bit fixed point could lead to a small phase drift. The Python
+code thus uses the :class:`~fractions.Fraction` class to perform exact
+arithmetic until after it has subtracted the whole cycles.
 
 Delays
 ^^^^^^
--- a/qualification/antenna_channelised_voltage/test_narrowband.py
+++ b/qualification/antenna_channelised_voltage/test_narrowband.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2024, National Research Foundation (SARAO)
+# Copyright (c) 2024-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -16,6 +16,7 @@
 
 """Test specific features of narrowband."""
 
+import numpy as np
 import pytest
 
 from katgpucbf.meerkat import BANDS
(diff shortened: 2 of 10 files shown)


qualification/antenna_channelised_voltage/test_narrowband.py
```diff
@@ -27,18 +28,14 @@ def test_narrowband_centre_frequency() -> None:
...
+    resolution is thus determined by float-point accuracy and depends on
+    the magnitude of the value. The magnitude is limited to 856 MHz
+    (since the value is specified at baseband) and at this magnitude,
+    1 ULP is approximately :math:`1.2\times 10^{-7}` Hz.
     """
     for band in ["u", "l"]:
-        assert BANDS[band].adc_sample_rate / 2**32 / 16 < 100.0
+        max_centre_frequency = BANDS[band].adc_sample_rate / 2
+        resolution = max_centre_frequency - np.nextafter(max_centre_frequency, 0.0)
+        assert resolution < 100.0
```
amishatishpatel, line 41: Aha, the `100.0` is from the requirement itself. TIL about `np.nextafter`!

src/katgpucbf/fgpu/ddc.py
```diff
@@ -203,41 +204,42 @@ def __init__(self, template: DDCTemplate, command_queue: AbstractCommandQueue, s
...
         assert weights.shape == self._weights_host.shape
+        # Passing a float implies that it has not been computed exactly
+        assert isinstance(mix_frequency, Fraction)
+        self._mix_frequency = mix_frequency
         # Quantise the mixer frequency so that cycles per *output* sample are
-        # represented in fixed point with 32 fractional bits.
-        self._mix_scale = round(mix_frequency * self.template.subsampling * 2**32)
-        self._weights_host[:] = weights * np.exp(2j * np.pi * self.mix_frequency * np.arange(len(weights)))
+        # represented in fixed point with 64 fractional bits.
+        self._mix_scale = round(mix_frequency * self.template.subsampling * 2**64) % 2**64
```
amishatishpatel, line 224: Ah, I see you also moved the `mod` up here.

src/katgpucbf/fgpu/engine.py
```diff
@@ -494,7 +495,7 @@ def __init__(self, output: Output, engine: "Engine", context: AbstractContext, d
         if isinstance(output, NarrowbandOutput):
             narrowband_config = NarrowbandConfig(
                 decimation=output.decimation,
-                mix_frequency=-output.centre_frequency / engine.adc_sample_rate,
+                mix_frequency=-Fraction(output.centre_frequency) / Fraction(engine.adc_sample_rate),
```
amishatishpatel, line 498: Aha, and doing `Fraction(centre_frequency / adc_sample_rate)` would result in the exact thing you're trying to avoid by introducing this update.

test/fgpu/test_ddc.py
```diff
@@ -70,7 +76,7 @@ def test_ddc(
     rng = np.random.default_rng(seed=1)
     h_in = rng.integers(0, 256, (n_pols, samples * input_sample_bits // BYTE_BITS), np.uint8)
     weights = rng.uniform(-1.0, 1.0, (taps,)).astype(np.float32)
-    mix_frequency = 0.21
+    mix_frequency = Fraction("0.21")
```
amishatishpatel, line 79: **Interesting**. I forgot how fascinating the world of number accuracy is.
image

doc/fgpu.design.rst
```diff
@@ -695,11 +695,18 @@ the fractional part. Even passing :math:`f` in single precision can lead
...
-stored in a 32-bit integer. When performing arithmetic on values encoded this
+represented as a fractional number of cycles, scaled by :math:`2^{64}` and
+stored in a 64-bit integer. When performing arithmetic on values encoded this
 way, the values may overflow and wrap. The high bits that are lost represent
 complete cycles, and so have no effect on phase.
 
+This is sufficient to avoid any rounding errors inside the kernel, where
+:math:`t` is limited to the number of samples processed by a single kernel
+invocation. However, over much longer timescales (days) even quantising the
+frequency to 64-bit fixed point could lead to a small phase drift. The Python
```
amishatishpatel, line 706: Just out of interest, Re: Longer timescales, is that a probable/likely use case/scenario?

amishatishpatel (approved):
Very cool.

qualification/antenna_channelised_voltage/test_narrowband.py
```diff
@@ -27,18 +28,14 @@ def test_narrowband_centre_frequency() -> None:
...
-
-    The interfaces which set and report the centre frequency use IEEE-754
-    double precision, which has significantly more precision than this and
-    is not the limiting factor.
+    Verified by analysis. The F-engine converts the text representation
+    of the centre frequency to double-precision floating point. The
+    resolution is thus determined by float-point accuracy and depends on
+    the magnitude of the value. The magnitude is limited to 856 MHz
+    (since the value is specified at baseband) and at this magnitude,
+    1 ULP is approximately :math:`1.2\times 10^{-7}` Hz.
```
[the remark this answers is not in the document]
bmerry: https://en.wikipedia.org/wiki/Unit_in_the_last_place

doc/fgpu.design.rst
bmerry: The mixer phase is based on the ADC timestamp, which is only reset every day or two with a global sync. It's 48-bit, so at the end of that time, the error will be on the order of 2^48 / 2^64 = 2^-16. That's probably not enough to actually worry about, but it's still much worse than fp32 accuracy.

PR #903, opened 2025-01-22
Some small improvements to copy-mk.py

bmerry:
- Help for --controller: SDP -> CBF
- Make error when subordinate is not known more readable (instead of mentioning dict_keys) and print it to stderr instead of stdout
- Fix a double message in message about dummy antenna
- Simplify the code for constructing the command line
- Fix removal of the first argument, which prevented the subarray product from inheriting the name from the subordinate it duplicates

Relates to NGC-1551

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -57,7 +57,7 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     )
     parser.add_argument("--image-override", action="append", metavar="NAME:IMAGE:TAG", help="Override a single image")
     parser.add_argument(
-        "--controller", default="cbf-mc.cbf.mkat.karoo.kat.ac.za", help="Hostname of the SDP master controller"
+        "--controller", default="cbf-mc.cbf.mkat.karoo.kat.ac.za", help="Hostname of the CBF master controller"
     )
     args = parser.parse_args(argv)
     return args
@@ -89,7 +89,7 @@ async def async_main(args) -> int:
         for n, grp in enumerate(mcast_groups[1:]):
             if grp == mcast_groups[0]:
                 # MK is not using a full power-of-2 size array. There are dummies from here onwards.
-                print("Dummy antenna found starting at ", (n + 1) // 2)
+                print(f"Dummy antenna found starting at {(n + 1) // 2}")
                 n_antennas = (n + 1) // 2
                 break
         digitiser_address = mcast_groups[0].split("+")[0]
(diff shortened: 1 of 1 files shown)


scratch/copy-mk.py
```diff
@@ -89,7 +89,7 @@ async def async_main(args) -> int:
         for n, grp in enumerate(mcast_groups[1:]):
             if grp == mcast_groups[0]:
                 # MK is not using a full power-of-2 size array. There are dummies from here onwards.
-                print("Dummy antenna found starting at ", (n + 1) // 2)
+                print(f"Dummy antenna found starting at {(n + 1) // 2}")
```
james-smith-za, line 92: Is there something inherently preferable about an f-string?

scratch/copy-mk.py
```diff
@@ -127,24 +131,20 @@ async def async_main(args) -> int:
...
     if args.image_override is not None:
         for override in args.image_override:
             out_cmd.append(f"--image-override={override}")
 
     out_cmd.append(args.controller)
 
-    print(f"Executing: {out_cmd}", "\n")
-    os.execv(out_cmd[0], out_cmd[1:])
+    print(f"Executing: {out_cmd}\n")
+    os.execv(out_cmd[0], out_cmd)
```
james-smith-za, line 147: > - Fix removal of the first argument, which prevented the subarray
  product from inheriting the name from the subordinate it duplicates

Is this what you are referring to here?

james-smith-za: Ah, I see in the docs 
> In either case, the arguments to the child process should start with the name of the command being run, but this is not enforced.

I hadn't noticed that previously.

is `os.execv()` passing the args via a shell like bash, or is this the list like one would expect in a C context, `argv[0]`?

james-smith-za (approved):
FWIW I am happy to merge, the above comments are more for my own understanding, rather than actual issues with the changes.

scratch/copy-mk.py
bmerry: I was undecided between doing it this way and just deleting the trailing space of the string literal, but I figure with f-strings it's harder to screw up the spacing.

scratch/copy-mk.py
bmerry: execv doesn't involve a shell. out_cmd[0] becomes argv[0] in the invoked process.

james-smith-za: That does make sense.

scratch/copy-mk.py
james-smith-za: That's a fair point.

PR #904, opened 2025-01-24
Update .readthedocs.yml to make it build again

bmerry:
See https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/


amishatishpatel (approved):
Fair enough.

PR #906, opened 2025-01-28
Restrict pytest-check to 2.2.2

bmerry:
Newer versions of pytest-check don't play nice with pytest-reportlog: https://github.com/okken/pytest-check/issues/173


amishatishpatel (approved):
Good catch. Although, why did it only (seem to) manifest itself so late in the qualification run?

bmerry:
> Good catch. Although, why did it only (seem to) manifest itself so late in the qualification run?

Probably only happens when there is a actually a failure in a pytest_check assertion.

PR #910, opened 2025-02-10
Add tx.next-timestamp sensors for XBEngine

bmerry:
These sensors are a past-the-end timestamp for the transmit stream. Care is taken to update them in the same event loop iteration that send-enabled is checked, which means that doing:
```
?capture-stop stream
?sensor-value stream.tx.next-timestamp
?capture-start stream
```
will give a timestamp that can be used to distinguish the data from the two capture sessions (it will be greater than any timestamp from the former and less than or equal to any timestamp from the latter). This will be useful in fixing NGC-1542, which appears to be caused by use of data from a previous capture session.

There will be a matching PR coming for katsdpcontroller to rename the sensors suitably.

### DIFF
--- a/src/katgpucbf/utils.py
+++ b/src/katgpucbf/utils.py
@@ -21,6 +21,7 @@ import enum
 import gc
 import ipaddress
 import logging
+import math
 import signal
 import time
 import weakref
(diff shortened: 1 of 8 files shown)


src/katgpucbf/utils.py
```diff
@@ -169,6 +170,33 @@ def func(value: str) -> list[_T]:  # noqa: D102
...
+        name,
+        description,
+        units,
+        default,
+        initial_status,
+        *args,
+        auto_strategy=aiokatcp.SensorSampler.Strategy.EVENT_RATE,
+        auto_strategy_parameters=(MIN_SENSOR_UPDATE_PERIOD, math.inf),
+        **kwargs,
+    )
```
amishatishpatel, line 197: Noted all around. I guess this helper was warranted because the sensor declarations were getting lengthy, so I certainly appreciate it.

Perhaps rename it to e.g. `make_rate_limited_sensor`? 
* At the moment it's used like a Class instantiation (in the various replacements), and
* Something doesn't sit right with me Re: Underscores instead of camel-case
* Also, that's the convention we've used in e.g. recv.py (`make_stream`)

test/xbgpu/test_bsend.py
```diff
@@ -62,6 +62,7 @@ def sensors(outputs: Sequence[BOutput]) -> SensorSet:
     sensors = SensorSet()
     for output in outputs:
         sensors.add(Sensor(int, f"{output.name}.beng-clip-cnt", "Number of output samples that are saturated."))
+        sensors.add(Sensor(int, f"{output.name}.tx.next-timestamp", "Timestamp that has not yet been sent"))
```
amishatishpatel, line 65: Aha, the BPipeline updates this in BSend. Whereas the XPipeline updates the corresponding sensor in the sender-loop.

Nice one.

amishatishpatel (changes requested):
Just one explicit change requested (helper function rename).

Otherwise, looks good.

test/xbgpu/test_bsend.py
bmerry: Yeah. Given unlimited time, I'd like to move more of that logic into XSend, so that it's easier to test in isolation.

src/katgpucbf/utils.py
bmerry: Fair enough - it's probably better for function names to contain a verb. I've also renamed `steady_state_timestamp_sensor` from the same file for consistency.

amishatishpatel (approved):
Please and thank you.

PR #911, opened 2025-02-10
Replace pydocstyle and flake8 with ruff

bmerry:
pydocstyle is no longer maintained, and doesn't support PEP 695 syntax
for generic functions and classes.

This required a few changes:

- Move some noqa comments to a different line
- Fix a few docstring grammar issues that pydocstyle had missed.
- In exception handlers, raise ... from None (for whatever reason
  flake8-bugbear hadn't picked these up). This is a semantic change, but
  since FailReply is normally caught anyway I don't expect this to be
  visible.
- Import some things from collections.abc instead of typing (pyupgrade)
- Not entirely related, but strip out the exclude= from [tool.black]
  since pre-commit will only pass files in git to black (and AFAIK black
  is also git-aware and will ignore non-git files).


james-smith-za:
I'm feeling somewhat hard done by that it's only after I leave ruff gets its day in the sun...

pyproject.toml
```diff
@@ -9,28 +9,32 @@ build-backend = "setuptools.build_meta"
...
+    "N",  # pep8-naming
+    "B",  # flake8-bugbear
+]
+extend-ignore = ["D105", "D107"]  # docstrings for magic methods
+[tool.ruff.lint.extend-per-file-ignores]
+"scratch/**/*.py" = ["D"]  # Don't fuss with docstrings in scratch/
+# Temporary, until the code can be cleaned up
+# (this is what pydocstyle used to do)
+"test_*.py" = ["D"]
+"test/**/*.py" = ["D"]
```
amishatishpatel, line 34: In my mind, I thought we were on top of docstrings in unit tests. Then I re-read the comment and noted
> (this is what pydocstyle used to do)

src/katgpucbf/utils.py
```diff
@@ -155,7 +155,7 @@ def comma_split(
         times.
     """
 
-    def func(value: str) -> list[_T]:  # noqa: D102
+    def func(value: str) -> list[_T]:
```
amishatishpatel, line 158: Why does this function lose the `D102` marker entirely, but `update_aggregate` (L191) retains it?

amishatishpatel: Ah, looking at it again, and re-reading https://www.pydocstyle.org/en/2.1.1/error_codes.html
* Perhaps it's because `func` is not a public method?

src/katgpucbf/xbgpu/xsend.py
```diff
@@ -119,7 +119,7 @@ def timestamp(self) -> int:  # noqa: D102
         return int(self._timestamp[()])
 
     @timestamp.setter
-    def timestamp(self, value: int) -> None:  # noqa: D102
+    def timestamp(self, value: int) -> None:
```
amishatishpatel, line 122: I see the `noqa` moniker is still on the `@property` above.
* Does it now hold for all subsequent/related usages of the property?
* i.e. Is this `noqa` removed because it's mentioned elsewhere?

amishatishpatel (commented):
A couple comments for clarification before I explicitly Approve. I should have a look at the built docs for this branch as well.

pyproject.toml
bmerry: See NGC-1465.

src/katgpucbf/utils.py
bmerry: Yes, I think it was a bug in pydocstyle that it required this. Or possibly someone pro-actively added the noqa when it wasn't needed.

One of ruff's optional warnings is to complain about an unneeded noqa. I turned that on to identify where I could delete. I didn't commit it, because it complained about all the noqa's in the test files where we told it not to check, but my intention is to enable it after NGC-1465 is addressed.

src/katgpucbf/xbgpu/xsend.py
bmerry: Yes ruff seems to be smarter about ignoring property setters.

amishatishpatel (approved):
Noted all around.

PR #912, opened 2025-02-17
Document no-discard DDC filter in signal path docs

bmerry:
See NGC-1399


amishatishpatel (approved):
Fine by me. Do you want Marcel to cast an eye on the explainer?

bmerry:
> Do you want Marcel to cast an eye on the explainer?

Nah, I don't think it's that critical.

PR #915, opened 2025-02-18
Use PEP 695 syntax for generic functions and classes

bmerry:
It's a little more readable and makes it clear where classes are generic.

Upgrade black since the previous version sometimes put line breaks in non-intuitive places with this syntax.

### DIFF
--- a/src/katgpucbf/fgpu/accum.py
+++ b/src/katgpucbf/fgpu/accum.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2024, National Research Foundation (SARAO)
+# Copyright (c) 2024-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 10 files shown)


src/katgpucbf/fgpu/accum.py
```diff
@@ -17,39 +17,34 @@
 """Compute statistics by bucketing into windows."""
 
 from dataclasses import dataclass
-from typing import Final, Generic, Protocol, Self, TypeVar
-
-_T_co = TypeVar("_T_co", covariant=True)
```
amishatishpatel, line 22: I don't pretend to understand covariance and contravariance, but is this "distinction" (?) not required anymore?

src/katgpucbf/fgpu/main.py
```diff
@@ -105,7 +103,7 @@ class _NarrowbandOutputDict(_OutputDict, total=False):
     pass_bandwidth: float
 
 
-def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str], None]) -> None:
+def _parse_stream[OD: _OutputDict](value: str, kws: OD, field_callback: Callable[[OD, str, str], None]) -> None:
```
amishatishpatel, line 106: I ~understand why you've done away with the underscore notation in previous cases. Here, however, still refers to a class whose name contains it - `_OutputDict`. Either way, do you reckon that distinction isn't as important anymore?

src/katgpucbf/xbgpu/main.py
```diff
@@ -97,7 +96,7 @@ class _XOutputDict(_OutputDict, total=False):
     heap_accumulation_threshold: int
 
 
-def _parse_stream(value: str, kws: _OD, field_callback: Callable[[_OD, str, str], None]) -> None:
+def _parse_stream[OD: _OutputDict](value: str, kws: OD, field_callback: Callable[[OD, str, str], None]) -> None:
```
amishatishpatel, line 99: Same `_OutputDict` question here.

amishatishpatel (approved):
Couple comments for clarification. Looks great to me, might just take some getting used to (if/when I do need to adopt the `T` route).

src/katgpucbf/fgpu/accum.py
bmerry: One of the features of PEP 695 is that it automatically infers whether a generic variable is covariant and/or contravariant based on how it is used.

src/katgpucbf/fgpu/main.py
bmerry: The difference is that `_OutputDict` exists at module scope, so it could be referenced from other modules and the `_` prefix is a warning to those modules not to do that. The `OD` here doesn't exist outside of this function, so it is already private and doesn't need a prefix to tell other users "hands off" because they can't refer to it anyway.

PR #917, opened 2025-02-18
Fix race conditions in capture-start tests

bmerry:
- Set a minimum timestamp to capture, to ensure that we're not capturing data from the end of a previous test.
- Ensure that the dsim signal has reached the tx end of the affected engine, rather than just the start (for the ACV test, we just check that it's been received at the XB-engine rx, since I haven't bothered to add a tx.next-timestamp to the F-engine). Failing to do this won't cause the test to fail, but it will affect the validity since otherwise the steady-state-timestamp could be less than the dsim timestamp and the test won't prove anything.

Closes NGC-1542.

- [x] If qualification tests are changed: attach a sample qualification report: [report-NGC-1542.pdf](https://github.com/user-attachments/files/18845557/report-NGC-1542.pdf)

### DIFF
--- a/qualification/antenna_channelised_voltage/test_gain.py
+++ b/qualification/antenna_channelised_voltage/test_gain.py
@@ -136,18 +136,22 @@ async def test_gains_capture_start(
     dsim_timestamp = await pcc.sensor_value(f"{cbf.dsim_names[0]}.steady-state-timestamp", int)
     pdf_report.detail(f"Set dsim signals to {signals}, starting with timestamp {dsim_timestamp}.")
 
-    pdf_report.step("Wait for injected signal to reach F-engine.")
+    pdf_report.step("Wait for injected signal to reach XB-engines.")
     label = receiver.input_labels[0]
     for _ in range(10):
-        rx_timestamp = await pcc.sensor_value(f"antenna-channelised-voltage.{label}.rx.timestamp", int)
+        tasks = []
+        async with asyncio.TaskGroup() as tg:
+            for i in range(receiver.n_xengs):
+                tasks.append(tg.create_task(pcc.sensor_value(f"baseline-correlation-products.{i}.rx.timestamp")))
+        rx_timestamp = min(task.result() for task in tasks)
         pdf_report.detail(f"rx.timestamp = {rx_timestamp}")
-        if rx_timestamp >= dsim_timestamp:
+        if rx_timestamp > dsim_timestamp:
             break
         else:
             pdf_report.detail("Sleep for 0.5s.")
             await asyncio.sleep(0.5)
     else:
-        pytest.fail("Digitiser signal did not reach F-engine.")
+        pytest.fail("Digitiser signal did not reach XB-engines in time.")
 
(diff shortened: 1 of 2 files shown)


qualification/antenna_channelised_voltage/test_gain.py
```diff
@@ -158,7 +162,9 @@ async def test_gains_capture_start(
 
     pdf_report.step("Capture and verify output")
     await pcc.request("capture-start", "baseline-correlation-products")
-    _, data = await receiver.next_complete_chunk(min_timestamp=0)
+    # We use dsim_timestamp as a minimum to ensure that we're not receiving
+    # data from a *previous* capture-start/stop.
+    _, data = await receiver.next_complete_chunk(min_timestamp=dsim_timestamp)
```
amishatishpatel, line 167: Are you saying that each successive test is guaranteed to have dsim output timestamps increasing?
* How is that ensured? You can just point me to something/a file.

amishatishpatel:
I see you did not run `test_gain.py::test_gains_capture_start`. Have you tested it separately?

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -50,32 +50,34 @@ async def _test_capture_start(
...
     await prepare()
 
     pdf_report.step("Capture and verify output")
     async with asyncio.TaskGroup() as tg:
         for stream in receiver.stream_names:
             tg.create_task(pcc.request("capture-start", stream))
-    _, data = await receiver.next_complete_chunk(min_timestamp=0)
+    # We use dsim_timestamp as a minimum to ensure that we're not receiving
+    # data from a *previous* capture-start/stop.
+    _, data = await receiver.next_complete_chunk(min_timestamp=dsim_timestamp)
```
amishatishpatel, line 80: Same question as earlier in acv/test-gain.

amishatishpatel (commented):
Comment for clarification within.

qualification/antenna_channelised_voltage/test_gain.py
bmerry: Not sure I follow - time only runs forwards (there is the special case that the timestamps roll over after 2^48 samples, but that takes longer than the entire qualification test run, never mind the duration of each subarray product).

bmerry:
> I see you did not run test_gain.py::test_gains_capture_start (in report-NGC-1542.pdf). Have you tested it separately?

Oh good point - let me check that.

qualification/antenna_channelised_voltage/test_gain.py
amishatishpatel: I was asking mainly in case something was being mocked out internally to force/ensure some expected timestamp value(s). Either way, noted.

bmerry:
Here's an updated qualification test report: [report-NGC-1542.pdf](https://github.com/user-attachments/files/18860215/report-NGC-1542.pdf)

amishatishpatel (approved):

PR #918, opened 2025-02-26
Grab input-labels from subarray for sim_correlator

amishatishpatel:
The parsing of the `input-labelling` sensor is a bit of a pain. Tried not to leave a mess while doing so.

Contributes to: NGC-1588.

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -115,7 +115,7 @@ async def async_main(args) -> int:
     # This returns a string representation of a list of tuples, with each item
     # in the format (input-label, input-index, LRU host, index-on-host).
     input_labelling_str = await corr2_client.sensor_value("input-labelling", str)
-    input_labelling: list[tuple[str,]] = ast.literal_eval(input_labelling_str)
+    input_labelling: list[tuple[str, ...]] = ast.literal_eval(input_labelling_str)
     input_labels = [label[0] for label in input_labelling]
 
     # This won't distinguish between the different kinds of S-band. It'll be wrong. I'm not quite
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -115,7 +115,7 @@ async def async_main(args) -> int:
     # This returns a string representation of a list of tuples, with each item
     # in the format (input-label, input-index, LRU host, index-on-host).
     input_labelling_str = await corr2_client.sensor_value("input-labelling", str)
-    input_labelling: list[tuple[str, ...]] = ast.literal_eval(input_labelling_str)
+    input_labelling: list[tuple[str, int, str, int]] = ast.literal_eval(input_labelling_str)
     input_labels = [label[0] for label in input_labelling]
 
     # This won't distinguish between the different kinds of S-band. It'll be wrong. I'm not quite
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -111,6 +111,10 @@ async def async_main(args) -> int:
     sync_time = await corr2_client.sensor_value("sync-time", float)
     channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
     sample_rate = await corr2_client.sensor_value("adc-sample-rate", float)
+    # This returns a string representation of a list of tuples, with each item
+    # in the format (input-label, input-index, LRU host, index-on-host).
+    input_labelling_str = await corr2_client.sensor_value("input-labelling", str)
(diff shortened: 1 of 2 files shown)


scratch/sim_correlator.py
```diff
@@ -38,10 +38,15 @@
 
 def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     """Parse the command-line arguments (which may be specified as a parameter)."""
+
+    def _parse_input_labels(value: str) -> list[str]:
+        return value.split(",")
```
amishatishpatel, line 43: I will admit, I was taking inspiration from our/katgpucbf's main.py parsing of e.g. `--wideband`. Happy to remove/update/tweak.

scratch/sim_correlator.py
```diff
@@ -96,6 +101,9 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
         args.digitisers = args.antennas
     if args.digitiser_address is not None and args.sync_time is None:
         parser.error("--sync-time is required when specifying --digitiser-address")
+    if args.input_labels is None:
+        print("No input labels received")
```
amishatishpatel, line 105: I see I left this print statement in. Would you like me to remove? Or move elsewhere? I'm not sure if it's important to note.

scratch/copy-mk.py
```diff
@@ -111,6 +111,10 @@ async def async_main(args) -> int:
     sync_time = await corr2_client.sensor_value("sync-time", float)
     channels = await corr2_client.sensor_value("antenna-channelised-voltage-n-chans", int)
     sample_rate = await corr2_client.sensor_value("adc-sample-rate", float)
+    # This returns a string representation of a list of tuples, with each item
+    # in the format (input-label, input-index, LRU host, index-on-host).
+    input_labelling_str = await corr2_client.sensor_value("input-labelling", str)
+    input_labels_list = [labels_tuple.split(",")[0][1:-1] for labels_tuple in input_labelling_str[1:-1].split("(")[1:]]
```
bmerry, line 117: Rather than trying to parse the string yourself, you can let Python do it. Import `ast`, then something like
```
input_labelling = ast.literal_eval(input_labelling_str)
input_labels_list = [label[0] for label in input_labelling]
```

scratch/sim_correlator.py
bmerry: I'm happy with that approach, but make this a top-level function rather than nesting it here (PEP 20: "Flat is better than nested.")

scratch/sim_correlator.py
bmerry: I think take it out.

scratch/copy-mk.py
bmerry: I'd also suggest just calling this `input_labels`. We don't generally bundle the type of a variable into the name unless it's needed for disambiguation or it's unexpected - for a variable with a plural name it's not unexpected that it's a list.

bmerry (changes requested):

scratch/copy-mk.py
amishatishpatel: Apologies, you are quite right. I had `input_labels_str` underneath it before I moved the `",".join(...)` to be in-line in the config dict.

scratch/copy-mk.py
```diff
@@ -114,7 +115,8 @@ async def async_main(args) -> int:
     # This returns a string representation of a list of tuples, with each item
     # in the format (input-label, input-index, LRU host, index-on-host).
     input_labelling_str = await corr2_client.sensor_value("input-labelling", str)
-    input_labels_list = [labels_tuple.split(",")[0][1:-1] for labels_tuple in input_labelling_str[1:-1].split("(")[1:]]
+    input_labelling: list[tuple[str,]] = ast.literal_eval(input_labelling_str)
```
bmerry, line 118:
```suggestion
    input_labelling: list[tuple[str, ...]] = ast.literal_eval(input_labelling_str)
```
(what you wrote is a list of tuples with each tuple containing exactly one str)

bmerry (changes requested):
Just one tweak to type annotations.

scratch/copy-mk.py
amishatishpatel: Oh gosh, apologies. I started writing that type hint then got sidetracked when we started discussing the input labels.

Fixing now. For the record, it's in the format (string, int, string, int).
image

bmerry: Oh good point, it's not all strings. So `list[tuple[str, int, str, int]]` then.

bmerry (approved):
LGTM

PR #922, opened 2025-03-04
Support real-time transfer of delays from MK to MK+ CBF

amishatishpatel:
Update copy-mk.py to facilitate the real-time transfer of delay values from MK
to MK+ CBF. This queries the equivalent sensors on the MK subarray but **does not**
use dummy antennas.

Contributes to: NGC-1588.

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -126,7 +126,17 @@ def delay_transfer_callback(
 
 
 async def async_main(args) -> int:
-    portal_client = KATPortalClient(args.portal, None)
+    # This queue is used in the case of --transfer-delays
+    # but is needed by the callback below when instantiating
+    # the portal client.
+    delays_queue: asyncio.Queue[tuple[int, str, str]] = asyncio.Queue(maxsize=500)
+    portal_client = KATPortalClient(
+        args.portal,
+        partial(
+            delay_transfer_callback,
+            delays_queue=delays_queue,
+        ),
+    )
     # Get name of the CBF proxy e.g. "cbf_1"
     prefix = await portal_client.sensor_subarray_lookup("cbf", None)
     cbf_sensor_names = [
@@ -150,9 +160,6 @@ async def async_main(args) -> int:
         return cbf_sensor_values[f"{prefix}_{name}"]
 
     input_labels: list[str] = cbf_sensor_value("input_labels").split(",")
-    # Create asyncio Queue for the delay-callback to push items onto
-    delays_queue: asyncio.Queue[tuple[int, str, str]] = asyncio.Queue(maxsize=len(input_labels) * 8)
-
     # Get the multicast groups. This is an annoying sensor because it's not
     # valid Python or JSON: it's a comma-separated list surrounded by brackets,
     # but the strings are not quoted.
@@ -256,7 +263,7 @@ async def async_main(args) -> int:
     if args.transfer_delays:
         qualified_labels = [f"antenna_channelised_voltage_{input_label}" for input_label in input_labels]
         # New client because it needs to be initialised with the callback (?)
-        callback_client = KATPortalClient(
+        portal_client = KATPortalClient(
             args.portal,
             partial(
                 delay_transfer_callback,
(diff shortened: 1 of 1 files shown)


scratch/copy-mk.py
```diff
@@ -185,6 +232,71 @@ def cbf_sensor_value(name: str) -> Any:
...
+                    delay_args[delays_queue_item[1]] = delays_queue_item
+                    complete_requests = all(delay_args_value is not None for delay_args_value in delay_args.values())
+
+                unix_timestamp = ref_loadmcnt / scale_factor_timestamp + sync_time
+                sorted_delay_args = dict(sorted(delay_args.items()))
+                await pc_client.request(
+                    "delays",
+                    "antenna-channelised-voltage",
+                    unix_timestamp,
+                    *[delay_arg_value[-1] for delay_arg_value in sorted_delay_args.values()],
```
amishatishpatel, line 292: This bit was a bit awkward to navigate. I could've done the `sorted(...)` in-line but the list comprehension wouldn't be that easy to comprehend. Happy to change if needed.

scratch/copy-mk.py
```diff
@@ -185,6 +230,71 @@ def cbf_sensor_value(name: str) -> Any:
...
+        delay_sensors_regex = (
+            "^(?:"
+            + "|".join(f"{prefix}_wide_{stream_name.replace("-", "_")}_delay" for stream_name in stream_names)
+            + ")$"
+        )
+        # TODO: Check against results returned from the following to
+        # ensure the `callback_client` is ready to proceed.
+        namespace = f"{args.name}_{uuid.uuid4()}"
+        await callback_client.subscribe(namespace)
+        await callback_client.set_sampling_strategies(namespace, delay_sensors_regex, "event")
```
amishatishpatel, line 255: I've got something locally to check the result here and print where it failed to subscribe to a sensor.

Should I make the script fail hard here if it can't configure all delay sensors? I genuinely don't know what length/breadth is required for tomorrow's integration test.

scratch/copy-mk.py
```diff
@@ -57,6 +69,11 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
         help="Endpoint of the CBF master controller",
     )
     parser.add_argument("--name", help="Name of the subarray product to create")
+    parser.add_argument(
+        "--product-controller",
+        type=endpoint_parser(31000),  # Not the actual default port
```
bmerry, line 74:
```suggestion
        type=endpoint_parser(None),
```
That will make it require a port IIRC.

scratch/copy-mk.py
```diff
@@ -66,6 +83,14 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     return args
 
 
+def _assemble_regex(names: Sequence[str]) -> str:
+    """Assemble a regex from a sequence of strings."""
+    regex = "|".join(re.escape(name) for name in names)
+    if not regex:
+        return {}  # Would get an error trying to query an empty regex
```
bmerry, line 90: That's not a valid return type for this function (I'm surprised mypy didn't tell you off - maybe it's not checking scratch/). Rather keep this logic in the caller.

scratch/copy-mk.py
```diff
@@ -89,34 +111,47 @@ async def multi_sensor_values(client: KATPortalClient, names: Sequence[str]) ->
...
     # phase-rate <radians per second>)
-    loadmcnt, delay, delay_rate, phase, phase_rate = delay_sensor_value.replace(" ", "").split(",")
-    request_args = f"{delay},{delay_rate}:{phase},{phase_rate[:-1]}"
-    request_tuple = (int(loadmcnt[1:]), src_name, request_args)
+    loadmcnt, delay, delay_rate, phase, phase_rate = ast.literal_eval(delay_sensor_value)
+    request_args = f"{delay},{delay_rate}:{phase},{phase_rate}"
+    # Example sensor name: cbf_1_wide_antenna_channelised_voltage_m000h_delay
+    input_label: str = msg_dict["msg_data"]["name"].split("_")[-2]
+    delay_queue_item = DelayQueueItem(
+        loadmcnt=int(loadmcnt),
```
bmerry, line 132: Is the cast to int still needed? ast.literal_eval probably gives you an int to start with.

bmerry (changes requested):

scratch/copy-mk.py
```diff
@@ -179,12 +259,58 @@ def cbf_sensor_value(name: str) -> Any:
...
+                    unix_timestamp,
+                    *sorted_delay_args,
+                )
+                delay_args = dict.fromkeys(input_labels)
+
+    except (aiokatcp.FailReply, ConnectionError) as exc:
+        print(f"Error: {exc}", file=sys.stderr)
+        return 1
+    finally:
+        portal_client.disconnect()
```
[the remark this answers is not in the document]
amishatishpatel: Fair point. I had a look inside [disconnect](https://github.com/ska-sa/katportalclient/blob/master/katportalclient/client.py#L510-L522) and it didn't strike me as something that would fail hard.

Still, probably best to check `if portal_client.is_connected:` first.

scratch/copy-mk.py
amishatishpatel: You may very well be right. I can't test at the moment because nothing is running. I'll remove the cast anyway.

bmerry (approved):

PR #924, opened 2025-03-04
copy-mk.py: add --sdp option

bmerry:
Currently it only has one valid value ("ingest") but I've made it an choice rather than just a boolean so that later we can add additional pieces of SDP.

See NGC-1588

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -77,6 +77,7 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
     parser.add_argument(
         "--transfer-delays", action="store_true", help="Continuously transfer F-engine delays from MK CBF to MK+ CBF"
     )
+    parser.add_argument("--sdp", choices=["ingest"], default=None, help="Configure SDP as well")
     parser.add_argument("--dry-run", action="store_true", help="Print config only")
     katgpucbf.configure_tools.add_arguments(parser)
     args = parser.parse_args(argv)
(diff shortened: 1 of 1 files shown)


scratch/copy-mk.py
```diff
@@ -166,13 +173,18 @@ async def async_main(args) -> int:
...
         "subarray_product_id",
     ]
-    cbf_sensor_values = await multi_sensor_values(
-        portal_client,
-        [f"{prefix}_{name}" for name in cbf_sensor_names],
-    )
+    sensor_names = [f"{prefix}_{name}" for name in cbf_sensor_names]
+    if args.sdp is not None:
+        sdp_prefix = await portal_client.sensor_subarray_lookup("sdp", None)
+        sdp_subarray_product_id = await single_sensor_value(portal_client, f"{sdp_prefix}_subarray_product_ids")
```
amishatishpatel, line 179: `_subarray_product_ids` plural?

scratch/copy-mk.py
```diff
@@ -89,6 +90,12 @@ def _assemble_regex(names: Sequence[str]) -> str:
     return f"^(?:{regex})$"  # Anchor the regex and return
 
 
+async def single_sensor_value(client: KATPortalClient, name: str) -> Any:
+    """Query the value of a single sensor from katportal."""
+    regex = "^" + re.escape(name) + "$"  # Anchor and escape the regex
+    return (await client.sensor_value(regex)).value
```
amishatishpatel, line 96: I didn't know you could wrap the await'd coroutine in brackets and treat it like a result...

amishatishpatel (approved):
One question on the subarray-product sensor for SDP. Otherwise, can't see too much to complain about.

scratch/copy-mk.py
bmerry: Yes - I have no idea why CAM named the sensor like this.

PR #927, opened 2025-03-06
Specify strict= parameter to zip() calls

bmerry:
Add target-version="py312" parameter to ruff configuration. This enables
it to give warnings that are not appropriate to older Python versions.
The primary is that it recommends specifying the 'strict' parameter to
zip() calls to indicate whether the inputs are expected to be the same
length. This is probably good documentation and a good sanity check so
I've done that.

- [x] If qualification tests are changed: attach a sample qualification report:
  - https://jenkins.cbf.mkat.karoo.kat.ac.za/job/qualification_katgpucbf/111 (had two failures, but since only one config is affected and the results are related to not receiving data, I'm assuming this is a network wobble rather than broken code)
  - https://jenkins.cbf.mkat.karoo.kat.ac.za/job/qualification_katgpucbf_lab/13 (still running)

### DIFF
--- a/qualification/tied_array_channelised_voltage/test_linearity.py
+++ b/qualification/tied_array_channelised_voltage/test_linearity.py
@@ -70,8 +70,8 @@ async def _test_linearity(
     assert scales[middle] == pytest.approx(1.0)
     powers = np.zeros_like(scales)
     for batch in batched(enumerate(scales), len(receiver.stream_names)):
-        # The last batch might be short (since the number of scales won't be
-        # multiple of the number of streams. When zipping with streams, we
+        # The last batch might be short (since the number of scales won't be a
+        # multiple of the number of streams). When zipping with streams, we
         # don't expect the iterables to be the same length, and some streams
         # will be unused.
         for (_, scale), beam_name in zip(batch, receiver.stream_names, strict=False):
--- a/qualification/tied_array_channelised_voltage/test_linearity.py
+++ b/qualification/tied_array_channelised_voltage/test_linearity.py
@@ -70,6 +70,10 @@ async def _test_linearity(
     assert scales[middle] == pytest.approx(1.0)
     powers = np.zeros_like(scales)
     for batch in batched(enumerate(scales), len(receiver.stream_names)):
+        # The last batch might be short (since the number of scales won't be
+        # multiple of the number of streams. When zipping with streams, we
+        # don't expect the iterables to be the same length, and some streams
+        # will be unused.
         for (_, scale), beam_name in zip(batch, receiver.stream_names, strict=False):
             await set_variable(client, beam_name, scale)
             pdf_report.detail(f"Set {variable} to {scale} on {beam_name}.")
--- a/qualification/tied_array_channelised_voltage/test_linearity.py
+++ b/qualification/tied_array_channelised_voltage/test_linearity.py
@@ -70,12 +70,12 @@ async def _test_linearity(
     assert scales[middle] == pytest.approx(1.0)
     powers = np.zeros_like(scales)
     for batch in batched(enumerate(scales), len(receiver.stream_names)):
-        for (_, scale), beam_name in zip(batch, receiver.stream_names):
+        for (_, scale), beam_name in zip(batch, receiver.stream_names, strict=False):
             await set_variable(client, beam_name, scale)
             pdf_report.detail(f"Set {variable} to {scale} on {beam_name}.")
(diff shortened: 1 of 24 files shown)


bmerry:
I've made this a draft because I still want to run a qualification test on it to ensure I didn't break anything.

test/fgpu/test_katcp_interface.py
```diff
@@ -218,7 +218,8 @@ def parse_delay_string(delay_str: str) -> tuple[float, float, float, float]:
             sensor_values_str = sensor_reading[1:-1].split(",")[1:]  # Drop the timestamp
             sensor_values = (float(field.strip()) for field in sensor_values_str)
 
-            for actual_value, expected_value in zip(sensor_values, parse_delay_string(correct_delay_strings[pol])):
+            expected_values = parse_delay_string(correct_delay_strings[pol])
+            for actual_value, expected_value in zip(sensor_values, expected_values, strict=True):
```
amishatishpatel, line 222: Any particular reason you introduced the intermediate variable?

src/katgpucbf/dsim/signal.py
amishatishpatel, line 1: I appreciate that this is a minor update/removal with a helpful commit message. Still, was a surprise to see this in the PR - would've appreciated a heads-up in the blurb.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -70,12 +70,12 @@ async def _test_linearity(
     assert scales[middle] == pytest.approx(1.0)
     powers = np.zeros_like(scales)
     for batch in batched(enumerate(scales), len(receiver.stream_names)):
-        for (_, scale), beam_name in zip(batch, receiver.stream_names):
+        for (_, scale), beam_name in zip(batch, receiver.stream_names, strict=False):
```
amishatishpatel, line 73: I think a comment like you've done in `acv/__init__.py` would be helpful here.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -70,12 +70,12 @@ async def _test_linearity(
...
             await set_variable(client, beam_name, scale)
             pdf_report.detail(f"Set {variable} to {scale} on {beam_name}.")
         _, data = await receiver.next_complete_chunk()
         pdf_report.detail("Received chunk.")
-        for (i, _), d, beam_name in zip(batch, data, receiver.stream_names):
+        for (i, _), d, beam_name in zip(batch, data, receiver.stream_names, strict=False):
```
amishatishpatel, line 78: Again, comment please.

amishatishpatel (changes requested):

test/fgpu/test_katcp_interface.py
bmerry: The line got rather long, and when black wrapped it, it became ugly.

qualification/tied_array_channelised_voltage/test_linearity.py
bmerry: Done.

qualification/tied_array_channelised_voltage/test_linearity.py
bmerry: Haven't added a separate one here - the one I've just added at the top of the for loop should cover both.

qualification/tied_array_channelised_voltage/test_linearity.py
```diff
@@ -70,6 +70,10 @@ async def _test_linearity(
     assert scales[middle] == pytest.approx(1.0)
     powers = np.zeros_like(scales)
     for batch in batched(enumerate(scales), len(receiver.stream_names)):
+        # The last batch might be short (since the number of scales won't be
+        # multiple of the number of streams. When zipping with streams, we
+        # don't expect the iterables to be the same length, and some streams
+        # will be unused.
```
amishatishpatel, line 76: Please and thank you. I have to unfortunately flag two things
1. I don't see a closing bracket for the second point in the first sentence
2. Adding to that, "since the number of scales won't be *a multiple [...]" ?

amishatishpatel (changes requested):
Sorry to be a pain.

qualification/tied_array_channelised_voltage/test_linearity.py
bmerry: Thanks. Guess I shouldn't have staying up working so late last night :-/

amishatishpatel (approved):

PR #929, opened 2025-03-10
qualification/Jenkinsfile improvements

bmerry:
Use the environment name to form the email subject (to distinguish lab from karoo).

Use extra_args to allow the user to specify extra command-line arguments for pytest.

Relates to NGC-1593.

- [x] If qualification tests are changed: attach a sample qualification report: see email notifications
- [x]  (n/a) If design has changed: ensure documentation is up to date

### DIFF
--- a/qualification/Jenkinsfile
+++ b/qualification/Jenkinsfile
@@ -33,6 +33,8 @@
  * - katsdpcontroller_image: image path for katsdpcontroller (for product controller)
  * - branch: branch of katgpucbf from which to obtain test code
  * - ini: relative path (from repo root) to pytest configuration file
+ * - extra_args: extra user-provided command-line arguments
+ * - environment: name for the test environment (for use in email notification)
  *
  * For more information on the qualification framework, see
  * katgpucbf/doc/qualification.rst
@@ -101,7 +103,8 @@ pipeline {
            --suppress-tests-failed-exit-code \
            --image-override katgpucbf:${katgpucbf_image} \
            --image-override katsdpcontroller:${katsdpcontroller_image} \
-           --junitxml=result.xml
+           --junitxml=result.xml \
+           ${extra_args}
          '''
        }
      }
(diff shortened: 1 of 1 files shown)


qualification/Jenkinsfile
```diff
@@ -171,7 +174,7 @@ pipeline {
         <br>
         <i>Note: This is an Automated email notification.</i>""",
         recipientProviders: [developers(), requestor()],
-        subject: 'Qualification Tests - katgpucbf - $BUILD_STATUS!',
+        subject: "${params.environment} Qualification Tests - katgpucbf - \$BUILD_STATUS!",
```
amishatishpatel, line 177: Why do you need to add the backslash before `$BUILD_STATUS!` ?

amishatishpatel: I'll also admit, I was expecting something more like 
* `"Qualification Tests - ${params.environment} - katgpucbf - \$BUILD_STATUS!"`
* i.e. The environment was another identifier in the "hierarchy"

I will concede that the whole point of this was to decipher Lab/Karoo as fast as possible
* Not to make it as aesthetically pleasing as possible

amishatishpatel (approved):
Big fan of `extra_args`.

qualification/Jenkinsfile
bmerry: > Why do you need to add the backslash before $BUILD_STATUS! ?

It's to ensure the right thing interprets it. Without the backslash, Groovy tries to interpolate the variable when it reads the Jenkinsfile - but BUILD_STATUS is not set as a Groovy variable (it can't be, since the job hasn't run yet). With the backslash, Groovy treats it as a literal `$`, but the emailext plugin substitutes the build status when it formulates the email.

bmerry: > I'll also admit, I was expecting something more like
>
>    "Qualification Tests - ${params.environment} - katgpucbf - \$BUILD_STATUS!"
>    i.e. The environment was another identifier in the "hierarchy"

It's not a hill I'd die on, but I was trying to make it somewhat like the description on the Jenkins page:

image

PR #932, opened 2025-03-12
copy-mk improvements

bmerry:
The main change is to add tracking of gains, and conversion to match the scale for MK. To do this, I made the katportalclient callback even simpler, just passing through the sensor name and value, with all the processing on it done in the main function.

A more minor change is to pass through a cam.http stream so that the SDP can query sensors from CAM.

Relates to NGC-1588.

### DIFF
--- a/scratch/copy-mk.py
+++ b/scratch/copy-mk.py
@@ -43,15 +43,20 @@ from katsdptelstate.endpoint import endpoint_parser
 
 import katgpucbf.configure_tools
 
+#: Scale factors between MK and MK+ wideband gains, based on number of channels
+GAIN_SCALE = {
+    1024: 5687.284403271646,
+    4096: 3096.9068735303676,
+    32768: 1147.5800238145268,
+}
+
 
 @dataclass
-class DelayQueueItem:
-    """Information gathered during delay sensor updates."""
+class SensorQueueItem:
+    """Information gathered during sensor updates."""
 
-    loadmcnt: int
-    input_label: str
-    #: In the format: delay,delay_rate:phase,phase_rate.
-    delay_request_args: str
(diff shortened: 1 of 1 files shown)


scratch/copy-mk.py
```diff
@@ -43,15 +43,20 @@
 
 import katgpucbf.configure_tools
 
+#: Scale factors between MK and MK+ wideband gains, based on number of channels
+GAIN_SCALE = {
+    1024: 5687.284403271646,
+    4096: 3096.9068735303676,
+    32768: 1147.5800238145268,
+}
+
```
amishatishpatel, line 52: Is this from any particular file/source of truth?

amishatishpatel (approved):
I gather this was used during the integration test yesterday, so I'm not too fussed about verifying things myself with a live subarray - unless you'd prefer I do so?

I was hoping to use the walrus operator during my `--transfer-delays` update; glad it makes an appearance now!

scratch/copy-mk.py
bmerry: It's computed from https://docs.google.com/document/d/1TZWZATFNrwVG3ugBEe2aaQVnJDY_plfu8QSgN5sfGsk/edit?tab=t.0#heading=h.x68achryfapa - but based on Ludwig's preliminary analysis, the values might be wrong. I'm tracking things in NGC-1636.

amishatishpatel:
> The main function is getting a bit unwieldy so I might like to refactor it at some point, but I think it would be sensible to do it just prior to another test so that we can test the modifications.

You know, I had a similar point typed up in my review but removed it because I didn't want to get ahead of myself. Either way, noted.

PR #933, opened 2025-03-13
Enable ruff pyupgrade warnings and fix them

bmerry:

### DIFF
--- a/qualification/report/generate_pdf.py
+++ b/qualification/report/generate_pdf.py
@@ -30,8 +30,8 @@ import tempfile
 from collections import defaultdict
 from collections.abc import Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass, field
-from datetime import date, datetime, timezone
-from typing import Literal, Union
+from datetime import UTC, date, datetime
+from typing import Literal
 from uuid import UUID
 
 import docutils.writers.latex2e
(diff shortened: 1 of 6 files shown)


src/katgpucbf/fgpu/send.py
```diff
@@ -316,7 +316,7 @@ def make_streams(
         # Adding len(endpoints) to accommodate descriptors sent for each substream
         max_heaps=n_data_heaps + len(endpoints),
     )
-    streams: list["spead2.send.asyncio.AsyncStream"]
```
amishatishpatel, line 319: Do you remember why we had to insert quotes in the first place?

amishatishpatel (approved):
Syntax will take some getting used to. Also, even though we did discuss it, still surprised this was the extend of the updates required.

qualification/report/generate_pdf.py
```diff
@@ -193,7 +193,7 @@ class Figure:
     type: str
 
 
-Item = Union[Detail, Failure, Figure, RawFigure]  # Any information provided inside a ``pdf_report.step``
+type Item = Detail | Failure | Figure | RawFigure  # Any information provided inside a ``pdf_report.step``
```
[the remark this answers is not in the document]
bmerry: It's pretty new (PEP 695, introduced in Python 3.12).

src/katgpucbf/fgpu/send.py
bmerry: That type is a fiction of the spead2 type annotations and doesn't exist at runtime. Python is a little inconsistent about where one needs to quote such fictional types. In function signatures you need the quotes as otherwise it tries to look up the type at runtime, but for local variables in a function you don't need it (I think that's because you can interrogate a function's signature at runtime using the `inspect` module). So I don't think these quotes were ever needed, but I got into the habit of putting them in rather than try to remember where they're required.

PR #943, opened 2025-05-07
Reduce karoo qualification run time

amishatishpatel:
This PR trims the `FULL_ANTENNAS` parameter list in qualification/conftest.py and introduces a new 
pytest.ini config option for limiting the decimation values used for VLBI-specific tests.

- [x] If qualification tests are changed: attach a sample qualification report: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_karoo/177/

Closes NGC-1671.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -95,6 +95,12 @@ ini_options = [
         type="args",
         default=["8"],
     ),
+    IniOption(
+        name="vlbi_decimation",
+        help="Space-separated list of VLBI narrowband decimation factors to test",
+        type="args",
+        default=["8"],
+    ),
     IniOption(name="bands", help="Space-separated list of bands to test", type="args", default=["l"]),
     IniOption(name="beams", help="Number of beams to produce", type="string", default="4"),
     IniOption(name="raw_data", help="Include raw data for figures", type="bool", default=False),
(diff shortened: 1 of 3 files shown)


qualification/conftest.py
```diff
@@ -44,7 +44,7 @@
 from .reporter import Reporter, custom_report_log
 
 logger = logging.getLogger(__name__)
-FULL_ANTENNAS = [1, 4, 8, 10, 16, 20, 32, 40, 55, 64, 65, 80]
+FULL_ANTENNAS = [4, 8, 20, 32, 40, 64, 80]
```
bmerry, line 47: Maybe let's leave 1 in: it can be a special case in some instances (no cross-correlations).

qualification/conftest.py
```diff
@@ -161,13 +167,24 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         metafunc.parametrize("n_antennas", values)
     if "band" in metafunc.fixturenames:
         metafunc.parametrize("band", metafunc.config.getini("bands"))
-    if "n_channels" in metafunc.fixturenames or "narrowband_decimation" in metafunc.fixturenames:
+    if (
+        "n_channels" in metafunc.fixturenames
+        or "narrowband_decimation" in metafunc.fixturenames
+        or "vlbi_decimation" in metafunc.fixturenames
```
bmerry, line 173: I don't think we have a *fixture* called vlbi_decimation, just an ini option.

qualification/pytest-jenkins-karoo.ini
```diff
@@ -22,4 +22,5 @@ max_antennas = 80
 wideband_channels = 1024 4096 8192 32768
 narrowband_channels = 32768
 narrowband_decimation = 8 16
+vlbi_decimation = 8
```
bmerry, line 25: Need to add it to pytest-jenkins-lab.ini as well. Also check if the documentation has an example ini file that needs updating - I seem to recall we removed the example and just referred the reader to the actual ini files for inspiration.

bmerry (changes requested):

qualification/conftest.py
amishatishpatel: Noted. Although, I didn't see a fixture called `narrowband_decimation` either. Thought this was more of a future-proofing/safety addition.

Also, I was thinking around "what would prompt this branch to execute". Either way, will revert.

qualification/pytest-jenkins-karoo.ini
amishatishpatel: Noted.

qualification/conftest.py
bmerry: It's not declared with `@pytest.fixture` (because it gets defined here), but it's used in parameter lists of other fixtures like `_cbf_config_and_description`.

bmerry (approved):
Approved, pending a successful run (can be merged even if we're still >8hrs).

qualification/pytest-jenkins-karoo.ini
amishatishpatel: For the record you were quite right Re:
> I seem to recall we removed the example and just referred the reader to the actual ini files for inspiration.

https://katgpucbf.readthedocs.io/en/latest/qualification.html#configuration

PR #951, opened 2025-06-19
Bump requests to 2.32.4

bmerry:
To keep dependabot happy. The actual CVE that is fixed (CVE-2024-35195) is a corner case that's unlikely to affect us.


amishatishpatel (approved):
I didn't even know that corner of GitHub existed. Appreciate you pointing me there as added reasoning for this version bump.

*That being said*, we were already on an 'unaffected' version (2.32.3 >= 2.32.0) - if I'm reading the info correctly?
image

But also, who am I to argue with dependabot.

PR #955, opened 2025-07-09
Tweak xbgpu benchmarks to use random input data 

amishatishpatel:
Taking inspiration from the fgpu's `compute_bench.py`.

Contributes to: NGC-1682.

### DIFF
--- a/scratch/xbgpu/benchmarks/correlate_bench.py
+++ b/scratch/xbgpu/benchmarks/correlate_bench.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 ################################################################################
-# Copyright (c) 2024, National Research Foundation (SARAO)
+# Copyright (c) 2024-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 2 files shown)


scratch/xbgpu/benchmarks/correlate_bench.py
```diff
@@ -52,7 +53,12 @@ def main():
     fn = template.instantiate(command_queue, args.heaps_per_fengine_per_chunk)
 
     fn.ensure_all_bound()
-    fn.buffer("in_samples").zero(command_queue)
+    buf = fn.buffer("in_samples")
+    h_data = buf.empty_like()
+    assert h_data.dtype == np.int8
+    rng = np.random.default_rng(seed=1)
+    h_data = rng.integers(-128, 127, size=h_data.shape, dtype=h_data.dtype)
```
bmerry, line 60: You're creating h_data, then discarding it to be garbage collected and replacing it with a different array. You could simplify things like so:
```suggestion
    assert buf.dtype == np.int8
    rng = np.random.default_rng(seed=1)
    h_data = rng.integers(-128, 127, size=h_data.shape, dtype=buf.dtype)
```

bmerry (changes requested):
Have commented on one file, but it applies equally to the other.

scratch/xbgpu/benchmarks/correlate_bench.py
amishatishpatel: Fair enough. I gather you meant the `size` arg to be `buf.shape`. Will update now-now.

bmerry: That would be fine too. Since it's done elsewhere let's go with that.

bmerry (approved):

PR #957, opened 2025-07-21
Update dithering seed selection in f/xbgpu

amishatishpatel:
Specifically, to address the concerns outlined in the ticket around unique seed values
for each/every engine launch.

- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match. See https://github.com/ska-sa/katsdpcontroller/pull/812.

Closes NGC-1418.

### DIFF
--- a/src/katgpucbf/__init__.py
+++ b/src/katgpucbf/__init__.py
@@ -1,7 +1,7 @@
 # noqa: D104
 
 ################################################################################
-# Copyright (c) 2020-2022, 2024, National Research Foundation (SARAO)
+# Copyright (c) 2020-2022, 2024-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy


src/katgpucbf/fgpu/engine.py
```diff
@@ -571,7 +574,7 @@ def __init__(self, output: Output, engine: "Engine", context: AbstractContext, d
             # The magic constant was chosen at random. It ensures that the
             # seed won't be the same as in other types of engine that also use
             # sync_time as the basis for seeding.
```
bmerry, line 576:
```suggestion
```

src/katgpucbf/__init__.py
```diff
@@ -42,6 +42,7 @@
 DEFAULT_JONES_PER_BATCH: Final = 2**20
 DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
+ENGINE_DITHER_SEED_BITWIDTH: Final = 2**64
```
bmerry, line 45: The value isn't a BITWIDTH - the bitwidth is 64. I'd suggest either
- Change the value to 64, and use `2**ENGINE_DITHER_SEED_BITWIDTH`; or
- Change this to `ENGINE_DITHER_SEED_MAX: Final = 2**64 - 1`, then use `randint` instead of `randrange`.

src/katgpucbf/xbgpu/engine.py
```diff
@@ -367,7 +370,7 @@ def __init__(
             # The magic constant was chosen at random. It ensures that the
             # seed won't be the same as in other types of engine that also use
             # sync_time as the basis for seeding.
```
bmerry, line 372:
```suggestion
```

src/katgpucbf/xbgpu/engine.py
```diff
@@ -420,10 +423,19 @@ def __init__(
...
-        self._populate_sensors()
+        self._populate_sensors(seed)
 
-    def _populate_sensors(self) -> None:
+    def _populate_sensors(self, seed: int) -> None:
         sensors = self.engine.sensors
+        sensors.add(
+            aiokatcp.Sensor(
+                str,
+                f"{self.name}.dither-seed",
```
bmerry, line 433: I don't think we have any other sensors with "bpipeline" in the name - we generally report things against individual beams, even when the value is the same for all the beams e.g. chan-range.

bmerry (changes requested):

src/katgpucbf/__init__.py
amishatishpatel: Argh, you are too right! My apologies. I must've been on autopilot with the naming. I might go with the former to follow even just the line above.

src/katgpucbf/xbgpu/engine.py
amishatishpatel: Yeah I did feel a bit uncomfortable adding it this way, but good point Re: chan-range sensor. Updating accordingly.

src/katgpucbf/__init__.py
```diff
@@ -42,7 +42,7 @@
 DEFAULT_JONES_PER_BATCH: Final = 2**20
 DIG_HEAP_SAMPLES: Final = 4096
 DIG_SAMPLE_BITS: Final = 10
-ENGINE_DITHER_SEED_BITWIDTH: Final = 2**64
+ENGINE_DITHER_SEED_BITWIDTH: Final = 64
```
bmerry, line 45: Since you mention the line above, what do you think of renaming this to ENGINE_DITHER_SEED_BITS for consistency? It seems BITWIDTH is only used in one other place (tests/xbgpu/test_engine.py), so maybe just change that one to BITS too.

bmerry (commented):
Code looks good. Just a bit of bikeshedding around the naming.

bmerry (approved):

bmerry (approved):

amishatishpatel:
Argh, I need to merge `main` into here and resolve any conflicts before I can mark this ready for review (again).

Apologies for any delay.

amishatishpatel:
I'm a bit suspicious of `main` merging so cleanly. I'll mark as Ready for Review now in case you want to have one more look on Monday @bmerry. I can always merge on Monday afternoon/evening along with the katsdpcontroller PR.

PR #961, opened 2025-08-11
Qualification: reuse receivers

bmerry:
This reduces the startup time for each test, because it doesn't need to recreate all the receiver infrastructure each time.

Tests that used the manual_start versions of the receivers now just use the regular version; the manual_start variant was essentially just an optimisation that is no longer needed.

Relates to NGC-1704.

Partly addresses NGC-1704.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -562,12 +562,91 @@ async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict
     return out
 
 
+class CoreAllocator:
+    """Provide CPU cores to receivers that need them.
+
+    It is initialised with a list of cores (from pytest config), with earlier
+    entries considered better than later ones. Cores are allocated in this
+    order. There is no mechanism to return cores; simply create a new
+    allocator to start fresh.
+    """
+
+    def __init__(self, cores: Iterable[int]) -> None:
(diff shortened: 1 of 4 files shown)


qualification/conftest.py
```diff
@@ -716,68 +750,21 @@ async def receive_baseline_correlation_products(
...
 async def receive_tied_array_channelised_voltage(
-    receive_tied_array_channelised_voltage_manual_start: TiedArrayChannelisedVoltageReceiver,
-    cbf_config: dict,
+    cbf: CBFRemoteControl,
     capture_start_streams: list[str],
 ) -> TiedArrayChannelisedVoltageReceiver:
-    """Create a spead2 receive stream for ingest the tied-array-channelised-voltage streams."""
-    receiver = receive_tied_array_channelised_voltage_manual_start
-    receiver.start()
+    """Get the spead2 receive stream for ingest the tied-array-channelised-voltage streams."""
```
amishatishpatel, line 758: I know it was just a slight tweak to this docstring, but it seems like the middle bit is a typo?
> for ingest the

Is it meant to read something like
- "for ingest *of the [...] streams", or
- "for *ingesting the [...] streams"?

I'll wait until you're back to raise a small PR accordingly.

PR #967, opened 2025-08-27
Fix warning from Jenkins

bmerry:
It complains that the branch and tag variables are not scoped appropriately:
```
14:36:55  Did you forget the `def` keyword? WorkflowScript seems to be setting a field named branch (to a value of type String) which could lead to memory leaks or other issues.
14:36:55  Did you forget the `def` keyword? WorkflowScript seems to be setting a field named tag (to a value of type String) which could lead to memory leaks or other issues.
```

- [x (n/a)] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -168,8 +168,8 @@ pipeline {
       }
       steps {
         script {
-          branch = env.BRANCH_NAME
-          tag = (branch == "main") ? "latest" : branch
+          String branch = env.BRANCH_NAME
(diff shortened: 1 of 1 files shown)


amishatishpatel (commented):
Trying to find info on variable declaration syntax in Jenkinsfiles and examples I've found so far do `def varname = "value"` - like the error suggests. 

Have you tested this syntax and/or do you know it to work?

PR #972, opened 2025-09-29
Update dependencies to latest versions

bmerry:
With the exception of the CUDA base Docker image: I upgraded to 12.9.1 rather than 13.x because that would require a driver upgrade on the hosts.

This required a few code changes because numpy now has more precise typing, which was causing mypy to know more about what type something is and complain, even in cases where we don't actually care about the different between Python floats and numpy floats.

- [x] If qualification tests are changed: attach a sample qualification report: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/248/
- [x] Test that benchmark_fgpu/benchmark_xbgpu still work

Relates to NGC-1717

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       - id: check-merge-conflict
       - id: check-shebang-scripts-are-executable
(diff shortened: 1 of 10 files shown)


bmerry:
I still plan to test them to make sure I don't break anything, but I reckon they don't need to block reviewing since I think they're low-risk and will take a while.

bmerry:
Lab qualification passed. I'll  just need to test benchmark_fgpu.

bmerry:
benchmark_fgpu/benchmark_xbgpu seem to be working.

amishatishpatel (approved):

PR #982, opened 2025-10-20
vgpu: add recv module and wire it up

bmerry:

- [x] (n/a) If design has changed: ensure documentation is up to date (**I'll make a separate PR to update the design docs for VLBI later**)
- [x] (n/a) If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match (**isn't yet a fake_server implementation of vgpu**)

Closes NGC-1693.

### DIFF
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -150,7 +150,7 @@ class VEngine(Engine):
                 ),
                 sink=self._recv_group,
             )
-            chunk.recycle()  # Adds to free_ringbuffer
+            chunk.recycle()  # Make available to the stream
 
     async def _run_receive(self) -> None:
         """Receive data from the tied-array-channelised-voltage streams."""
--- a/src/katgpucbf/vgpu/main.py
+++ b/src/katgpucbf/vgpu/main.py
@@ -61,7 +61,7 @@ def parse_args(arglist: Sequence[str] | None = None) -> argparse.Namespace:
         "--recv-jones-per-batch",
         type=int,
         default=DEFAULT_JONES_PER_BATCH,
-        help="Number of tied-antenna-channelised-voltage Jones vectors in each batch. [%(default)s]",
+        help="Number of tied-array-channelised-voltage Jones vectors in each batch. [%(default)s]",
     )
     parser.add_argument(
         "--recv-samples-between-spectra",


src/katgpucbf/vgpu/engine.py
```diff
@@ -62,6 +122,68 @@ def _populate_sensors(self, sensors: aiokatcp.SensorSet, send_pols: tuple[str, s
...
+        self._recv_group = recv.make_stream_group(
+            layout, data_ringbuffer, free_ringbuffer, recv_affinity, self.recv_pol_labels
+        )
+        for _ in range(recv_chunks):
+            chunk = recv.Chunk(
+                present=np.empty(
+                    (N_POLS, layout.n_batches_per_chunk, layout.n_pol_substreams),
+                    np.uint8,
+                ),
+                data=cupyx.empty_pinned(
```
amishatishpatel: I know this is a bit late to be asking (considering the PR to introduce cupy-cuda12x was before this), but
- What's the reasoning/requirement for pinned memory here?

Like the [documentation says](https://docs.cupy.dev/en/stable/reference/generated/cupyx.empty_pinned.html#cupyx-empty-pinned)
> This is a convenience function which is just [numpy.empty()](https://numpy.org/doc/stable/reference/generated/numpy.empty.html#numpy.empty), except that the underlying memory is pinned/pagelocked.

src/katgpucbf/vgpu/recv.py
```diff
@@ -0,0 +1,286 @@
...
+
+    user_data = np.zeros(N_POLS, dtype=user_data_type.dtype)
+    user_data["pol"] = np.arange(N_POLS)
+    group = base_recv.make_stream_group(
+        layout=layout,
+        spead_items=[TIMESTAMP_ID, FREQUENCY_ID, BEAM_ANTS_ID, spead2.HEAP_LENGTH_ID],
+        max_active_chunks=MAX_CHUNKS,
+        data_ringbuffer=data_ringbuffer,
+        free_ringbuffer=free_ringbuffer,
+        affinity=[recv_affinity] * N_POLS,
```
amishatishpatel: I know the data rates are low in this Rx, but is this... "fine"? 
- Or rather, how did you arrive at this conclusion? i.e. same core for both recv streams

amishatishpatel:
Sorry for the further delay. I thought I was done with the review and I somehow missed the `missing` and `reorder` parameterisation in the test_recv addition. I need to go through your logic there a bit closer before I can ask any useful questions.

test/vgpu/test_recv.py
```diff
@@ -0,0 +1,363 @@
...
+                    expected_present[:, :n_chunk_batches] = 1
+                    expected_data = np.zeros_like(chunk.data)
+                    expected_data[:, :n_chunk_batches] = data[:, start_batch : start_batch + n_chunk_batches]
+                    if missing:
+                        match i:
+                            case 1:
+                                expected_present[:, 1:] = 0
+                            case 2:
+                                expected_present[:, :1] = 0
+                                expected_present[0, 1, 0] = 0
```
amishatishpatel: I understand there is a lot of setup for data Tx and Rx here; did you have any other reasoning for consolidating the `missing` branch of this test into here?
- As far as I can tell, the other `test_recv`s have a separate `test_missing_heaps` which definitely improves readability (in my opinion)

I know we ultimately dictate how the test is done and verified, but I think this `if missing` would be more digestible if it was confined to its own test.

bmerry: I didn't want to duplicate 90% of the test and then have to maintain both copies. I think the other test_recv modules evolved a bit differently (particularly for fgpu, which changed substantially once stream-groups were introduced in spead2) which got them to where they are.

src/katgpucbf/vgpu/engine.py
amishatishpatel: Noted and thank you for that reminder - I was wondering why the layout looks similar, but the declaration is different. Follow-up question,
- Why use cupyx over katsdpsigproc.accel?

bmerry: Because that's what katcbf-vlbi-resample uses. That was a matter of expediency: it's much quicker to write something that works, and in particular it has some built-in signal processing like upfirdn that I would have had to hand-code if I was using katsdpsigproc.

test/vgpu/test_recv.py
```diff
@@ -0,0 +1,363 @@
...
+        reorder
+            If true, shuffle the heaps randomly before sending them.
+        """
+        n_batches = 25
+        rng = np.random.default_rng(seed=1)
+        data = rng.integers(
+            -127, 127, size=(N_POLS, n_batches, layout.n_channels, layout.n_spectra_per_heap, COMPLEX), dtype=np.int8
+        )
+        first_chunk_id = 123
+        first_timestamp = first_chunk_id * layout.chunk_timestamp_step
```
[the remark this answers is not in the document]
bmerry: The timestamp is what we (meaning the user of the recv module) actually care about. The chunk ID is just an implementation detail of the recv implementation.

PR #985, opened 2025-10-27
Fix link to documentation for pyproject.toml

bmerry:
pip seems to have reorganised the documentation somewhat, which broke doc builds.


bmerry:
> *Out of interest, did you spot this e.g. via Jenkins doc builds?

Yes.

PR #990, opened 2025-11-10
Misc documentation updates

bmerry:
This is a start on updating the documentation to account for V-engines, but I also noted some other outdated content along the way. This is still a work in progress (lots more documentation is needed around the V-engines); but I'm switching to other tasks at the moment so I wanted to get this merged before it drifts out of sync.

Relates to NGC-1811.

### DIFF
--- a/doc/engines.rst
+++ b/doc/engines.rst
@@ -132,8 +132,8 @@ The general operation of the DSP engines is illustrated in the diagram below:
    \end{scope}
 
 The F-engine and V-engine each use two input streams to align two incoming
-polarisations, but in the XB-engine there is only one as the polarisations are
-already interleaved in the incoming packets.
+polarisations. The XB-engine only uses one input stream as the polarisations
+are already interleaved in the incoming packets.
 
 There might not always be multiple processing pipelines. When they exist, they
 are to support multiple outputs generated from the same input, such as wide-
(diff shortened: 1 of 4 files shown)


doc/engines.rst
```diff
@@ -131,8 +131,9 @@ The general operation of the DSP engines is illustrated in the diagram below:
    \draw[queue] (group) -- (upload);
    \end{scope}
 
-The F-engine uses two input streams and aligns two incoming polarisations, but
-in the XB-engine there is only one.
+The F-engine and V-engine each use two input streams to align two incoming
+polarisations, but in the XB-engine there is only one as the polarisations are
+already interleaved in the incoming packets.
```
amishatishpatel: Very minor observation. I reckon these can be two separate sentences. The second sentence can read something like
> The XB-engine only uses one input stream as the polarisations are already interleaved in the incoming packets.

Just breaks up any ambiguity/confusion (like I had when first reading the sentence).

doc/introduction.rst
```diff
@@ -49,19 +49,20 @@ nodes. This eliminates the need to design costly custom backplanes.
 .. _CASPER: https://casper.berkeley.edu/
 
 
+This package
+------------
 
-This module
------------
```
amishatishpatel: There are still a few lingering references/usages of "module" rather than "package" in this file (Ctrl+F easily enough). Did you want to make a wholesale update to the new convention?

Worth mentioning `dev-guide.rst` still uses the "module" convention.

bmerry:
Thanks, all good comments. Should all be addressed now.

amishatishpatel (approved):
Much appreciated.

PR #1026, opened 2026-03-16
Add argparse option for multicast groups

EatonEmmerich:

- [x] If new source files are added: add copyright notices dated with the current year
- [x] `(N/A)` If qualification tests are changed: attach a sample qualification report
- [x] `(N/A)` If design has changed: ensure documentation is up to date
- [x] `(N/A)` If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match

Part one of NGC1910


bmerry (changes requested):

doc/benchmarking.rst
```diff
@@ -122,15 +122,13 @@ may need:
 
    Override the server names to find in the servers file.
 
+.. option:: --multicast-group <group>
+
+   Override the multicast group to use. The default is 239.192.128.0/20.
```
bmerry: "Group" in "multicast group" refers to the group of machines that do (possibly) many-to-many communication using a single IP address. 239.192.128.1 is a single multicast group. 239.192.128.0/20 is a set of 4096 multicast groups. So the command-line option, documentation, and variable names should be plural when referring to multiple IP addresses.

scratch/benchmarks/benchmark_tools.py
```diff
@@ -50,6 +51,30 @@
...
+        raise ValueError("offset must be non-negative")
+    if index >= network.num_addresses:
+        raise ValueError(f"offset {index} is greater than the number of addresses in the network {network}")
+    return ipaddress.ip_address(int(network.network_address) + index)
+
+
+def _split_network(
+    multicast_group: ipaddress.IPv4Network | ipaddress.IPv6Network,
+) -> tuple[ipaddress.IPv4Network | ipaddress.IPv6Network, ipaddress.IPv4Network | ipaddress.IPv6Network]:
+    """Split a CIDR in half and return the two subnets."""
```
bmerry: Take a look at `IPv4Network.subnets()` - it basically does this anyway (except that if the network is already a single host, it just returns that one rather than erroring).

scratch/benchmarks/benchmark_xbgpu.py
```diff
@@ -35,6 +36,8 @@
 from benchmark_tools import (
     PROMETHEUS_PORT_BASE,
     Benchmark,
+    _address_at_index,
+    _split_network,
```
bmerry: Since these are being imported, they're public functions of the `benchmark_tools` module. So remove the leading underscore.

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -53,6 +56,7 @@ def dsim_factory(
     single_pol: bool,
     sync_time: int,
     args: argparse.Namespace,
+    multicast_group: ipaddress.IPv4Network | ipaddress.IPv6Network,
```
bmerry: The type `ipaddress.IPv4Network | ipaddress.IPv6Network` appears a lot. You could introduce a type alias for it to reduce repetition.

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -131,20 +140,25 @@ def fgpu_factory(
     else:
         interface = server.interfaces[index % len(server.interfaces)]
         recv_cores = 2
+
+    narrowband_addresses_per_fgpu = args.xb // args.narrowband_decimation
```
bmerry: That's the number of addresses per fgpu *per output*. To get the number of addresses per fgpu you'll need to multiply by `args.narrowband`.

Check through the places where this is used though - I think if you just fix this formula here then you'll break some of the other calculations.

scratch/benchmarks/benchmark_xbgpu.py
```diff
@@ -102,6 +106,8 @@ def xbgpu_factory(
     adc_sample_rate: float,
     sync_time: int,
     args: argparse.Namespace,
+    multicast_group: ipaddress.IPv4Network | ipaddress.IPv6Network,
+    fsim_multicast_group: ipaddress.IPv4Network | ipaddress.IPv6Network,
```
bmerry: For fgpu_factory you had these arguments in the opposite order (first the producer's multicast groups, then the consumers).

bmerry (changes requested):
I'm still a little concerned that using split_network could cause the multicast groups selected to not distribute evenly when taken modulo 8, which would then cause load imbalances on our switch. When you've addressed the minor comments, give me a shout for some pair programming and I'll try to show you what I have in mind that I think would be both simpler and make more compact use of the address space.

scratch/benchmarks/benchmark_fgpu.py
EatonEmmerich: renamed to narrowband_addresses_per_fgpu_dst

scratch/fgpu/run-fgpu.sh
```diff
@@ -48,7 +48,7 @@ esac
 shift
 
 set -x
-exec spead2_net_raw taskset -c $other_affinity fgpu \
+exec fgpu \
```
bmerry: Is this debugging left over?

scratch/xbgpu/run-xbgpu.sh
```diff
@@ -66,7 +66,7 @@ fi
 
 set -x
 
-exec schedrr spead2_net_raw numactl -C $other_affinity xbgpu \
+exec schedrr numactl -C $other_affinity xbgpu \
```
bmerry: Is this debugging left over?

scratch/benchmarks/benchmark_fgpu.py
```diff
@@ -224,6 +62,168 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+            # (because fgpu_factory expects them to arrive on the same interface)
+            interface = server.interfaces[index // 2 % len(server.interfaces)]
+        katcp_port = KATCP_PORT_BASE + index
+        prometheus_port = PROMETHEUS_PORT_BASE + index
+        name = f"feng-dsim-{index}"
+        if self.single_pol:
+            addresses = [self.multicast_allocator.as_list(8)]
+        else:
+            addresses = [self.multicast_allocator.as_list(8), self.multicast_allocator.as_list(8)]
+        self.dsim_addresses = chain(self.dsim_addresses, iter(addresses))
```
bmerry: I guess this works, but it feels a bit messy and hard to debug because `dsim_addresses` will be a nested set of chains of iterators.

In retrospect, a better solution than passing iterators around may be to make `dsim_addresses` a `collections.deque`: then here we just use `self.dsim_addresses.extend(addresses)` and in the fgpu_factory we can use `popleft` to extract addresses.

scratch/benchmarks/benchmark_xbgpu.py
```diff
@@ -189,14 +73,130 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+            corrprod_number = index * self.args.corrprods + i
+            command += f"--corrprod=name=corrprod{corrprod_number},"
+            command += f"dst={self.multicast_allocator()},"
+            command += f"heap_accumulation_threshold={threshold} "
+        return command
+
+    @override
+    def reset(self) -> None:
+        super().reset()
+        self.fsim_addresses = iter([])
```
bmerry, line 192: What do you think about putting this at the top of `run_producers` instead, since it's the producers that then populate this? I'm happy either way.

scratch/benchmarks/benchmark_tools.py
```diff
@@ -371,7 +384,9 @@ async def trial(self, adc_sample_rate: float) -> TrialResult:
         sync_time = int(time.time())
         async with await self.run_producers(adc_sample_rate, sync_time):
             async with await self.run_consumers(adc_sample_rate, sync_time):
-                return await self.process(adc_sample_rate)
+                result = await self.process(adc_sample_rate)
+                self.reset()
```
bmerry: I *think* this is safe, in that if the code above throws an exception it'll bring down the whole process rather than being handled and followed by another trial. But I'd feel more sure it was safe if we did the reset at the beginning instead of the end, which would guarantee that the producers and consumers are started in a clean state rather than relying on the previous trial to have properly cleaned up.

bmerry (changes requested):

scratch/benchmarks/benchmark_xbgpu.py
EatonEmmerich: Would need to put both fsim_addresses and multicast address allocator resets in there to make it make sense.
There is opportunity to refactor the interface for Benchmark class more to improve this instead but the change will be large, we can do it in a follow up instead.

bmerry: Fair enough.

bmerry (approved):

PR #1029, opened 2026-03-25
vgpu: send VDIF packets

bmerry:

- [x] If new source files are added: add copyright notices dated with the current year

Closes NGC-1694.


test/vgpu/test_send.py
```diff
@@ -85,3 +95,164 @@ async def test(self) -> None:
...
+            return self._make_sender([(f"239.102.0.{i}", VTP_DEFAULT_PORT) for i in range(4)])
+
+    async def test_settings(self, sender: VDIFSender) -> None:
+        """Test that the sockets are correctly initialised."""
+        assert len(sender._socks) == 4
+        multicast_if = [b"\0\0\0\0\x7f\x00\x00\x01", b"\0\0\0\0\x7f\x00\x00\x02"]
+        for i, sock in enumerate(sender._socks):
+            sock.setsockopt.assert_any_call(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL, self.TTL)  # type: ignore
+            sock.setsockopt.assert_any_call(socket.IPPROTO_IP, socket.IP_MULTICAST_IF, multicast_if[i % 2])  # type: ignore
+            sock.setsockopt.assert_any_call(socket.SOL_SOCKET, socket.SO_SNDBUF, self.BUFFER)  # type: ignore
```
amishatishpatel, line 140: Any particular reason you are not checking/ensuring the specific order of these calls? Although, even typing out a suggestion shows me how much extra code there would need to be to ensure/confirm the order of calls.

Maybe it isn't imperative we ensure the call order?

test/vgpu/test_send.py
```diff
@@ -85,3 +95,164 @@ async def test(self) -> None:
...
+            sock.getsockopt.return_value = self.BUFFER
+            return sock
+
+        with mock.patch("socket.socket", autospec=True, side_effect=make_socket):
+            return self._make_sender([(f"239.102.0.{i}", VTP_DEFAULT_PORT) for i in range(4)])
+
+    async def test_settings(self, sender: VDIFSender) -> None:
+        """Test that the sockets are correctly initialised."""
+        assert len(sender._socks) == 4
+        multicast_if = [b"\0\0\0\0\x7f\x00\x00\x01", b"\0\0\0\0\x7f\x00\x00\x02"]
```
amishatishpatel: TIL backslash-zero is shorthand for `b"\x00"`

amishatishpatel: Although, by that logic, any particular reason for not doing the second pair of `\x00\x00` as `\0\0`? I take it for readability.

test/vgpu/test_send.py
```diff
@@ -85,3 +95,164 @@ async def test(self) -> None:
...
+        for i, sock in enumerate(sender._socks):
+            sock.setsockopt.assert_any_call(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL, self.TTL)  # type: ignore
+            sock.setsockopt.assert_any_call(socket.IPPROTO_IP, socket.IP_MULTICAST_IF, multicast_if[i % 2])  # type: ignore
+            sock.setsockopt.assert_any_call(socket.SOL_SOCKET, socket.SO_SNDBUF, self.BUFFER)  # type: ignore
+            sock.connect.assert_called_with((f"239.102.0.{i}", VTP_DEFAULT_PORT))  # type: ignore
```
amishatishpatel, line 141: I've just realised: A lot more `type: ignore`s in here than usual. Is it a necessary evil at this point? Slash, not worth of resolving?

test/vgpu/test_send.py
```diff
@@ -85,3 +95,164 @@ async def test(self) -> None:
...
+    async def test_sndbuf_failed(self, caplog: pytest.LogCaptureFixture) -> None:
+        """Test the logging if setting the socket buffer size fails."""
+
+        def setsockopt_side_effect(level: int, optname: int, value: int | Buffer, /) -> None:
+            if level == socket.SOL_SOCKET and optname == socket.SO_SNDBUF:
+                raise OSError(errno.EPERM, "Operation not permitted")
+
+        def make_socket(*args, **kwargs):
+            sock = mock.MagicMock()
+            sock.setsockopt.side_effect = setsockopt_side_effect
```
amishatishpatel, line 157: So, a `.side_effect` can be something that raises an Exception

OR

An Exception itself? e.g. Lines 212 and 253.

I mean, it's clearly working, but I did find that interesting.

doc/vgpu.design.rst
```diff
@@ -31,3 +31,38 @@ upstream before they are requested.
...
+   frameset (group of frames with the same timestamp but different thread IDs)
+   as a burst that is transmitted without intervening sleeps. This will
+   typically create smaller such bursts than the default in spead2, but
+   combined with the point above the actual number of bytes between bursts can
+   be quite large.
+3. The burst (catch-up) rate is set significantly higher than the default in
+   spead2, to compensate for potentially long pauses due to both Python's
+   stop-the-world garbage collector and to asyncio multiplexing work onto a
+   single kernel thread rather than having a dedicated thread for
+   transmission.
```
amishatishpatel: Could we split this up into (at least) two sentences, please. That is, if I'm understanding correctly, the potentially long pauses are confirmed to be due to
-  _either_ the garbage collector _or_ asyncio
- But more than likely **both**

```suggestion
3. The burst (catch-up) rate is set significantly higher than the default in
   spead2 to compensate for potentially long pauses. This can be due to Python's
   stop-the-world garbage collector, and asyncio multiplexing work onto a single
   kernel thread (rather than having a dedicated thread for transmission).
```

src/katgpucbf/vgpu/engine.py
```diff
@@ -309,16 +305,16 @@ async def _capture(self) -> None:
...
         it_rms = RecordPower(it_rms, sensors=self._sensors)
-        it = katcbf_vlbi_resample.power.NormalisePower(
-            it_rms, baseband.base.encoding.TWO_BIT_1_SIGMA / config.threshold
+        it = katcbf_vlbi_resample.power.NormalisePower(it_rms, 1.0)
+        it = katcbf_vlbi_resample.vdif_writer.VDIFEncode2Bit(
+            it, samples_per_frame=send_config.n_samples_per_frame, threshold=config.threshold
         )
-        it = katcbf_vlbi_resample.vdif_writer.VDIFEncode2Bit(it, samples_per_frame=send_config.n_samples_per_frame)
-        it = katcbf_vlbi_resample.cupy_bridge.AsNumpy(it)
+        it = katcbf_vlbi_resample.cupy_bridge.AsNumpy(it, 2)
```
amishatishpatel: Could I be a pain and ask you to make the naked `2` a named arg, please? Just Re: Principle of Least Astonishment.
- Or even add a comment around it?
- e.g. Even if I go to the commit, the message is "Try to get vgpu closer to being able to send reliably", which doesn't entirely explain what's going on here.

src/katgpucbf/vgpu/engine.py
```diff
@@ -353,6 +349,21 @@ def __init__(
...
+        # buffer).
+        queue_size = round(2 * config.power_int_time * send_config.bandwidth / send_config.n_samples_per_frame)
+        self._sender = send.VDIFSender(
+            send_config.dsts,
+            send_rate,
+            send_rate * 2.0,  # Python can introduce large pauses, so catch up aggressively
+            queue_size,
+            interfaces=send_config.interfaces,
+            ttl=send_config.ttl,
+            buffer=send_config.buffer,
```
amishatishpatel: Just me being weird, why did you order these args this way? In `VDIFSender` it's
```
*,
ttl: int,
buffer: int,
interfaces: list[str],
```

src/katgpucbf/vgpu/engine.py
```diff
@@ -191,7 +190,12 @@ class SendConfig:
     pols: tuple[str, str]
     bandwidth: float
     n_samples_per_frame: int
+    rate_factor: float
     station: str
+    dsts: list[tuple[str, int]]
+    interfaces: list[str]
+    buffer: int
```
amishatishpatel: I have to ask, any particular reason for naming this `buffer`  instead of e.g. `buffer_size` or `buffer_depth`?
- I read attribute `buffer` and keep expecting it to be of type `Buffer` e.g. `collections.abc.Buffer`
  - I know in a different file, but still figured I'd flag

src/katgpucbf/vgpu/engine.py
```diff
@@ -353,6 +349,21 @@ def __init__(
         )
         self._populate_sensors(self.sensors, recv_config.pol_labels, send_config.pols, recv_sensor_timeout)
         self._capture: _CaptureSession | None = None
+        send_rate = send_config.bandwidth * send_config.rate_factor
+        # Data comes out of the processing chain in chunks of size
+        # power_int_time. We need to smooth that out, so we use a send
+        # queue that is deeper than that (2 is the number of chunks to
+        # buffer).
+        queue_size = round(2 * config.power_int_time * send_config.bandwidth / send_config.n_samples_per_frame)
```
amishatishpatel, line 357: This `2 is the number of chunks to buffer` is the same `2` in Line 312? `it = katcbf_vlbi_resample.cupy_bridge.AsNumpy(it, 2)`
- I reckon ripe for a separate constant declared somewhere

src/katgpucbf/vgpu/send.py
```diff
@@ -0,0 +1,234 @@
...
+                now = loop.time()
+                target = max(self._next, self._next_burst)
+                # Don't try to sleep for short times. We tend to oversleep and
+                # then are unable to catch up.
+                if target - now > 1e-3:
+                    future = loop.create_future()
+                    loop.call_at(target, _set_result, future)
+                    await future
+                else:
+                    await asyncio.sleep(0)  # Give other asyncio tasks a chance to run
```
amishatishpatel: I think I'm following you Re: `Don't try to sleep for short times`, but it is a bit confusing to explicitly see an `asyncio.sleep` for times < 1ms...
- Do you reckon you could reword to speak about "delaying execution", or something like that?
- Maybe it's my own fault for overloading the `sleep` term

src/katgpucbf/vgpu/send.py
```diff
@@ -0,0 +1,234 @@
...
+    async def _process_item(self, item: T) -> None:
+        """Implement processing of an item.
+
+        This method does not handle the rate limiting.
+        """
+
+    async def _run(self) -> None:
+        """Process queue items.
+
+        This is scheduled as an asyncio task, only when the queue is non-empty.
```
amishatishpatel: I'll be honest, I've revisited this a couple times now and I still can't quite wrap my head around the reason for scheduling `only when the queue is non-empty`. Maybe it's easiest explained in real time.

amishatishpatel (commented):
Sorry I thought I hit Submit on this review last week!

src/katgpucbf/vgpu/engine.py
bmerry: I think I named it to match the command-line argument (`--send-buffer`), but I see the other class members are not consistently matching the command-line argument names, and I agree it's confusing. Will look at renaming it.

src/katgpucbf/vgpu/engine.py
bmerry: No, they're unrelated. They both give some extra wiggle room in buffering (to paper over any jitter), but they're controlling unrelated buffers.

src/katgpucbf/vgpu/engine.py
bmerry: Sure I can reorder them.

src/katgpucbf/vgpu/send.py
bmerry: It makes shutdown a little easier since you don't need to explicitly take care of stopping the task: it'll naturally shut itself down once the queue is drained. But I'll take another look; it might be that the extra complexity of checking whether the task is running mean it'll be simpler just to keep the task running full time.

test/vgpu/test_send.py
bmerry: Yes, it's a shortcut that unittest.mock provides: instead of writing a function just to throw an exception, you can just set an exception as the side effect.

src/katgpucbf/vgpu/engine.py
bmerry: Instead I've reverted the change. It was something I did in desperation before I figured out that VDIFSender needed a queue to buffer things, and now it seems fine without it.

src/katgpucbf/vgpu/send.py
bmerry: sleep(0) is a special case that doesn't ask the operating system to sleep. It just lets asyncio (in userspace) run other tasks that are waiting their turn.

test/vgpu/test_send.py
bmerry: I've changed them to be consistently `\x00` here.

test/vgpu/test_send.py
bmerry: No, we don't need to check the call order - there are independent and can happen in any order as long as they happen.

test/vgpu/test_send.py
bmerry: I think necessary evil because we're using mocks. Statically, `sock.connect` is normally a regular function (of the `socket.socket` class), but because it's been replaced by a mock at runtime it has methods that the type checker doesn't know about.

test/vgpu/test_send.py
```diff
@@ -130,7 +130,9 @@ def make_socket(*args, **kwargs):
     async def test_settings(self, sender: VDIFSender) -> None:
         """Test that the sockets are correctly initialised."""
         assert len(sender._socks) == 4
-        multicast_if = [b"\0\0\0\0\x7f\x00\x00\x01", b"\0\0\0\0\x7f\x00\x00\x02"]
+        # The first four bytes in each case are unused; the second four encode IP addresses
+        # 127.0.0.1 and 127.0.0.2 (in big endian).
```
amishatishpatel, line 134: Much appreciated.

src/katgpucbf/vgpu/send.py
```diff
@@ -89,37 +89,32 @@ async def _process_item(self, item: T) -> None:
...
-                    await self._process_item(item)
-                except Exception:
-                    logger.exception("Exception in processing rate-limited item")
-
-        finally:
-            self._run_task = None
+        while True:
+            item = await self._queue.get()
+            if item is None:
+                break
```
amishatishpatel, line 98: Not to be deliberately obtuse, but who is putting `None`s on this queue?
- I'm digging around katcbf-vlbi-resample and I do see some mentions/`__aiter__`s that indicate it's done

Have I got that right?

*That is, I don't see anything explicit or otherwise leading in vgpu's engine.py or send.py Re: Putting `None`s on this `_queue`. 
- I was also ~intrigued by `engine.py::_CaptureSession::_capture` where `VDIFFormatter` doesn't seem to "allow for" `None`s (returned to `frameset_it`)

amishatishpatel: Argh, sorry. I see you've updated `RateLimiter` to implement `stop()` method that does exactly this.

amishatishpatel (approved):

PR #1035, opened 2026-04-08
Dockerfile: don't use --link for copying schedrr

bmerry:
For reasons I haven't investigated (possibly it being run with a newer Docker version), it lead to an error message when trying to run setcap against it.


vincentseotlo (approved):
LGTM.

PR #1036, opened 2026-04-13
NGC-1914 v engine prevent downloading leap second table on capture start

vincentseotlo:
Astropy can be configured not to download the leap seconds table.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -187,5 +187,5 @@ RUN setcap cap_sys_nice+ep /usr/local/bin/schedrr
 COPY --link --from=build-py-requirements /venv /venv
 COPY --link docker/tuning.db /root/.cache/katsdpsigproc/tuning.db
 COPY --link --from=build-py /install-root/venv /venv
-COPY --link docker/astropy /etc/astropy
-ENV XDG_CONFIG_HOME=/etc
+RUN mkdir -p /root/.astropy/config
+COPY --link docker/astropy/astropy.cfg /root/.astropy/config/astropy.cfg
--- a/Dockerfile
+++ b/Dockerfile
@@ -187,3 +187,5 @@ RUN setcap cap_sys_nice+ep /usr/local/bin/schedrr
 COPY --link --from=build-py-requirements /venv /venv
 COPY --link docker/tuning.db /root/.cache/katsdpsigproc/tuning.db
 COPY --link --from=build-py /install-root/venv /venv
+COPY --link docker/astropy /etc/astropy
+ENV XDG_CONFIG_HOME=/etc
--- a/astropy/astropy.cfg
+++ b/astropy/astropy.cfg
@@ -2,5 +2,5 @@
 
 [utils.iers.iers]
 auto_download = False
-auto_max_age = 13
+auto_max_age = 15
 iers_auto_url =
--- a/doc/control.rst
+++ b/doc/control.rst
@@ -210,9 +210,9 @@ the Astropy config file.
 
 Set ``auto_download = False`` in the astropy.cfg resouce file as illustrated above.
 Use the environment variable `XDG_CONFIG_HOME` to point to the folder containing
-the astropy config directory. If the environment variable `XDG_CONFIG_HOME` is not
-set, Astropy will use the configuration found at `~/.astropy/config/` in the home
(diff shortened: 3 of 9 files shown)


doc/control.rst
```diff
@@ -196,6 +196,28 @@ data on a machine that doesn't support ibverbs, you could use
...
+If you need `ASTROPY`_ to run offline, set the IERS auto-download settings in
+the Astropy config file.
+
+.. code-block:: ini
+   :caption: astropy/astropy.cfg
+   :linenos:
+
+    [utils.iers.iers]
+    auto_download = False
+    auto_max_age = 15
```
bmerry: What is the purpose of this? If anything it seems to make it more likely to download data than the default (which is 30 days).

doc/control.rst
```diff
@@ -196,6 +196,28 @@ data on a machine that doesn't support ibverbs, you could use
...
+    iers_auto_url =
```
bmerry: What is the purpose of this? Is auto_download=False insufficient to prevent downloading?

doc/control.rst
```diff
@@ -196,6 +196,28 @@ data on a machine that doesn't support ibverbs, you could use
...
+
+Set ``auto_download = False`` in the astropy.cfg resouce file as illustrated above.
+Use the environment variable `XDG_CONFIG_HOME` to point to the folder containing
```
bmerry, line 212:
```suggestion
Use the environment variable :envvar:`XDG_CONFIG_HOME` to point to the folder containing
```

doc/control.rst
```diff
@@ -196,6 +196,28 @@ data on a machine that doesn't support ibverbs, you could use
...
+the astropy config directory. If the environment variable `XDG_CONFIG_HOME` is not
```
bmerry:
```suggestion
the astropy config directory. If the environment variable :envvar:`XDG_CONFIG_HOME` is not
```

astropy/astropy.cfg
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2026, National Research Foundation
+
+[utils.iers.iers]
+auto_download = False
+auto_max_age = 15
+iers_auto_url =
```
bmerry: See comments on the docs about the last two lines.

docker/astropy/astropy.cfg
bmerry, line 1: The Dockerfile should install this file into the appropriate place (it looks like the Docker image doesn't set XDG_CONFIG_DIR, so presumably ~/.astropy).

When doing that, also move this file to the `docker/` directory in the repo: that's where we keep stuff related to building the Docker image.

bmerry (changes requested):

astropy/astropy.cfg
vincentseotlo: True, we do need only auto_download. I'll exclude the other two in the next PR.

doc/control.rst
vincentseotlo: This isn't needed. I'll remove

docker/astropy/astropy.cfg
vincentseotlo: I was thinking we can place this in /etc?

bmerry: If that works then great. But from a quick look at the astropy code, I'm not sure it looks in /etc.

Dockerfile
```diff
@@ -187,3 +187,5 @@ RUN setcap cap_sys_nice+ep /usr/local/bin/schedrr
 COPY --link --from=build-py-requirements /venv /venv
 COPY --link docker/tuning.db /root/.cache/katsdpsigproc/tuning.db
 COPY --link --from=build-py /install-root/venv /venv
+COPY --link docker/astropy /etc/astropy
+ENV XDG_CONFIG_HOME=/etc
```
bmerry: I don't think messing with XDG_CONFIG_HOME like this is a good idea: it will impact many packages (it's a standardised environment variable) and possibly cause unexpected behaviour. Rather copy the config to /root/.astropy (similar to how docker/tuning.db is copied to /root/.cache).

bmerry (changes requested):

bmerry (approved):

PR #1045, opened 2026-05-05
fix no heaps case

EatonEmmerich:

- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [x] If the interface between the product controller and an engine has changed, update the

### DIFF
--- a/scratch/benchmarks/benchmark_tools.py
+++ b/scratch/benchmarks/benchmark_tools.py
@@ -203,7 +203,7 @@ class TrialResult:
 
     @property
     def state(self) -> ResultState:
-        if self.missing_heaps > 0:
+        if self.missing_heaps > 0 and self.heaps > 0:
             return ResultState.FAILED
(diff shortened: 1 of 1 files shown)


scratch/benchmarks/benchmark_tools.py
```diff
@@ -203,7 +203,7 @@ class TrialResult:
 
     @property
     def state(self) -> ResultState:
-        if self.missing_heaps > 0:
+        if self.missing_heaps > 0 and self.heaps > 0:
```
bmerry: I think it would be simpler to just hoist the check for `self.heaps == 0` to be first. Then you don't have to keep excluding it in other checks (e.g. in `0 < self.heaps < (1.0 - HEAPS_TOL) * self.expected_heaps`).

bmerry (changes requested):

bmerry (approved):

PR #1046, opened 2026-05-07
Make test_group_delay a bit more relaxed

bmerry:
The tolerance in the test was set to 5 sigma, which sounds like a lot but there are various biases that seem to act to pull things out of whack occasionally. Relax it to 7 sigma to hopefully reduce the noise from false negatives.

- [x] If qualification tests are changed: attach or link to a sample qualification report: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/478/ (some configs failed with lack of resources due to concurrent use of the lab)

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -801,7 +801,7 @@ async def test_group_delay(
     delay2, period2, std2 = await measure_once((-steps * dsim_resolution, steps * dsim_resolution))
 
     pdf_report.step("Analyse results.")
-    assert 6 * std1 < period2  # Ensure the first test localised things sufficiently
+    assert tolerance * std1 < period2  # Ensure the first test localised things sufficiently
     # Solve (approximately) delay1 = delay2 + k * period2
     k = round((delay1 - delay2) / period2)
     delay = delay2 + k * period2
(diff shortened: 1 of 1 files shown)


qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -797,7 +801,7 @@ async def measure_once(rel_freqs: tuple[float, float]) -> tuple[float, float, fl
     delay2, period2, std2 = await measure_once((-steps * dsim_resolution, steps * dsim_resolution))
 
     pdf_report.step("Analyse results.")
-    assert 5 * std1 < period2  # Ensure the first test localised things sufficiently
+    assert 6 * std1 < period2  # Ensure the first test localised things sufficiently
```
amishatishpatel: I don't admit to knowing these tests inside out, but why is this value 6 now? (And e.g. not the new `tolerance` variable)

amishatishpatel (commented):
Just clarifying something before I Approve. Thanks for the note on why some qualification test configs would fail.

qualification/antenna_channelised_voltage/test_delay.py
bmerry: Whoops, I started changing 5's to 6's and then decided to make a variable for it (and change it to 7) and missed this one. Will fix now.

amishatishpatel (approved):

PR #1050, opened 2026-05-15
Create qualification test that spins up v engine

EatonEmmerich:
Requires update to katsdpcontroller to prevent capture-start await on v engine kernel for the `develop: data_timeout=30` setting

- [x] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [x] If the interface between the product controller and an engine has changed, update the

Closes NGC-1922.

[ngc1922-report.pdf](https://github.com/user-attachments/files/28012049/ngc1922-report.pdf)

Depends on  https://github.com/ska-sa/katsdpcontroller/pull/836


qualification/pytest-jenkins-lab.ini
```diff
@@ -18,7 +18,7 @@ addopts = --report-log=report.json
 array_dir = arrays
 
 default_antennas = 8
-max_antennas = 16
+max_antennas = 8
```
bmerry: Is that because it was failing with 16? I think it *ought* to still work with 16, given that it's been passing even when somebody accidentally leaves one machine out of the pool.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,53 @@
...
+from ..cbf import CBFRemoteControl
+
+
+@pytest.mark.vlbi_only
+@pytest.mark.name("VLBI configuration")
+async def test_configuration(
+    cbf_mode_config: dict,
+    cbf: CBFRemoteControl,
+    start_tied_array_resampled_voltage_stream: bool,
+    receiver: TiedArrayChannelisedVoltageReceiver,
```
bmerry: I feel like pytest is going to complain that there is no fixture called `receiver`?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Verification method
+    -------------------
+    Verified by means of test. Create a product via the product controller client with no errors.
+    """
+    pdf_report.step("Verify configuration of tied-array-resampled-voltage stream.")
+    pcc = cbf.product_controller_client
+    dsim_timestamp = await pcc.sensor_value(f"{cbf.dsim_names[0]}.steady-state-timestamp", int)
+
+    _, data = await receiver.next_complete_chunk(min_timestamp=dsim_timestamp)
```
bmerry: Seems to be unused.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,44 @@
+################################################################################
+# Copyright (c) 2025-2026, National Research Foundation (SARAO)
```
bmerry: This file has only just been created, so it should be

```suggestion
# Copyright (c) 2026, National Research Foundation (SARAO)
```

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,44 @@
...
+
+from ..cbf import CBFRemoteControl
+
+
+@pytest.mark.vlbi_only
+@pytest.mark.name("VLBI configuration")
+async def test_configuration(
+    cbf: CBFRemoteControl,
+    pdf_report: Reporter,
+) -> None:
```
bmerry: I take it this test is intended as an example, which will probably be removed once we have some real specifications to test?

qualification/conftest.py
```diff
@@ -160,20 +160,30 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         values = [value for value in values if value <= max_antennas]
         metafunc.parametrize("n_antennas", values)
     if "band" in metafunc.fixturenames:
-        metafunc.parametrize("band", metafunc.config.getini("bands"))
+        if rel_path.parts[0] != "tied_array_resampled_voltage":
+            bands = metafunc.config.getini("bands")
+        else:
+            bands = ["l"]
+        metafunc.parametrize("band", bands)
```
bmerry: I would rather filter the specified bands, so that if the user explicitly configures the bands as `["u", "s"]` we don't run L-band tests anyway. Something like:

```suggestion
        bands = metafunc.config.getini("bands")
        if rel_path.parts[0] == "tied_array_resampled_voltage":
            bands = [band for band in bands if band == "l"]  # Limit to supported band
        metafunc.parametrize("band", bands)
```

qualification/conftest.py
```diff
@@ -160,20 +160,30 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
     if "n_channels" in metafunc.fixturenames or "narrowband_decimation" in metafunc.fixturenames:
         # NOTE: Each config tuple is in the format (n_channels, decimation, vlbi_mode).
```
bmerry: Need to update this comment.

bmerry (changes requested):
I left a few low-level comments, but realised that I think we should make a high-level change. When we initially put in testing for the VLBI flavour of the narrowband F-engine, we left it orthogonal to the band since that was just easier. But since it only exists for use with VLBI (including the V-engine), I don't think it makes sense to have parametrisations where narrowband_vlbi is true while start_vlbi is false.

So I would rather make the band non-orthogonal to the narrowband_vlbi parameter i.e. it will be part of the big "configs" tuple that gets built up. We can only rename `narrowband_vlbi` to just `vlbi`.

We can leave vlbi_decimation in the config files as "4 8", since we do eventually want to support VLBI with decimation 4, and it looks like Vincent's work on NGC-1928 might make that functional already.

With all that done, we won't need to handle the tied_array_resample_voltage directory any differently; the `vlbi_only` mark on the tests in that directory will be sufficient to ensure the tests are only run on appropriate configs.

qualification/conftest.py
```diff
@@ -159,32 +159,46 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
-            (int(n_channels), int(nb_decimation), False)
+            (int(n_channels), int(nb_decimation), band, False)
             for nb_decimation in metafunc.config.getini("narrowband_decimation")
             for n_channels in metafunc.config.getini("narrowband_channels")
+            for band in bands
         )
+        if rel_path.parts[0] != "tied_array_resampled_voltage":
+            vlbi_decimation = metafunc.config.getini("vlbi_decimation")
+        else:
+            vlbi_decimation = [8]
```
bmerry: Revert this, and just use metafunc.config.getini("vlbi_decimation") as before.

qualification/conftest.py
```diff
@@ -159,32 +159,46 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
-            configs = [config for config in configs if config[2]]
+            configs = [config for config in configs if config[3]]
         if not configs:
             raise RuntimeError(f"Contradictory markers on test {metafunc.function.originalname}")
-        metafunc.parametrize("n_channels, narrowband_decimation, narrowband_vlbi", configs)
+        metafunc.parametrize("n_channels, narrowband_decimation, band, vlbi", configs)
+    else:
+        if "band" in metafunc.fixturenames:
+            bands = metafunc.config.getini("bands")
+            metafunc.parametrize("band", bands)
```
bmerry: Use "elif" rather than "else: if:"

qualification/conftest.py
```diff
@@ -358,6 +372,17 @@ async def _cbf_config_and_description(
                 "src_pol": pol_idx,
             }
 
+    n_vengines = 0
```
bmerry: Let's rather call this vlbi_beams. The fact that we use one engine to implement one tied-array-resamples-voltage stream is an implementation detail. Similarly for where you've used "vengines" below.

bmerry (changes requested):
Much happier with the structure now; just a few localised things to fix up.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: no longer needed since we start the engine with the existing filter test too.

qualification/conftest.py
```diff
@@ -397,12 +393,14 @@ async def _cbf_config_and_description(
         "vlbi": str(vlbi),
         "dsims": str(n_dsims),
         "beams": str(n_beams),
-        "vengines": str(n_vengines),
+        "vengines": str(vlbi_beams),
```
bmerry: Also rename the key to "vlbi_beams".

qualification/conftest.py
bmerry: Can remove this now.

qualification/conftest.py
```diff
@@ -397,12 +393,14 @@ async def _cbf_config_and_description(
         "vlbi": str(vlbi),
```
bmerry: Can you check where this gets used? It's redundant now given that we have vlbi_beams, but I don't know off the top of my head what will break if we remove it.

qualification/conftest.py
```diff
@@ -369,17 +390,20 @@ async def _cbf_config_and_description(
...
         "dsims": str(n_dsims),
         "beams": str(n_beams),
+        "vengines": str(vlbi_beams),
     }
     long_description = (
         f"{n_antennas} antennas, {n_channels} channels, {n_beams} beams, "
         f"{BANDS[band].long_name}-band, {int_time}s integrations, {n_dsims} dsims"
     )
+    if vlbi:
+        long_description += f", {vlbi_beams} vlbi-beams"
```
bmerry:
```suggestion
        long_description += f", {vlbi_beams} VLBI beams"
```

bmerry (changes requested):

qualification/report/generate_pdf.py
```diff
@@ -96,7 +96,7 @@ def cbf_sort_key(config: dict) -> tuple:
         float(config["integration_time"]),
         int(config["dsims"]),
         int(config["narrowband_decimation"]),
-        str_to_bool(config.get("narrowband_vlbi", "False")),
+        str_to_bool(config.get("vlbi_beams", "0")),
```
bmerry: Should be `int` rather than `str_to_bool`, since vlbi_beams is a (stringified) integer, not a bool.

Also keep the `narrowband_vlbi` key after the vlbi_beams key. That way if we run the report generation on the output of an old test, it can still sort on that.

qualification/report/generate_pdf.py
```diff
@@ -302,8 +302,8 @@ def mode_str(self, *, expand: bool = False) -> str:
         Short or long description of CBF mode.
         """
         narrowband_decimation = int(self.mode_config["narrowband_decimation"])
-        # Older report files won't contain the narrowband_vlbi key
-        narrowband_vlbi = str_to_bool(self.mode_config.get("narrowband_vlbi", "False"))
+        # Older report files won't contain the vlbi_beams key
+        vlbi_beams = int(self.mode_config.get("vlbi_beams", "0"))
```
bmerry, line 310: It looks like we're trying to maintain backwards compatibility here. So if vlbi_beams is not present, we should fall back to checking if `narrowband_vlbi` is present and using that instead (cast to int).

bmerry (changes requested):

qualification/report/generate_pdf.py
EatonEmmerich: Do we have a timeframe in which we support these old reports?

bmerry: There's no explicit policy. We just try to do so when it's not going to be a massive pain.

qualification/report/generate_pdf.py
```diff
@@ -96,7 +96,8 @@ def cbf_sort_key(config: dict) -> tuple:
         float(config["integration_time"]),
         int(config["dsims"]),
         int(config["narrowband_decimation"]),
-        str_to_bool(config.get("vlbi_beams", "0")),
+        str_to_bool(config.get("narrowband_vlbi", "False")),
```
bmerry: I'd suggest a comment e.g. "# for compatibility with old test runs that pre-date vlbi_beams"

bmerry (approved):
LGTM - one minor suggestion you can take or leave

bmerry (approved):

PR #1052, opened 2026-05-26
Bump idna to newest version

bmerry:
Addresses CVE-2026-45409

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -125,7 +125,7 @@ hiredis==3.3.0
     #   katsdptelstate
 identify==2.6.16
     # via pre-commit
-idna==3.11
+idna==3.16
     # via
     #   -c requirements.txt
     #   requests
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -125,7 +125,7 @@ hiredis==3.3.0
     #   katsdptelstate
 identify==2.6.16
     # via pre-commit
-idna==3.11
```
EatonEmmerich, line 128: why are we duplicating the requirements in requirements-dev.txt?

EatonEmmerich (approved):
Small question, looks good otherwise

requirements-dev.txt
bmerry: TBH it's long enough ago that I don't recall why we set it up this way. It does mean you can set up for dev with just installing from one file, instead of needing to install both requirements.txt and requirements-dev.txt.

PR #1054, opened 2026-06-02
upgrade to cuda13.x

vincentseotlo:

- [X] If dependencies are added/removed: update `pyproject.toml` and `.pre-commit-config.yaml`.
- [X] If new source files are added: add copyright notices dated with the current year.
- [X] If qualification tests are changed: attach or link to a sample qualification report.
- [X] If design has changed: ensure documentation is up to date.
- [X] If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [X] If the interface between the product controller and an engine has changed, update the

Closes NGC-1982.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libcurand-dev-13-2 \
     libcublas-13-2 \
     libcufft-13-2 \
+    libcusparse-13-2 \
+    libnvjitlink-13-2 \
     libvulkan1 \
     libegl1 \
     libxext6
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1
 
 ################################################################################
-# Copyright (c) 2021-2025, National Research Foundation (SARAO)
+# Copyright (c) 2021-2026, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1
 
 ################################################################################
-# Copyright (c) 2021-2025, National Research Foundation (SARAO)
+# Copyright (c) 2021-2026, National Research Foundation (SARAO)
```
bmerry: Revert please.

Dockerfile
```diff
@@ -20,7 +20,7 @@
 # nvidia and cuda runtime and development tools. pycuda needs nvcc, so
 # the development tools are necessary.
 
-FROM nvidia/cuda:12.9.1-base-ubuntu24.04 AS base
+FROM nvidia/cuda:13.2.0-base-ubuntu24.04 AS base
```
bmerry:
```suggestion
FROM nvidia/cuda:13.2.1-base-ubuntu24.04 AS base
```

bmerry (changes requested):
You'll also need to replace cupy-cuda12x with cupy-cuda13x in requirements files.

Dockerfile
vincentseotlo: Done

Dockerfile
vincentseotlo: Done

Dockerfile
vincentseotlo: Done

pyproject.toml
```diff
@@ -30,6 +30,7 @@ dependencies = [
     "katsdptelstate",
     "numba",
     "numpy",
+    "nvidia-nvjitlink",
```
bmerry: If this is still needed at all (after reverting cupy to 13.6), please install it with apt so that we get all the CUDA bits from one place (instead of some from apt and some from pip). I think the relevant package would be `libnvjitlink-13-2` (if that doesn't work, try `libnvjitlink-dev-13-2`).

bmerry (changes requested):

bmerry (approved):
LGTM

PR #1055, opened 2026-06-02
Improve vgpu end-to-end unit testing

bmerry:
Instead of just mocking out the send function at the top level and asserting that it was called, mock out socket.sendmsg at the lower level and collect the actual packet data sent. This allows us to verify the sequence numbers and VDIF header fields. That also simplifies the tests somewhat because the test_min_timestamp test can be eliminated in favour of just having test_smoke handle a min_timestamp, while at the same time making the min_timestamp testing more strict.

Contributes to NGC-1695.

### DIFF
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -124,6 +124,7 @@ class RecvStream:
                     zero_arr.attrs["time_bias"] = timestamp // self._samples_between_spectra
                     yield zero_arr
                 last_chunk_id = chunk.chunk_id
+                print(chunk.timestamp)
                 yield arr
 
 
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -315,6 +315,7 @@ class _CaptureSession:
         )
         async for frameset in frameset_it:
             await self._sender.send(frameset)
+        await self._sender.flush()
         self._capture_complete()
 
     async def stop(self) -> None:
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -124,7 +124,6 @@ class RecvStream:
                     zero_arr.attrs["time_bias"] = timestamp // self._samples_between_spectra
                     yield zero_arr
                 last_chunk_id = chunk.chunk_id
-                print(chunk.timestamp)
                 yield arr
 
 
--- a/src/katgpucbf/vgpu/send.py
+++ b/src/katgpucbf/vgpu/send.py
@@ -72,7 +72,7 @@ class RateLimiter[T](ABC):
         self._per_unit = 1 / rate if rate else 0.0
         # Seconds per unit for `burst_rate`
         self._per_unit_burst = 1 / burst_rate if burst_rate else 0.0
-        self._queue: asyncio.Queue[T | None] = asyncio.Queue(capacity)
+        self._queue: asyncio.Queue[T | asyncio.Future[None] | None] = asyncio.Queue(capacity)
         self._run_task: asyncio.Task = asyncio.create_task(self._run(), name="RateLimiter")
 
     @abstractmethod
@@ -96,6 +96,10 @@ class RateLimiter[T](ABC):
             item = await self._queue.get()
             if item is None:
                 break
+            elif isinstance(item, asyncio.Future):
+                if not item.done():
+                    item.set_result(None)
+                continue
 
             now = loop.time()
             target = max(self._next, self._next_burst)
@@ -130,8 +134,18 @@ class RateLimiter[T](ABC):
             self._next = max(self._next, now)
             self._next_burst = max(self._next_burst, now)
 
-    async def stop(self) -> None:
+    async def flush(self) -> None:
         """Wait until all queued items have been processed."""
+        future = asyncio.Future[None]()
+        await self._queue.put(future)
+        await future
+
+    async def stop(self) -> None:
+        """Shut down the rate limiter.
+
+        This blocks until all queued items have been processed. No other
+        methods should be called after this one, as they may deadlock.
+        """
         if not self._run_task.done():
             await self._queue.put(None)  # Signals run task to stop
         await self._run_task
--- a/src/katgpucbf/vgpu/send.py
+++ b/src/katgpucbf/vgpu/send.py
@@ -72,6 +72,11 @@ class RateLimiter[T](ABC):
         self._per_unit = 1 / rate if rate else 0.0
         # Seconds per unit for `burst_rate`
         self._per_unit_burst = 1 / burst_rate if burst_rate else 0.0
+        # Each item in the queue is either:
+        # - an item to send;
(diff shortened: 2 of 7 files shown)


src/katgpucbf/vgpu/engine.py
```diff
@@ -124,6 +124,7 @@ async def __aiter__(self) -> AsyncIterator[xr.DataArray]:
                     zero_arr.attrs["time_bias"] = timestamp // self._samples_between_spectra
                     yield zero_arr
                 last_chunk_id = chunk.chunk_id
+                print(chunk.timestamp)
```
EatonEmmerich: forgotten debug print.

test/vgpu/test_engine.py
```diff
@@ -122,47 +128,69 @@ def engine_arglist(self) -> list[str]:
             "239.10.2.0",
         ]
 
+    @pytest.mark.parametrize(
+        "min_timestamp",
+        [0, FIRST_TIMESTAMP + 0x400000],
```
EatonEmmerich: nit:
I prefer reading the different rows of data parameterised on seperate lines.
ie:

```suggestion
        [
            0,
            FIRST_TIMESTAMP + 0x400000
        ],
```
makes it easier to distinguish an array passed in from the value in the array passed in.

test/vgpu/test_send.py
```diff
@@ -96,6 +96,27 @@ async def test(self) -> None:
...
+        for _ in range(5):
+            await limiter.send(1)
+        # Start a flush, and asynchronously also add more data
+        async with asyncio.TaskGroup() as tg:
+            flush_task = tg.create_task(limiter.flush())
+            await asyncio.sleep(0.1)  # Give flush_task time to start
+            for _ in range(5):
+                await limiter.send(1)
+            print(loop.time() - start_time)
+            await flush_task
```
EatonEmmerich: This would happen instantaneusly anyway, right?
Suggest we add some items with sleep to confirm await behaviour.

I would also suggest not using the magic value of `4` to do that with another method, maybe a new `set_delay(sleep=0.5)` on the `DummyRateLimiter`

test/vgpu/test_engine.py
```diff
@@ -122,47 +128,69 @@ def engine_arglist(self) -> list[str]:
...
         self,
         engine: VEngine,
         engine_client: aiokatcp.Client,
         mock_recv_streams: list[spead2.InprocQueue],
-        frameset_count: list[int],
+        mock_sendmsg: list[bytes],
         capture_complete_event: asyncio.Event,
+        min_timestamp: int,
     ) -> None:
         """Test that an engine can be started and receives some data.
```
EatonEmmerich:
```suggestion
        """Test that an engine can be started, receives and transmits some data.
```

test/vgpu/test_engine.py
```diff
@@ -122,47 +128,69 @@ def engine_arglist(self) -> list[str]:
...
 
-        This is a weak test that will need to be filled out later to ensure
-        that the framesets contain the right headers and data.
+        This is a weak test that considers only the headers and not the data.
```
EatonEmmerich:
```suggestion
        This is a weak test that considers only the headers and not the data of the VDIF output.
```

test/vgpu/conftest.py
```diff
@@ -41,3 +42,21 @@ async def engine(
     await server.start()
     yield server
     await server.stop()
+
+
+@pytest.fixture
+def mock_sendmsg(monkeypatch: pytest.MonkeyPatch) -> list[bytes]:
```
EatonEmmerich: nit: this is technically not a mock, we are monkeypatching sendmsg with a 'mock', but this function is returning the bytes we sent to that mock instead.
Suggest rename to something like:

```suggestion
def sendmsg_captured_bytes(monkeypatch: pytest.MonkeyPatch) -> list[bytes]:
```

src/katgpucbf/vgpu/send.py
```diff
@@ -96,6 +96,10 @@ async def _run(self) -> None:
             item = await self._queue.get()
             if item is None:
                 break
+            elif isinstance(item, asyncio.Future):
```
EatonEmmerich, line 104: why check the future result here if we are doing awaits on line 113 before processing item?
Can't we just move this elif condition to after those awaits?

bmerry: I'm not really following this comment, but I'm adding some extra documentation where _queue is created which will maybe make things clearer about what this branch is for.

test/vgpu/conftest.py
bmerry: Nice idea - it makes the code using this easier to read since it's clear what the variable corresponding to the fixture actually holds. I just called it `sendmsg_packets` since `bytes` is a Python type not a description, and I wanted to keep the name somewhat short.

test/vgpu/test_engine.py
bmerry: I like it. Done.

test/vgpu/test_send.py
bmerry: > This would happen instantaneusly anyway, right?

No, this blocks until the rate limiter has actually processed the 5 items - if you remove this line the test fails. Recall that `await limiter.send(1)` doesn't actually await transmission of the item, it just awaits having space in the queue, and we've given the queue a capacity of 100 so it doesn't block.

I'm not so keen on adding an out-of-band `set_delay` method on DummyRateLimiter, because it gets hard to reason about which queue items will actually be affected by it (since the rate limiter is drip-feeding items from the queue into `_process_item` asynchronously from any calls to a `set_delay`). What I'll do is change item from an int to a dataclass that explicitly specifies the sleep time, to avoid any magic numbers.

test/vgpu/test_engine.py
```diff
@@ -122,47 +128,69 @@ def engine_arglist(self) -> list[str]:
...
+            seq = struct.unpack("<Q", fh.read(8))[0]
+            frame = VDIFFrame.fromfile(fh)
+            header = frame.header
+            frame_time = frame.get_time(frame_rate=frame_rate)
+            expected_time = start_time + i // N_THREADS / frame_rate
+            assert seq == i
+            assert not header.complex_data
+            assert not header["invalid_data"]
+            assert header.bps == 2
+            assert header.nchan == 1
```
EatonEmmerich, line 182: I'm not sure I understand how this is possible (the spec specifies factor of 2 for this?)

EatonEmmerich: nvm I realize now that the value is `log2(nchan)` not nchan.

bmerry: It is the number of channels. In baseband, attributes of header are semantic values whereas using it as a dict gives raw fields (e.g. `header["thread_id"]` is used instead of header.thread_id: the latter doesn't exist because there is no need for it as the raw field doesn't require any interpretation).

The output stream has 4 VDIF threads, one for each polarisation of each of the 2 channels. But the VDIF header describes just the data in the frame itself, which is 1 channel.

EatonEmmerich (approved):

PR #1062, opened 2026-06-23
Implement ?vlbi-delay

bmerry:
The actual implementation is quite simple, but some major changes were needed to test it. I've edited the history to create a number of self-contained commits.

Closes NGC-1695.

### DIFF
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -90,45 +90,44 @@ class RecvStream:
 
     async def __aiter__(self) -> AsyncIterator[xr.DataArray]:
         last_chunk_id: int | None = None
-        with self._recv_iter:
-            async for chunk in self._recv_iter:
-                with chunk:
-                    if chunk.timestamp < self._min_timestamp:
-                        continue
-                    # TODO: need to do something with the presence flags
-                    # TODO: pipeline these transfers (but keeping in mind
-                    # that we need to recycle the chunk only when the transfer
-                    # is complete).
-                    data = cp.asarray(chunk.data, blocking=False)
-                    await katcbf_vlbi_resample.utils.stream_future(None)
-                    # There are two time axes. Transpose to place them together, then flatten
-                    # over them. The current shape is
-                    # (N_POLS, n_batches_per_chunk, channels, n_spectra_per_heap, COMPLEX)
-                    data = data.transpose(0, 1, 3, 2, 4)
-                    # Now it is
-                    # (N_POLS, n_batches_per_chunk, n_spectra_per_heap, channels, COMPLEX)
-                    data = data.reshape(N_POLS, -1, self.channels, COMPLEX)
-                    # Now it is
-                    # (N_POLS, n_spectra_per_chunk, channels, COMPLEX)
-                    # Convert Gaussian integers to complex
-                    data = cp.ascontiguousarray(data.astype(np.float32)).view(np.complex64)[..., 0]
-                    arr = xr.DataArray(
-                        data,
-                        dims=("pol", "time", "channel"),
-                        coords={"pol": list(self._pol_labels)},
-                        attrs={"time_bias": chunk.timestamp // self._samples_between_spectra},
-                    )
-                    # TODO (NGC-1689): need to properly handle missing data in
-                    # katcbf-vlbi-resample. This is a quick hack to keep things
-                    # running by injecting zero data into the stream.
-                    while last_chunk_id is not None and last_chunk_id < chunk.chunk_id - 1:
-                        last_chunk_id += 1
-                        zero_arr = xr.zeros_like(arr)
-                        timestamp = last_chunk_id * self._layout.chunk_timestamp_step
-                        zero_arr.attrs["time_bias"] = timestamp // self._samples_between_spectra
-                        yield zero_arr
-                    last_chunk_id = chunk.chunk_id
-                    yield arr
+        async for chunk in self._recv_iter:
+            with chunk:
+                if chunk.timestamp < self._min_timestamp:
+                    continue
+                # TODO: need to do something with the presence flags
+                # TODO: pipeline these transfers (but keeping in mind
+                # that we need to recycle the chunk only when the transfer
+                # is complete).
+                data = cp.asarray(chunk.data, blocking=False)
+                await katcbf_vlbi_resample.utils.stream_future(None)
+                # There are two time axes. Transpose to place them together, then flatten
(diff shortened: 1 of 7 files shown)


test/vgpu/test_engine.py
```diff
@@ -70,28 +71,35 @@ async def _send_data(
...
+    rng = np.random.default_rng(seed=seed)
+    data[:] = rng.integers(-127, 127, data.shape, data.dtype)
+
+
 class TestVEngine:
     """Test :class:`.VEngine`."""
 
     @pytest.fixture
-    def capture_complete_event(self, engine: VEngine, monkeypatch: pytest.MonkeyPatch) -> asyncio.Event:
+    async def capture_complete_event(self, engine: VEngine, monkeypatch: pytest.MonkeyPatch) -> asyncio.Event:
```
bmerry, line 102: I made this async because constructing an Event requires a current event loop. There happened to always be one when this got called, but making it async guarantees it.

test/vgpu/test_engine.py
```diff
@@ -70,28 +71,35 @@ async def _send_data(
...
         factory(data)
-    for heap in gen_heaps(layout, data, FIRST_TIMESTAMP):
+    for heap in gen_heaps(layout, data, first_timestamp):
         await beng.async_send_heap(heap.heap, substream_index=heap.substream_index)
 
 
+def _randomize(data: np.ndarray, seed: int) -> None:
+    """Randomize `data` in place, using a fixed seed."""
+    rng = np.random.default_rng(seed=seed)
+    data[:] = rng.integers(-127, 127, data.shape, data.dtype)
```
amishatishpatel, line 95: Wow, TIL about using random number generators with the same seed.

test/vgpu/test_engine.py
```diff
@@ -211,23 +219,74 @@ async def test_capture_stop_while_not_capturing(self, engine_client: aiokatcp.Cl
...
+            # Prepare for the next pass
+            sendmsg_packets.clear()
+            capture_complete_event.clear()
+            # The fixture cleanup will do this, but only at the end of the
+            # test. Release the resources now.
+            engine_client.close()
+            await engine_client.wait_closed()
+            await engine.stop()
+
+        assert len(packets[0]) > 0
```
amishatishpatel, line 278: Why only check if one list of packets has nonzero length? And e.g. not an `all(len(p) > 0 for p in packets)`

src/katgpucbf/vgpu/engine.py
```diff
@@ -438,9 +450,23 @@ async def on_stop(self) -> None:  # noqa: D102
...
 
     async def request_vlbi_delay(self, ctx: aiokatcp.RequestContext, delay: float) -> None:
-        """Set the delay applied to the stream, in second."""
-        # TODO: will need to be rounded/quantised
-        self.sensors["delay"].value = delay
+        """Set the delay applied to the stream, in seconds."""
+        if self._capture is not None:
+            raise aiokatcp.FailReply("cannot set vlbi-delay while capturing")
+        # Compute the fractional part of the precise sync epoch
+        sync_frac = (self.config.recv_config.time_converter.sync_time % 1.0 + delay % 1.0) % 1.0
```
amishatishpatel, line 461: Do you need the extra `% 1.0` outside the brackets?

amishatishpatel (commented):
Minor comments before explicitly Approving.

amishatishpatel:
Also, Re: marking the below (n/a)
> If design has changed: ensure documentation is up to date.

Is this already documented somewhere? I see mention in https://katgpucbf.readthedocs.io/en/latest/control.html#vgpu, but nowhere else.

src/katgpucbf/vgpu/engine.py
bmerry: If sync_time and delay are both 0.9, then without the last % 1.0, sync_frac would be 1.8 whereas with the % 1.0 it would be 0.8.

Mathematically, *none* of the % 1.0's are necessary, but because the values are floating point, throwing away the whole seconds leaves more mantissa bits for the fractions of a second. So I'm doing it at every step to minimise rounding error.

test/vgpu/test_engine.py
bmerry: It's really to check that `chunks` is big enough that some data comes out the far end of the pipeline rather than just partially filling buffers; and for that either list is as good as the other.

The `strict=True` on the line below will ensure that the two lists have the same length.

I'm happy to add an explicit check that the lengths are the same if you think that will be clearer.

amishatishpatel: > The `strict=True` on the line below will ensure that the two lists have the same length.

Well... [more specifically](https://peps.python.org/pep-0618/#specification)
> the [...] iterator will raise a `ValueError` if the arguments are exhausted at different lengths

I gather the amount of data sent is fixed (by `chunks`), so we don't "expect" data to go missing. I think some kind of note would be helpful here.

src/katgpucbf/vgpu/engine.py
amishatishpatel: Ah, gosh, too right. Thank you for clarifying.

amishatishpatel (commented):
Not hitting Approve just yet as I'd asked a question after the last round of reviewing
- https://github.com/ska-sa/katgpucbf/pull/1062#issuecomment-4797921199

> Also, Re: marking the below (n/a)
>> If design has changed: ensure documentation is up to date.
> Is this already documented somewhere? I see mention in https://katgpucbf.readthedocs.io/en/latest/control.html#vgpu, but nowhere else.

bmerry:
> Not hitting Approve just yet as I'd asked a question after the last round of reviewing
> 
>     * [#1062 (comment)](https://github.com/ska-sa/katgpucbf/pull/1062#issuecomment-4797921199)
> 
> 
> > Also, Re: marking the below (n/a)
> > > If design has changed: ensure documentation is up to date.
> > > Is this already documented somewhere? I see mention in https://katgpucbf.readthedocs.io/en/latest/control.html#vgpu, but nowhere else.

Thanks, I'd missed that. I've added a bit to vgpu.design.rst.

test/vgpu/test_engine.py
bmerry: I've added an assertion that they have the same length, to make it clearer that that's something we expect. Does that address your concern?

amishatishpatel (approved):

test/vgpu/test_engine.py
amishatishpatel: I think that is sufficient. Thanks.

PR #1066, opened 2026-06-30
Calculate slope values after updating --startup-time

EatonEmmerich:

- [x] `(N/A)` If dependencies are added/removed: update `pyproject.toml` and `.pre-commit-config.yaml`.
- [x] `(N/A)` If qualification tests are changed: attach or link to a sample qualification report.
- [x] `(N/A)` If design has changed: ensure documentation is up to date.
- [x] `(N/A)` If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [x] `(N/A)` If the interface between the product controller and an engine has changed, update the

Closes NGC-1945.


scratch/benchmarks/runcalibratexbgpu.sh
```diff
@@ -0,0 +1,71 @@
...
+n=$2
+low_freq=$3
+high_freq=$4
+
+max_ssh_retries=${MAX_SSH_RETRIES:-3}
+ssh_retry_delay=${SSH_RETRY_DELAY:-30}
+
+EMAIL_ADDRESS=${EMAIL_ADDRESS:-"$(whoami)@sarao.ac.za"}
+send_email() {
+  local subject=$1
```
vincentseotlo: I am of the opinion we can use the $1, $2 directly,

scratch/benchmarks/runcalibratexbgpu.sh
```diff
@@ -0,0 +1,71 @@
...
+  local body=$2
+  ssmtp ${EMAIL_ADDRESS}<<EOF
+Subject: ${subject}
+${body}
+EOF
+}
+
+is_ssh_failure() {
+  local logfile=$1
```
vincentseotlo: Same as above, I think we can use $1 directly

scratch/benchmarks/aggregate_runs.py
```diff
@@ -0,0 +1,79 @@
...
+            line = line.strip()
+            if not line:
+                continue
+            parts = line.split()
+            try:
+                freq = float(parts[0])
+            except ValueError:
+                print(f"Warning: could not convert {parts[0]} to float, skipping line: {line} in {path}")
+                continue
+            vals = [int(v) for v in parts[-3:]]
```
vincentseotlo, line 69: can we confirm we got length == 3 here

vincentseotlo (changes requested):
Just some minors, but great stuff overall.

scratch/benchmarks/aggregate_runs.py
EatonEmmerich: should be == 4 (we don't use the freq value here, we just take the last 3 as int)

vincentseotlo (approved):
LGTM. 

Thanks, I also noted the length check above.

PR #1061, opened 2026-06-23
Allow for per-output send-enabled in xbgpu

amishatishpatel:
Specifically add this to the `--beam` and `--corrprod` command-line CSVs.

Also, tweak the unit tests accordingly to either maintain previous functionality
or test the updated configuration syntax.

Re: The two unchecked points below
- I'm not entirely sure if the **design** has changed, and
- Similarly, I'm not sure where the line is to declare and _Interface change_ between product controller and engine.

Checklist (if not applicable, edit to add `(N/A)` and mark as done):

- [x] (n/a) If dependencies are added/removed: update `pyproject.toml` and `.pre-commit-config.yaml`.
- [x] (n/a) If new source files are added: add copyright notices dated with the current year.
- [x] (n/a) If qualification tests are changed: attach or link to a sample qualification report.
- [x] If design has changed: ensure documentation is up to date.
- [x] (n/a) If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [x] (n/a) If the interface between the product controller and an engine has changed, update the
      VERSION constant for the engine (update the major number if sensors/requests have been
      removed/changed, minor number if only backwards-compatible changes have been done).
      Also, update doc/control.rst with the new version.

Closes NGC-1990.

PR #1071, opened 2026-07-27
Update send enabled usages for correlator startup

amishatishpatel:
As per @bmerry's comment on NGC-1990, this PR ensures both `scratch/sim_correlator.py` and
the qualification tests are updated according to #1061 and https://github.com/ska-sa/katsdpcontroller/pull/845

Leaving this as draft as I still need to run do a full qualification run with the branch in its current state.

@EatonEmmerich mainly an FYI in the hopes I haven't made your life more difficult with whatever you're working on.

- [x] If qualification tests are changed: attach or link to a sample qualification report - https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/627/ (PDF report is too large for GitHub)
- [x] If design has changed: ensure documentation is up to date --> This has been updated in #1061.

Contributes to: NGC-1990.

### DIFF
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -428,7 +428,23 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
 
 
 @pytest.fixture
-async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict) -> list[str]:
+async def capture_stop_streams(cbf_config: dict, vlbi: bool) -> list[str]:
+    """List of streams for which capture-stop will automatically be issued."""
+    out = []
+    for name, conf in cbf_config["outputs"].items():
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    if vlbi:
+        vlbi_config = cbf_config["outputs"]["tied-array-resampled-voltage"]
+        for beam_stream in vlbi_config["src_streams"]:
+            logger.info(f"Beam {beam_stream} required for VLBI")
+            out.remove(beam_stream)
+    return out
+
+
+@pytest.fixture
+async def capture_start_streams(request: pytest.FixtureRequest, capture_stop_streams: list[str]) -> list[str]:
     """List of streams for which capture-start will automatically be issued."""
     no_capture_start: set[str] = set()
     for marker in request.node.iter_markers("no_capture_start"):
@@ -437,8 +453,8 @@ async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict
         no_capture_start.update(marker.args)
 
     out = []
-    for name, conf in cbf_config["outputs"].items():
-        if name not in no_capture_start and conf["type"] in _CAPTURE_TYPES:
+    for name in capture_stop_streams:
+        if name not in no_capture_start:
             out.append(name)
     return out
 
@@ -527,6 +543,7 @@ async def cbf(
     cbf_mode_config: dict,
     core_allocator: CoreAllocator,
     capture_start_streams: list[str],
+    capture_stop_streams: list[str],
     tied_array_channelised_voltage_receive_streams: list[str],
     pdf_report: Reporter,
 ) -> AsyncGenerator[CBFRemoteControl, None]:
@@ -588,9 +605,8 @@ async def cbf(
 
     yield cbf
 
-    for name, conf in cbf.config["outputs"].items():
-        if conf["type"] in _CAPTURE_TYPES:
-            await pcc.request("capture-stop", name)
+    for name in capture_stop_streams:
+        await pcc.request("capture-stop", name)
 
 
 @pytest.fixture
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -436,10 +436,11 @@ async def capture_stop_streams(cbf_config: dict, vlbi: bool) -> list[str]:
             out.append(name)
 
     if vlbi:
-        vlbi_config = cbf_config["outputs"]["tied-array-resampled-voltage"]
-        for beam_stream in vlbi_config["src_streams"]:
-            logger.info(f"Beam {beam_stream} required for VLBI")
-            out.remove(beam_stream)
+        for conf in cbf_config["outputs"].values():
+            if conf["type"].startswith("gpucbf."):
+                presence_filter = filter(lambda src_stream: src_stream in out, conf["src_streams"])
+                for stream in presence_filter:
+                    out.remove(stream)
     return out
 
 
--- a/qualification/conftest.py
+++ b/qualification/conftest.py
@@ -428,19 +428,17 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
 
 
 @pytest.fixture
-async def capture_stop_streams(cbf_config: dict, vlbi: bool) -> list[str]:
+async def capture_stop_streams(cbf_config: dict) -> list[str]:
     """List of streams for which capture-stop will automatically be issued."""
     out = []
     for name, conf in cbf_config["outputs"].items():
         if conf["type"] in _CAPTURE_TYPES:
             out.append(name)
 
-    if vlbi:
-        for conf in cbf_config["outputs"].values():
-            if conf["type"].startswith("gpucbf."):
-                presence_filter = filter(lambda src_stream: src_stream in out, conf["src_streams"])
-                for stream in presence_filter:
-                    out.remove(stream)
+    for conf in cbf_config["outputs"].values():
+        for stream in conf.get("src_streams", []):
+            if stream in out:
+                out.remove(stream)
     return out
 
 
--- a/qualification/tied_array_channelised_voltage/test_steady_state.py
+++ b/qualification/tied_array_channelised_voltage/test_steady_state.py
@@ -14,22 +14,7 @@
 # limitations under the License.
 ################################################################################
 
-"""Test capture-start following state change requests.
-
-.. note::
-
-    capture-start requests cannot be issued to tied-array-channelised-voltage
-    streams that feed tied-array-resampled-voltage (VLBI) streams.
-    :meth:`._test_capture_start` filters out any such streams from the list
-    to which capture-start requests are issued.
-
-    Due to the way the CBF config is generated, tied-array-channelised-voltage-{0x, 0y}
-    are always used as src_streams for a tied-array-resampled-voltage stream. These two
-    beam streams are always first in the list of stream names for a
-    :class:`.TiedArrayChannelisedVoltageReceiver`. As a result, the tests herein only
-    only issue state change requests to the last beam stream, which is guaranteed to
-    not be a src_stream for any tied-array-resampled-voltage stream.
-"""
+"""Test capture-start following state change requests."""
 
 import asyncio
 from collections.abc import Awaitable, Callable
@@ -43,9 +28,28 @@ from ..cbf import CBFRemoteControl
 from ..recv import TiedArrayChannelisedVoltageReceiver
 
 
+@pytest.fixture
+def streams_in_use(
+    receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver, capture_stop_streams: list[str]
+) -> list[str]:
+    """The streams that are both in the receiver and in `capture_stop_streams`."""
+    return sorted(list(set(receive_tied_array_channelised_voltage.stream_names) & set(capture_stop_streams)))
+
+
+@pytest.fixture
+def stream_index_under_test(
+    streams_in_use: list[str],
+    receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver,
+    stream_index_in_use: int = 0,
+) -> int:
+    """Index of the first stream in `streams_in_use` in the receiver."""
+    return receive_tied_array_channelised_voltage.stream_names.index(streams_in_use[stream_index_in_use])
+
+
 async def _test_capture_start(
     cbf: CBFRemoteControl,
     receiver: TiedArrayChannelisedVoltageReceiver,
+    streams_in_use: list[str],
     pdf_report: Reporter,
     prepare: Callable[[], Awaitable],
 ) -> np.ndarray:
@@ -54,6 +58,13 @@ async def _test_capture_start(
     Each test provides a callback that issues a request to the product
     controller. The received data is returned for the test to do
     verification.
+
+    capture-start requests cannot be issued to tied-array-channelised-voltage
+    streams that feed tied-array-resampled-voltage (VLBI) streams. `streams_in_use`
+    is filtered to exclude such streams.
+
+    The tests are run with the assumption that the first stream in `streams_in_use`
+    is the stream under test.
     """
     pcc = cbf.product_controller_client
 
@@ -69,7 +80,7 @@ async def _test_capture_start(
     pdf_report.step("Wait for injected signal to reach XB-engines Tx.")
     # Only need to query one stream, since it's the same engine backing
     # all of them.
-    stream_name = receiver.stream_names[0]
+    stream_name = streams_in_use[0]
     for _ in range(30):
         tasks = []
         async with asyncio.TaskGroup() as tg:
@@ -88,17 +99,9 @@ async def _test_capture_start(
     await prepare()
 
     pdf_report.step("Capture and verify output")
-    # Make a copy of this list as the metohd (and its fixtures) are reused between tests
-    capture_start_streams = receiver.stream_names.copy()
-    # Note: If CBF has a VLBI stream, we cannot issue ?capture-start to its
-    # src beam streams.
-    vlbi_config = cbf.config["outputs"].get("tied-array-resampled-voltage", None)
-    if vlbi_config is not None:
-        for src_beam_stream in vlbi_config["src_streams"]:
-            capture_start_streams.remove(src_beam_stream)
 
     async with asyncio.TaskGroup() as tg:
-        for stream in capture_start_streams:
+        for stream in streams_in_use:
             tg.create_task(pcc.request("capture-start", stream))
     # We use dsim_timestamp as a minimum to ensure that we're not receiving
     # data from a *previous* capture-start/stop.
@@ -111,6 +114,8 @@ async def _test_capture_start(
 async def test_beam_quant_gains_capture_start(
     cbf: CBFRemoteControl,
     receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver,
+    streams_in_use: list[str],
+    stream_index_under_test: int,
     pdf_report: Reporter,
 ) -> None:
     """Test that beam-quant-gains issued before capture-start is not delayed.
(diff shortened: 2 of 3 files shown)


amishatishpatel:
I'm not sure why it's still showing 13 files changed. When you do have a look, please only review the first and last commit shown here.

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -14,7 +14,22 @@
...
+    streams that feed tied-array-resampled-voltage (VLBI) streams.
+    :meth:`._test_capture_start` filters out any such streams from the list
+    to which capture-start requests are issued.
+
+    Due to the way the CBF config is generated, tied-array-channelised-voltage-{0x, 0y}
+    are always used as src_streams for a tied-array-resampled-voltage stream. These two
+    beam streams are always first in the list of stream names for a
+    :class:`.TiedArrayChannelisedVoltageReceiver`. As a result, the tests herein only
+    only issue state change requests to the last beam stream, which is guaranteed to
+    not be a src_stream for any tied-array-resampled-voltage stream.
```
amishatishpatel: If I'm completely honest, I'm not overly happy with this story I've had to spin here. I had moved to using separate fixture that had VLBI src-streams filtered out, but then pytest started complaining about "recursive dependency involving fixture `cbf` detected".

I opted for annoying you all with this instead of delving into that error (for now). I know this relies on quite a few assumptions, but it felt like the simpler of two evils right now.

As always, happy to change/overhaul to be a bit less... hardcoded.

amishatishpatel: Also, slightly related, I was planning on tweaking the [tacv-receive-streams fixture](https://github.com/ska-sa/katgpucbf/blob/main/qualification/conftest.py#L499-L508) to subscribe to the "last _n_" streams, rather than the "first _n_" - if that makes sense
- i.e. `stream_names[:max_streams]` --> `stream_names[-max_streams:]`

Although that would've likely had unintended side-effects if/when a test specifically needed beams 0x and 0y.

amishatishpatel:
Oh dear, the qualification run in the lab has failed, and probably points to me needing to address that naive method of updating which beam stream gets the ?beam requests
- https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/624/testReport/

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -14,7 +14,22 @@
...
+    capture-start requests cannot be issued to tied-array-channelised-voltage
+    streams that feed tied-array-resampled-voltage (VLBI) streams.
+    :meth:`._test_capture_start` filters out any such streams from the list
+    to which capture-start requests are issued.
+
+    Due to the way the CBF config is generated, tied-array-channelised-voltage-{0x, 0y}
+    are always used as src_streams for a tied-array-resampled-voltage stream. These two
+    beam streams are always first in the list of stream names for a
+    :class:`.TiedArrayChannelisedVoltageReceiver`. As a result, the tests herein only
+    only issue state change requests to the last beam stream, which is guaranteed to
```
EatonEmmerich:
```suggestion
    :class:`.TiedArrayChannelisedVoltageReceiver`. As a result, the tests herein only
    issue state change requests to the last beam stream, which is guaranteed to
```

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -73,8 +88,17 @@ async def _test_capture_start(
     await prepare()
 
     pdf_report.step("Capture and verify output")
+    # Make a copy of this list as the metohd (and its fixtures) are reused between tests
```
EatonEmmerich:
```suggestion
    # Make a copy of this list as the method (and its fixtures) are reused between tests
```

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -73,8 +88,17 @@ async def _test_capture_start(
...
+    capture_start_streams = receiver.stream_names.copy()
+    # Note: If CBF has a VLBI stream, we cannot issue ?capture-start to its
+    # src beam streams.
```
EatonEmmerich: Question: Could we no just issue a capture start, and if the response error is "already capturing" ignore the error?

qualification/conftest.py
```diff
@@ -428,7 +428,23 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
...
+    out = []
+    for name, conf in cbf_config["outputs"].items():
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    if vlbi:
+        vlbi_config = cbf_config["outputs"]["tied-array-resampled-voltage"]
+        for beam_stream in vlbi_config["src_streams"]:
+            logger.info(f"Beam {beam_stream} required for VLBI")
+            out.remove(beam_stream)
```
EatonEmmerich: Question: Wouldn't the capture stop work if we simply enforced order of stopping instead (ie. stop vlbi stream before stopping tacv?)

qualification/conftest.py
```diff
@@ -428,7 +428,23 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
...
-async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict) -> list[str]:
+async def capture_stop_streams(cbf_config: dict, vlbi: bool) -> list[str]:
+    """List of streams for which capture-stop will automatically be issued."""
+    out = []
+    for name, conf in cbf_config["outputs"].items():
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    if vlbi:
+        vlbi_config = cbf_config["outputs"]["tied-array-resampled-voltage"]
```
bmerry: This is hard-coding an assumption about how many VLBI streams we have and what they're called. Let's generalise it:
- Iterate over `cbf_config["outputs"].items()` again.
- For each item, remove all its source streams from `out` (if they're present).

qualification/conftest.py
bmerry: If for some reason we had two VLBI streams using the same beam, this would attempt to remove it from `out` twice, and the second time would raise ValueError.

scratch/sim_correlator.py
```diff
@@ -336,7 +339,8 @@ def main(argv: Sequence[str] | None = None) -> int:
             f.write("\n")  # json.dump doesn't write a final newline
         return 0
     else:
-        return asyncio.run(issue_config(args.controller, args.port, args.name, config))
+        # Pass a boolean that indicates VLBI or not
```
bmerry: That is a bit of a useless comment. The code already tells you that.

scratch/sim_correlator.py
```diff
@@ -315,6 +315,9 @@ async def issue_config(host: str, port: int, name: str, config: dict) -> int:
         product_client = await aiokatcp.Client.connect(product_host, product_port)
         for output_name, output in config["outputs"].items():
             if output["type"] in CAPTURE_START_TYPES:
+                if has_vlbi and "narrow0-tied-array-channelised-voltage" in output_name:
+                    # vgpu src-streams are pre-enabled by katsdpcontroller
+                    continue
```
bmerry: I'd prefer to make this more generic again: for every output stream whose type starts with `gpucbf.` (so as to exclude the SDP streams), remove all its source streams from a list of streams to start.

qualification/tied_array_channelised_voltage/test_steady_state.py
bmerry: I agree, it's not that pretty.

You don't necessarily need a separate fixture to hold the filtered stream list. It could just be an attribute on CBFRemoteControl that is populated by CBFRemoteControl.connect.

Alternatively, you have a `capture_stop_streams` fixture which contains all the streams that would normally be started. You could just take the intersection between that and `receiver.stream_names` to get the list of streams to start for the receiver.

qualification/tied_array_channelised_voltage/test_steady_state.py
bmerry: Typo "metohd"

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -14,7 +14,22 @@
...
 
-"""Test capture-start following state change requests."""
+"""Test capture-start following state change requests.
+
+.. note::
+
+    capture-start requests cannot be issued to tied-array-channelised-voltage
+    streams that feed tied-array-resampled-voltage (VLBI) streams.
+    :meth:`._test_capture_start` filters out any such streams from the list
+    to which capture-start requests are issued.
```
bmerry: This note sounds like it belongs on the `_test_capture_start` function.

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -14,7 +14,22 @@
...
 ################################################################################
 
-"""Test capture-start following state change requests."""
+"""Test capture-start following state change requests.
+
+.. note::
+
+    capture-start requests cannot be issued to tied-array-channelised-voltage
+    streams that feed tied-array-resampled-voltage (VLBI) streams.
+    :meth:`._test_capture_start` filters out any such streams from the list
```
bmerry: It's a function, not a method.

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -101,10 +125,10 @@ async def test_beam_quant_gains_capture_start(
     async def prepare() -> None:
         pdf_report.step("Send request.")
         pdf_report.detail("Set beam-quant-gains to 0 on first beam.")
```
bmerry: Comment is out of date.

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -188,5 +188,5 @@ async def prepare() -> None:
     # We use data[2] instead of data[1], because data[1] is the other
```
bmerry: This comment is out of date.

bmerry (changes requested):

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -43,9 +28,28 @@
 from ..recv import TiedArrayChannelisedVoltageReceiver
 
 
+@pytest.fixture
+def streams_in_use(
+    receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver, capture_stop_streams: list[str]
+) -> list[str]:
+    """The streams that are both in the receiver and in `capture_stop_streams`."""
+    return sorted(list(set(receive_tied_array_channelised_voltage.stream_names) & set(capture_stop_streams)))
```
bmerry: AFAIK `sorted` takes any iterable, so you don't need to force it to a list first.

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -43,9 +28,28 @@
...
+
+
+@pytest.fixture
+def stream_index_under_test(
+    streams_in_use: list[str],
+    receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver,
+    stream_index_in_use: int = 0,
```
bmerry: Having a default for an argument to a fixture doesn't seem to make sense to me: unless you're also calling this as a regular function (which it doesn't seem you are) it's just a constant pretending to be an argument. I'd just hard-code the 0 into the body (that's what the docstring claims the function does anyway).

qualification/tied_array_channelised_voltage/test_steady_state.py
```diff
@@ -179,14 +192,16 @@ async def test_beam_delays_capture_start(
...
-    # tolerance allows for some rounding error plus dithered quantisation.
-    np.testing.assert_allclose(data[-1], -data[-3], atol=2)
+        await cbf.product_controller_client.request("beam-delays", streams_in_use[0], *delays)
+
+    data = await _test_capture_start(cbf, receiver, streams_in_use, pdf_report, prepare)
+    assert (
+        np.sum(data[stream_index_under_test] != 0) >= data[stream_index_under_test].size // 2
+    )  # Should be mostly non-zero
+    # We use a data subset two indices over because the immediate neighbour
+    # is the other polarisation and so experiences different F-engine dithering.
```
bmerry, line 203: This might index off the end of the array: the implementation of `tied_array_channelised_voltage_receive_streams` only requires that we have 2 dual-pol beams (4 tacv streams). If streams 0 and 1 are used for VLBI, then `stream_index_under_test` will be 2 and `stream_index_under_test + 2` will be 4.

I don't think the test requires that the reference beam is part of streams_in_use i.e. it's fine if it's a beam used for VLBI. So if `stream_index_under_test >= 2` then we could use `stream_index_under_test - 2` instead.

qualification/conftest.py
```diff
@@ -428,7 +428,24 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
...
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    if vlbi:
+        for conf in cbf_config["outputs"].values():
+            if conf["type"].startswith("gpucbf."):
+                presence_filter = filter(lambda src_stream: src_stream in out, conf["src_streams"])
+                for stream in presence_filter:
+                    out.remove(stream)
+    return out
```
bmerry: Since you're already using a loop, you may as well just do the filtering inside the loop, and it'll be easier to follow:
```suggestion
                for stream in conf["src_streams"]:
                    if stream in out:
                        out.remove(stream)
    return out
```

qualification/conftest.py
```diff
@@ -428,7 +428,24 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
...
 @pytest.fixture
-async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict) -> list[str]:
+async def capture_stop_streams(cbf_config: dict, vlbi: bool) -> list[str]:
+    """List of streams for which capture-stop will automatically be issued."""
+    out = []
+    for name, conf in cbf_config["outputs"].items():
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    if vlbi:
```
bmerry: No need for `if vlbi` any more; this should be valid in general.

scratch/sim_correlator.py
```diff
@@ -313,13 +314,20 @@ async def issue_config(host: str, port: int, name: str, config: dict, has_vlbi:
...
-                    continue
-                print(f"Enabling {output_name} transmission...")
-                await product_client.request("capture-start", output_name)
+            if output["type"] in GPUCBF_DATA_STREAMS:
+                capture_start_streams.append(output_name)
+                src_streams.extend(output["src_streams"])
+
+        for output_name, output in config["outputs"].items():
+            if output["type"].startswith("gpucbf.") and output_name in src_streams:
+                capture_start_streams.remove(output_name)
```
bmerry: I think this could try to remove an element from capture_start_streams twice if it was used twice.

scratch/sim_correlator.py
```diff
@@ -39,7 +39,8 @@
 from katgpucbf.meerkat import BANDS
 
 #: Stream types for which ``?capture-start`` is valid
-CAPTURE_START_TYPES = {
+GPUCBF_DATA_STREAMS = {
+    "gpucbf.antenna_channelised_voltage",
```
bmerry: I see what you're trying to do, but you've missed a subtlety: gpucbf.antenna_channelised_voltage streams are not allowed to be stopped (or started), even if they have no downstream consumers. So running `sim_correlator.py --last-stage f` will probably go wrong now.

bmerry (changes requested):

qualification/conftest.py
amishatishpatel: You know, I was doing something vastly similar (if not exactly the same) and started thinking it's becoming too nested.

Updating now.

scratch/sim_correlator.py
amishatishpatel: Indeed indeed. Although, for the record, I was confusing the `capture_start_streams` with the list of all `src_streams`
- That is, I got to a point thinking `capture_start_streams` would have all the dsim streams in there
- But seeing as we pre-populate with all X, B and V streams, that avoids that problem/concern

scratch/sim_correlator.py
amishatishpatel: Sorry yeah I was wondering about that too. I've updated to use the same approach as in `qualification/conftest.py`

qualification/conftest.py
```diff
@@ -428,7 +428,22 @@ async def cbf_cache(pytestconfig: pytest.Config) -> AsyncGenerator[CBFCache, Non
...
-async def capture_start_streams(request: pytest.FixtureRequest, cbf_config: dict) -> list[str]:
+async def capture_stop_streams(cbf_config: dict) -> list[str]:
+    """List of streams for which capture-stop will automatically be issued."""
+    out = []
+    for name, conf in cbf_config["outputs"].items():
+        if conf["type"] in _CAPTURE_TYPES:
+            out.append(name)
+
+    for conf in cbf_config["outputs"].values():
+        for stream in conf.get("src_streams", []):
```
amishatishpatel, line 439: I realised why I did this under a `.startswith("gpucbf.")`
- digitiser streams don't have a `src_streams` attribute

amishatishpatel:
Should be done in the next 30 mins
- https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/635/console

scratch/sim_correlator.py
```diff
@@ -315,15 +314,14 @@ async def issue_config(host: str, port: int, name: str, config: dict) -> int:
...
-                src_streams.extend(output["src_streams"])
-
-        for output_name, output in config["outputs"].items():
-            if output["type"].startswith("gpucbf.") and output_name in src_streams:
-                capture_start_streams.remove(output_name)
+        for name, conf in config["outputs"].items():
+            if conf["type"] in CAPTURE_START_TYPES:
+                capture_start_streams.append(name)
+
+        for conf in config["outputs"].values():
```
bmerry, line 317: Here you do need to check if it startswith `gpucbf.`, because otherwise when specifying `--last-stage ingest` we'll remove the baseline-correlation-products stream from capture_start_streams.

bmerry (changes requested):

bmerry (approved):

EatonEmmerich (commented):
I realize only now I never submitted the comments, sorry about that.

qualification/tied_array_channelised_voltage/test_steady_state.py
amishatishpatel: If I'm honest, I did think about this approach as well. Un/Fortunately, there are a couple different scenarios where the root method (`capture_start_stop`) could raise a `FailReply(...)` - i.e. not just for this case
- https://github.com/ska-sa/katsdpcontroller/blob/NGC-1990-per-output-send-enabled/src/katsdpcontroller/product_controller.py#L1583-L1642

In that case, we'd have to match exactly against the error/exception message, which is... not great.

qualification/conftest.py
amishatishpatel: Hmmmm, that's probably a fair point. Just thinking out aloud, I recall something Bruce put into [V-engine documentation](https://katgpucbf.readthedocs.io/en/latest/vgpu.design.html#reception) where
- vgpu may not be _transmitting_, but
- is still receiving data from its src streams (beams)

Either way, it's not a guarantee that we're "done with the V-engine" when we issue `?capture-stop` to it.

bmerry: That may have been a viable approach, but not the one I went for when I wrote the ICD. The ICD says it may not be stopped. I think that was largely for consistency with how F-engines work (those may not be stopped either), which in turn was done for consistency with the SKARAB correlator.

bmerry:
@EatonEmmerich has verbally confirmed that he's happy, so I'm going to merge this.

PR #1077, opened 2026-08-04
Fix crash when there is a gap in input data

bmerry:
When there is a gap between input chunks, the output chunk is fast-forwarded to avoid referencing missing data. This is done with batch granularity, so as not to lose more output data than necessary. However, that means that the OutQueueItem is not necessarily aligned (in time) to a multiple of the output chunk size. In our unit tests it generally has been, because there is no delay model and so it ends up aligned to the *input* chunk size; but when there is a non-zero coarse delay, that's no longer the case. Meanwhile, the code for handling the dig-rms-dbfs sensor works with "windows" which are some multiple of the chunk size and aligned to the window size, and does not cope with OutQueueItems that cross a window boundary. In theory this could also be an issue if a non-zero delay was set before the first data arrived.

Fix it by changing the computation of total digitiser power from being done per-output-chunk (on the GPU) to per-output-batch. I initially experimented with doing it per-output-spectrum, but that was too expensive; even with per-output-batch it needed some experimentation to get the CUDA compiler to generate reasonable code that didn't slow things down massively. I've tested with benchmark_fgpu and found no statistically significant impact on performance:

```
./benchmark_fgpu.py -v --narrowband=1 --image harbor.cbf.mkat.karoo.kat.ac.za/dpp/katgpucbf:latest
1873.0 MHz - 1893.0 MHz

./benchmark_fgpu.py -v --narrowband=1 --image harbor.cbf.mkat.karoo.kat.ac.za/dpp/katgpucbf:NGC-2074-data-gap
1868.0 MHz - 1888.0 MHz

./benchmark_fgpu.py -v --image harbor.cbf.mkat.karoo.kat.ac.za/dpp/katgpucbf:latest
1955.0 MHz - 1975.0 MHz

./benchmark_fgpu.py -v --image harbor.cbf.mkat.karoo.kat.ac.za/dpp/katgpucbf:NGC-2074-data-gap
1957.0 MHz - 1977.0 MHz
```

Add a non-zero delay to the existing unit tests for missing data. That made it a lot more complicated unfortunately.

Closes NGC-2074.

### DIFF
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -1032,7 +1032,7 @@ class Pipeline:
         present
             Whether each output batch is fully present
         """
-        batch_samples = out_item.capacity // len(present) * out_item.spectra_samples
+        batch_samples = self.output.spectra_per_heap * self.output.spectra_samples
         for pol, (accum, trg) in enumerate(zip(dig_total_power_accums, dig_total_power, strict=True)):
             for i, p in enumerate(present):
                 power: int | None = trg[i] if p else None
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -369,7 +369,7 @@ class OutQueueItem(QueueItem):
     spectra: accel.DeviceArray
     #: Output saturation count, per pol
     saturated: accel.DeviceArray
-    #: Output sum of squared samples, per pol
+    #: Output sum of squared samples, per batch and pol
     dig_total_power: accel.DeviceArray | None
     #: Per-spectrum fine delays
     fine_delay: MappedArray
@@ -573,6 +573,7 @@ class Pipeline:
             engine.send_sample_bits,
             output.dither,
             narrowband=narrowband_config,
+            total_power_spectra=output.spectra_per_heap,
         )
         seed = SystemRandom().randrange(2**ENGINE_DITHER_SEED_BITS)
         self._compute = template.instantiate(
@@ -1024,7 +1025,7 @@ class Pipeline:
         dig_total_power_accums
             Accumulators tracking long-term digitiser total power (one per polarisation)
         dig_total_power
-            The total power per spectrum and polarisation in `out_item`
+            The total power per batch and polarisation in `out_item`
         out_item
             The current :class:`OutQueueItem`
         """
--- a/src/katgpucbf/fgpu/engine.py
+++ b/src/katgpucbf/fgpu/engine.py
@@ -1017,6 +1017,7 @@ class Pipeline:
         dig_total_power_accums: list[Accum],
         dig_total_power: accel.HostArray,
         out_item: OutQueueItem,
+        present: np.ndarray,
     ) -> None:
         """Update digitiser power sensors.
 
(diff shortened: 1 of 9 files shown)


bmerry:
I'm requesting @vincentseotlo as primary reviewer. @EatonEmmerich is added more for interest / learning.

src/katgpucbf/fgpu/kernels/pfb_fir.mako
```diff
@@ -98,22 +99,23 @@ KERNEL REQD_WORK_GROUP_SIZE(WGS, 1, 1) void pfb_fir(
...
-    in += pol * in_stride;
-    out += pol * out_stride;
 
     // Figure out where this thread has to work.
     int lid = get_local_id(0);
-    int pos = group_x * WGS + lid; // pos is the position within the step (i.e. spectrum) that this thread will work on.
-    int offset = group_y * stepy + pos; // This thread probably doesn't work on the very beginning of the data, so we
-                                        // make the indexing easier for ourselves later.
+    // pos is the position within the step (i.e. spectrum) that this thread will work on.
+    int pos = get_group_id(0) * WGS + lid;
```
vincentseotlo, line 106: I think we can keep the standard of using `group_x`, so we're in line with `group_y`. Granted group_y is `shifted` by `stepy`.

src/katgpucbf/fgpu/pfb.py
```diff
@@ -101,6 +105,12 @@ def __init__(
         self.unzip_factor = unzip_factor
         self.complex_input = complex_input
         self.n_pols = n_pols
+        if complex_input:
+            # Ensures that the given value definitely has no effect. Choosing 1
+            # ensures that tests that it divides into other values are
+            # satisfied.
+            total_power_spectra = 1
+        self.total_power_spectra = total_power_spectra
```
vincentseotlo, line 113: Is there any danger in just saying `self.total_power_spectra = 1`, then line 138: `"total_power_spectra": self.total_power_spectra,`

src/katgpucbf/fgpu/engine.py
```diff
@@ -1024,31 +1026,34 @@ def _update_dig_power_sensors(
...
+                        fs = ((1 << (self.engine.recv_layout.sample_bits - 1)) - 1) ** 2 / 2
+                        avg_power = measurement.total / (measurement.end_timestamp - measurement.start_timestamp) / fs
+                        # If for some reason there's zero power, avoid reporting
+                        # -inf dB by assigning the most negative representable value
+                        avg_power_db = 10 * math.log10(avg_power) if avg_power else np.finfo(np.float64).min
+                        sensor.set_value(avg_power_db, timestamp=update_timestamp)
+                    else:
+                        sensor.set_value(
+                            np.finfo(np.float64).min, status=aiokatcp.Sensor.Status.FAILURE, timestamp=update_timestamp
+                        )
```
EatonEmmerich, line 1056: what do we expect to happen if we accumulate with power=None on an existing value in the accumulator?

test/fgpu/test_engine.py
```diff
@@ -98,9 +98,9 @@ def jones_per_batch(channels: int, request: pytest.FixtureRequest) -> int:
 
 
 @pytest.fixture
-def dig_rms_dbfs_window_chunks() -> int:
+def dig_rms_dbfs_window_batches() -> int:
     """Number of chunks per window for ``dig-rms-dbfs`` sensors."""
-    return 2
+    return 5
```
EatonEmmerich: why not just use a constant?

src/katgpucbf/fgpu/kernels/pfb_fir.mako
```diff
@@ -145,47 +147,60 @@ KERNEL REQD_WORK_GROUP_SIZE(WGS, 1, 1) void pfb_fir(
...
     // We'll be at our most memory-bandwidth-efficient if rows >> TAPS.
     // Launching ~256K threads should ensure this.
-    for (int i = 0; i < rows; i++)
+% if complex_input:
+    for (int i = spectrum0; i < n; i++)
     {
-        // Load the raw data for the sample
-        sample_t sample = unpack_read(&unpack);
-        unpack_advance(&unpack, step);
+        {  // Block just to balance things with the !complex_input case.
```
EatonEmmerich: nit(non-blocking): missed the `!` here first time I read the comment and got confused.
Suggest writing it as `not complex_input` case.

src/katgpucbf/fgpu/engine.py
```diff
@@ -1106,7 +1111,7 @@ async def run_transmit(self) -> None:
                 await async_wait_for_events([download_marker])
 
             if dig_total_power is not None:
-                self._update_dig_power_sensors(dig_total_power_windows, dig_total_power, out_item)
+                self._update_dig_power_sensors(dig_total_power_windows, dig_total_power, out_item, chunk.present)
```
EatonEmmerich, line 1114: can we pass the `batch_samples` in here instead then calculate it from the `spectra_per_heap`  like we do in `_dig_rms_dbfs_window_samples`?

src/katgpucbf/fgpu/kernels/pfb_fir.mako
bmerry: `pos` is not a group-wide value though, because it depends on the local (thread) ID.

src/katgpucbf/fgpu/engine.py
bmerry: Do you mean if we call `accum.add(..., power=None)`? The window containing [start_timestamp, end_timestamp) will have a power of `None` (since `None` means "don't know" in this context, and anything + don't-know = don't-know). There are a few cases for what `accum.add` returns, but I *think* for all of them, if power=None and it returns a measurement, that measurement will have `total` of `None`.

src/katgpucbf/fgpu/engine.py
bmerry: Not entirely following the suggestion, but I'll update the calculation of `batch_samples` inside `_update_dig_power_sensors` to use the same formula as `_dig_rms_dbfs_window_samples`.

src/katgpucbf/fgpu/pfb.py
bmerry: Not quite sure what the suggestion is. Are you suggesting that here I do
```python
if complex_input:
    self.total_power_spectra = 1
else:
    self.total_power_spectra = total_power_spectra
```

That would certainly work, but I'm not seeing the advantage; the (minor) disadvantage of doing it this way is that someone updating the code might accidentally use `total_power_spectra` when they should have used `self.total_power_spectra`, whereas the way I've done it, the two are kept the same so there is no danger of using the wrong value.

test/fgpu/test_engine.py
bmerry: Good idea. I probably had some idea that I might need to override this fixture for some specific tests, but it hasn't happened, so I've simplified things with a constant.

vincentseotlo (approved):
LGTM 👍

src/katgpucbf/fgpu/engine.py
EatonEmmerich: that was the sentiment, yes. :bow:

src/katgpucbf/fgpu/engine.py
EatonEmmerich: The reason I asked was that I think we can simplify it a little more by not needing the `if p else None` as well as the `if measurement.total is not None:`

Assuming I understand this correctly:
* if not p, power is None 
 * we add None to accum. and now total is None.
 * we check if accum.total is not none.
  * set value
 * else set failure

could just be:
* if p
 * add to accum and set value
* else
 * we set failure and add none to accum

EatonEmmerich (approved):
small non blocking nit.

LGTM otherwise

src/katgpucbf/fgpu/engine.py
bmerry: I might be misunderstanding your suggestion, but it sounds like you're setting the sensor on every loop iteration (either "set value" or "set failure" which is not the goal: we only want to set the sensor once per window, as otherwise we're going to spam sensor updates. And the decoupling between `p` and the sensor updates makes sense in my head, because `p` refers to the current window, while `measurement` is a (potential) measurement for the previous window.

bmerry:
> small non blocking nit.

I'm going to merge anyway so that this bug fix can catch the bus to get deployed tomorrow, but feel free to open a ticket.

src/katgpucbf/fgpu/engine.py
bmerry: Actually it's even a bit more complicated than my explanation, because `measurement` could refer to either the same or a previous window than `p`. Basically, I prefer to think of the accumulator as a black box which we throw per-batch information into, and now and then it'll disgorge a Measurement to tell you something about the past, but I'd rather not have this piece of code try to know or care too much about the internals of exactly when Measurements might come out and how they might relate to the information that went in.

PR #1080, opened 2026-08-04
NGC-2023 Test mean-power metrics on tied array resampled voltage stream

EatonEmmerich:

- [x] `(N/A)`  If design has changed: ensure documentation is up to date.
- [x] `(N/A)`  If ICD-defined sensors have been added: update `fake_servers.py` in katsdpcontroller to match.
- [x] `(N/A)`  If the interface between the product controller and an engine has changed, update the

Closes NGC-2023.

[report.pdf](https://github.com/user-attachments/files/32329248/report.pdf)


EatonEmmerich:
Up for review but still blocked by changes regarding vlbi-delay required

scratch/sim_correlator.py
```diff
@@ -237,7 +237,9 @@ def generate_tied_array_resampled_voltage(args: argparse.Namespace, outputs: dic
     if args.vlbi:
         outputs["tied-array-resampled-voltage"] = {
             "type": "gpucbf.tied_array_resampled_voltage",
-            "src_streams": [f"narrow0-tied-array-channelised-voltage-0{pol_name}" for pol_name in "xy"],
+            "src_streams": [
+                f"narrow0-tied-array-channelised-voltage-{i}{pol_name}" for i in range(1) for pol_name in "xy"
+            ],
```
EatonEmmerich:
```suggestion
            "src_streams": [f"narrow0-tied-array-channelised-voltage-0{pol_name}" for pol_name in "xy"],
```

bmerry (commented):
Getting @amishatishpatel to take a look first.

scratch/sim_correlator.py
amishatishpatel: Yeah that's fair enough 😅

src/katgpucbf/utils.py
```diff
@@ -427,7 +427,7 @@ def unix_to_adc(self, timestamp: float) -> float:
         return (timestamp - self.sync_time) * self.adc_sample_rate
 
     def adc_to_unix(self, samples: float) -> float:
-        """Convert an ADC sample count to a UNIX timstamp."""
+        """Convert an ADC sample count to a UNIX timstamp (seconds)."""
```
amishatishpatel: I mean, while you're here can I ask you to fix that typo, please
- _timstamp_ --> **timestamp**

pyproject.toml
```diff
@@ -66,6 +66,7 @@ test = [
     "baseband",
     "katsdpsigproc[CUDA]",
     "matplotlib",
+    "prometheus-api-client>=0.7.0",
```
amishatishpatel: Did something happen that resulted in this minimum-version requirement?

qualification/recv.py
```diff
@@ -642,3 +645,23 @@ def chunk_place(data_ptr, data_size, user_data_ptr):
...
+class TiedArrayResampledVoltageReceiver:
+    """Receive tied-array-resampled-voltage streams from the V-engines."""
+
+    def __init__(
+        self,
+        cbf: CBFRemoteControl,
+        stream_names: Sequence[str],
+        interface_address: str,
+    ) -> None:
+        self.stream_names = stream_names
```
amishatishpatel:
```suggestion
self.stream_names = list(stream_names)
```

amishatishpatel: Mainly for consistency with the `XBReceiver` above. I think it largely ensures mypy and friends are happy with you doing list-type actions on `self.stream_names` later on.

qualification/recv.py
```diff
@@ -642,3 +645,23 @@ def chunk_place(data_ptr, data_size, user_data_ptr):
...
+
+
+class TiedArrayResampledVoltageReceiver:
+    """Receive tied-array-resampled-voltage streams from the V-engines."""
+
+    def __init__(
+        self,
+        cbf: CBFRemoteControl,
+        stream_names: Sequence[str],
+        interface_address: str,
```
amishatishpatel: I see why you've put it here, but I reckon we can remove it for now (add back in when we need it later).

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,170 @@
...
+            mean_power_sensor_values[i, j] = sensor_watcher.sensors[name].value
+            mean_power_sensor_timestamps[i, j] = sensor_watcher.sensors[name].timestamp
+
+        return all(
+            [
+                sensor_watcher.sensors[name].timestamp >= min_sensor_time
+                and sensor_watcher.sensors[name].value == pytest.approx(tacv_power, rel=5e-3)
+                for name in sensor_names
+            ]
+        )
```
amishatishpatel, line 145: I see why you've done this, but something feels odd to me. For some reason it feels like you're un/intentionally using two different sets of values here.
1. Those you capture on lines 137 and 138, and
2. Those you use in `all(...)` 

Maybe the cadence/frequency of update means each call to `sensor_watcher.sensors[...]` returns the same values in both cases (?).

I'll interrogate further and concede/double-down accordingly.

amishatishpatel: Also, would it be quicker to use `np.all(...)` over the numpy ndarrays' results?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,170 @@
...
+    pcc = cbf.product_controller_client
+
+    async with asyncio.TaskGroup() as tg:
+        pdf_report.step("Setup signal generator and gains.")
+        for i, name in enumerate(receive_tied_array_channelised_voltage.stream_names):
+            gains = [0.0] * len(receive_tied_array_channelised_voltage.source_indices[i])
+            gains[0] = 1.0
+            tg.create_task(pcc.request("beam-weights", name, *gains))
+
+    pdf_report.detail(f"Set beam weights and delay to {gains}.")
```
amishatishpatel: I started typing up a full `suggestion`, but that became annoying to edit in this textbox.
TL;DR Have a look at [tacv/test_steady_state.py](https://github.com/ska-sa/katgpucbf/blob/main/qualification/tied_array_channelised_voltage/test_steady_state.py) for "more informative" ways to add things to `report_pdf.{step, detail}` - screenshot at the end of the comment
1. `weights` instead of `gains` because the beams have a `?beam-quant-gain` request (which is not what's being used here)
2. It's useful to have updates on each beam's weights, but probably not wise to print out the entire list (e.g. 80 antennas' weights will look untidy)
  a. I don't quite have neat suggestion to replace your line 101.
  b. That being said, it's a list of 80 items printed in ~one line, rather than the Python `__repr__` with a `\n` after each element...
3. I see why your `pdf_report.step` says you're setting up `signal generator and gains`
  a. To my eye, you're doing the beam *weights setting before the dsim (signal gen)
  b. I reckon update the `pdf_report.step` to capture what you're doing on the following lines (separate `pdf_report.step`s is fine, I reckon)
4. Also, maybe again just a mistype, but you are not setting the beam **delay** in this test (nor the vlbi-delay)

Image

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,170 @@
...
+
+    async with asyncio.TaskGroup() as tg:
+        for dsim_name in cbf.dsim_names:
+            tg.create_task(pcc.request("dsim-signals", dsim_name, "common=wgn(0.02);common;common;"))
+    pdf_report.detail("Set dsim signals white noise.")
```
amishatishpatel: Similar to the comment above, perhaps something like
```suggestion
    dsim_signals = "common=wgn(0.02);common;common;"
    async with asyncio.TaskGroup() as tg:
        for dsim_name in cbf.dsim_names:
            tg.create_task(pcc.request("dsim-signals", dsim_name, dsim_signals))
    pdf_report.detail(f"Set dsim signals to {dsim_signals}.")
```

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Argh, my laptop froze and I lost this ~lengthy comment I was writing. Basically
- We're setting beam **weights** not _gains_ (nor delays)
  - Update the variables and log messages accordingly
- Have a look at the tests in [tacv/test_steady_state.py](https://github.com/ska-sa/katgpucbf/blob/main/qualification/tied_array_channelised_voltage/test_steady_state.py#L111-L139) for guidance/inspiration on how to log `pdf_report.{step, detail}` in a more informative way
  - e.g. Reporting beam weights entirely looks fine for small arrays, but may look a bit silly for 80A (one weight per antenna)
  - I could also be very wrong and it could be fine - let me know

Image

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Similar to the comment above, would be nice to log the actual command/config used for the dsim.
- I reckon see other tests where the `dsim_signals` are declared as a variable and used accordingly

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,170 @@
...
+
+    pdf_report.step("Wait for mean-power sensors to reach steady state.")
+    await sensor_watcher.synced.wait()  # Implicitly waits for connection too
+    time_converter = TimeConverter(receiver.sync_time, receiver.scale_factor_timestamp)
+    steady_state_unix = time_converter.adc_to_unix(await cbf.steady_state_timestamp())
+    # TODO: Because the v engine reciever is padding zeros,
+    # for now just retry with 2 second intervals since steady state is unknown.
+    min_sensor_time = steady_state_unix + receiver.power_int_time
```
amishatishpatel: Could I ask you to add the ticket number in this note, NGC-2099?

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel, line 145: Damn and blast, my comment here has disappeared as well. I was basically thinking out aloud Re: 
- Reading from/Getting sensor.{timestamp, value} twice
- Instead of using the values you'd captured on lines 137 and 138
- I don't know what the update cadence/period is of those sensors, and
- Whether reading them in quick session opens up room for failure/discrepancy

Related to that, and largely for readability,
- I would be in favour/support of doing two separate `all(...)` checks on `values` and `timestamps`, and 
- one final `all_timestamps and all_values` check

Have you come across `numpy.testing`'s suite of checks? Maybe increased accuracy/speed there as well?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,170 @@
...
+            ]
+        )
+
+    pdf_report.step("Compare mean-power sensors against TACV power.")
+    test_passed, last_sample = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
+    with check:
+        assert test_passed, f"Power does not agree to within 0.5% after {samples} retries."
+        assert tacv_power > 0.0
+
+    mean_power_sensor_timestamps = mean_power_sensor_timestamps - mean_power_sensor_timestamps[:, :1]
```
amishatishpatel: I think I figured out what this did and maybe my comment was lost. I figured it just zeroed out the first column of readings. Is that right?

qualification/recv.py
amishatishpatel: Seems we aren't using this at the moment. I reckon you can remove for now and re-add later/at some point.

scratch/sim_correlator.py
```diff
@@ -309,6 +311,13 @@ async def issue_config(host: str, port: int, name: str, config: dict) -> int:
         print(f"Product controller is at {product_host}:{product_port}")
 
         product_client = await aiokatcp.Client.connect(product_host, product_port)
+
+        if vlbi:
+            for dsim_name, dsim in config["outputs"].items():
+                if dsim["type"] == "sim.dig.baseband_voltage":
+                    tasks = await product_client.sensor_value(f"{dsim_name}.tasks", str)
+                    tasklist = json.loads(tasks)
+                    await product_client.request("dsim-signals", tasklist[0], "common=wgn(0.02); common; common;")
```
amishatishpatel: Now that I think about it, this doesn't necessarily belong with a qualification test addition/update PR.

I think it's still useful, but perhaps you could give a fair warning to users you are setting "non-zero" dsim-signals?

amishatishpatel (changes requested):
Unfortunately (for me) I lost some comments when my laptop froze. I think I re-commented on the most important/useful bits.

As always, let me know if anything is unclear. Happy to discuss through via Zoom if easier/you'd prefer.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: It removes the baseline timestamp from all the timestamps so we don't have humungus unix timestamp values on the x axis. (all time relative to first timestamp sample instead)

pyproject.toml
EatonEmmerich: artifact from ngc-1922 thanks, removing.

scratch/sim_correlator.py
EatonEmmerich: just pressing it in here automatically is probably not the best sotution anyway, I'll just revert instead.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: just mentioning there are cases where the signal string is not logged and therefore not assigned to a var.
Personally I don't think logging the dsim DSL in the qualifications directly is a good idea, but i'll opt for it now as it's what's expected.

amishatishpatel:
@EatonEmmerich Could I ask you to link to a new qualification run for this test? Specifically after your recent updates.

Please and thank you.

EatonEmmerich:
> @EatonEmmerich Could I ask you to link to a new qualification run for this test? Specifically after your recent updates.
> 
> Please and thank you.

I updated the link in the description to be the new pdf.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,172 @@
...
+    lambda_function: Callable[[int], Awaitable[bool]], max_retries: int, retry_interval: float
+) -> tuple[bool, int]:
+    """Test a subroutine with a maximum number of retries and a retry interval."""
+    sleep_period = retry_interval
+    for try_number in range(max_retries):
+        start_time = time.time()
+        if await lambda_function(try_number):
+            return True, try_number
+        sleep_period = retry_interval - (time.time() - start_time)
+        await asyncio.sleep(sleep_period)
```
amishatishpatel: I forgot to comment on this during the last review.
1. This is very cool, can't say I've seen/needed to do this before.
2. I thought there would be an issue with a negative `sleep_period`, but seems like it behaves differently to the synchronous `time.sleep()` (?)

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,172 @@
...
+async def max_retry_test(
+    lambda_function: Callable[[int], Awaitable[bool]], max_retries: int, retry_interval: float
+) -> tuple[bool, int]:
+    """Test a subroutine with a maximum number of retries and a retry interval."""
+    sleep_period = retry_interval
+    for try_number in range(max_retries):
+        start_time = time.time()
+        if await lambda_function(try_number):
+            return True, try_number
+        sleep_period = retry_interval - (time.time() - start_time)
```
amishatishpatel: A couple of admittedly slightly pedantic points.

1. I'll admit, I would've probably named this Callable `lambda_function` as well, but something about that name doesn't feel right. Other Callables in `qualification/` have been named according to their general purpose.
  a. What do you reckon about renaming this to something like `test_to_retry`?
2. Similarly, `try_number` irks me for some reason. It's not you, it's me.
  a. Maybe it's because it would read as `try_number = 0`, and English (nor Afrikaans?) has the concept of the [zeroth](https://www.merriam-webster.com/dictionary/zeroth) try - or do we?
  b. Either way, this maybe not the most important thing to ponder.
3. Why do you declare `sleep_period` at the top of the function?
  a. I don't see it used until it's updated(/re-declared?) on Line 65
  b. In my mind, it still works out if you remove that first declaration (also, `sleep_period` is only used inside the `for` loop)

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Now that I've made the comment above, have you considered the case of `asyncio.sleep(...)` essentially turning into a `no-op`?
- That is, how does that play with what you're trying to accomplish with the retries?
- Specifically, does retrying **immediately** actually cause more problems?
- I guess in this case it's just reading sensor values, but I figured you did it this way to accommodate for future uses/use-cases

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,172 @@
...
+    mean_power_sensor_values = np.zeros((len(sensor_names), samples))
+    mean_power_sensor_timestamps = np.zeros((len(sensor_names), samples))
+
+    async def wait_mean_power_steady_state(j: int) -> bool:
+        measurements = np.zeros((len(sensor_names), 2))
+        for i, name in enumerate(sensor_names):
+            measurement = sensor_watcher.sensors[name]
+            mean_power_sensor_values[i, j] = measurement.value
+            mean_power_sensor_timestamps[i, j] = measurement.timestamp
+            measurements[i] = (measurement.value, measurement.timestamp)
```
amishatishpatel: Just while I'm putting things into words elsewhere, traditionally we do `timestamp` first, then `value` - like how you'd read a log record.

amishatishpatel: Also, technically Line 140 grabs the `aiokatcp.Sensor` object itself. I'd originally thought it got a [Reading](https://github.com/ska-sa/aiokatcp/blob/master/src/aiokatcp/sensor.py#L84-L102)
- If I could be a pain, could I ask you to rename this variable to `measured_sensor` or `sensor_obj`
- I know an IDE/the pre-commit config knows what it is, but just so the reader is also clear on what's being used

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,172 @@
...
+    tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
+    pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
+
+    sample_rate = 5
+    samples = ceil(min_sensor_time - sensor_watcher.sensors[sensor_names[0]].timestamp) * sample_rate
+    mean_power_sensor_values = np.zeros((len(sensor_names), samples))
+    mean_power_sensor_timestamps = np.zeros((len(sensor_names), samples))
+
+    async def wait_mean_power_steady_state(j: int) -> bool:
+        measurements = np.zeros((len(sensor_names), 2))
```
amishatishpatel: If I could be a small pain yet again
```suggestion
        measurements = np.zeros(shape=(len(sensor_names), 2), dtype=float)
```
- [Seems numpy defaults to numpy.float64](https://numpy.org/doc/stable/reference/generated/numpy.zeros.html#numpy-zeros), and
- Also checking via `ipython`, the Python float on my machine is also a numpy.float64
- I don't know if that changes across platforms, but at least explicitly indicating it takes out some of the guess work
- _Also_, the sensor timestamps and values are of type Python float

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,172 @@
...
+    pdf_report.step("Measure power from tied-array channelised voltage.")
+    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk()
+    tacv_data = tacv_data.astype(np.float64).view(np.complex128)[..., 0]  # Convert to complex128
+    # Only use the pass channels for beam zero for the power calculation.
+    tacv_data = tacv_data[0][pass_channels]
+    tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
+    pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
+
+    sample_rate = 5
+    samples = ceil(min_sensor_time - sensor_watcher.sensors[sensor_names[0]].timestamp) * sample_rate
```
amishatishpatel: After typing up my life story here, I think I'm following. 
TL;DR Perhaps a comment/`TODO` on why `sample_rate = 5` at the moment, please

I think a couple thinks contributed to my confusion
- I couldn't quite figure out why you needed to use a 'sample-rate' value
- Then I couldn't converge on reasoning _why_ we needed to take so many readings
- That fed into mis/understanding the retry logic/mechanism in `max_retry_test`

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Apologies for the delay. I was trying to get my suggestion to work, but now I'm frustrated and hoping to make it your problem 😅  Basically, I'm certain we can do something like
```
mean_power_sensor_readings = np.zeros(shape=(len(sensor_names), samples, 2), dtype=float)
.
.
.
async def wait_neam_power_steady_state(j: int) -> bool:
    for i, name in enumerate(sensor_names):
        sensor_obj = sensor_watcher.sensors[sensor_name]
        mean_power_sensor_readings[i, j] = (sensor_obj.timestamp, sensor_obj.value)
```

Then later in the `plot_focus` just tweak the indexing to something like
```
mean_power_sensor_readings[i, :last_sample, 0], # Timestamps
mean_power_sensor_readings[i, :last_sample, 1], # Values
```

amishatishpatel (changes requested):
Apologies if my comments seem all over the place. Was trying to make intermediate comments while I figured out if my suggestion would work.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: retrying immediately doesn't cause issues but the value is often the same as the previous value (sensor update rate isn't very high, about 0.5s) and given I want to store the data for a plot having a more consistent time resolution made sense.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: 3. is an artifact of a refactor thanks for spotting 🙏 
2. 🤷  would it be better to do a start from 1 instead?
1. valid points, naming is always fun (sarcasm), how about `test_procedure`?
I can also create an typealias for the callable signature if it makes things easier: IE.

```
alias RetryableTest = Callable[int, bool]

async def max_retry_test(
    f: RetryableTest, ...
```

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: what about the comment on line 114? could I should just clear it up a bit there?
Although I see there is a typo in it (should be .2 second intervals)

EatonEmmerich: Attempted to be clearer.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: I didn't realize I thought it was the reading!
I'd rather just grab the reading instead.

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: I see 3. is still there (i.e. Line 59 still shows `sleep_period = retry_interval`).

4. `try_number` / `attempt_nr`
  a. Can I really be a pain and ask you to rename it to `retry_attempt`? Just to be ~consistent with `retry_interval` (i.e. the `retry_` prefix)
  b. As for starting from 1 instead, I think that's wise Re: the Assertion on Line 152 (`after {samples} retries}`)
5. That alias looks pretty cool, but maybe not necessary
  a. e.g. [test-steady-state.py](https://github.com/ska-sa/katgpucbf/blob/main/qualification/tied_array_channelised_voltage/test_steady_state.py#L53) has something similar but doesn't use an alias
6. Related to the above, maybe a one-liner in the docstring here to explicitly state what you're expecting from `test_procedure`, please?
  a. e.g. `test_procedure must accept the retry iteration value and return a boolean indicating success/failure.`

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Wait, you've resolved this comment but haven't responded/the code is still the same?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -129,18 +129,17 @@ async def test_mean_power(
...
         measurements = np.zeros((len(sensor_names), 2))
         for i, name in enumerate(sensor_names):
-            measurement = sensor_watcher.sensors[name]
-            mean_power_sensor_values[i, j] = measurement.value
-            mean_power_sensor_timestamps[i, j] = measurement.timestamp
-            measurements[i] = (measurement.value, measurement.timestamp)
+            reading = sensor_watcher.sensors[name].reading
+            mean_power_sensor_readings[0, i, j] = reading.timestamp
+            mean_power_sensor_readings[1, i, j] = reading.value
+            measurements[i] = (reading.value, reading.timestamp)
```
amishatishpatel: Sorry I got sidetracked after we spoke. As discussed, you can collapse/remove the usage of the `measurement` intermediate variable.

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Ah, I guess `measurements` is no longer needed.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,174 @@
...
+        f"Mean power sensor readings from {np.min(mean_power_sensor_readings[0, :, 0])}"
+        f" to {np.max(mean_power_sensor_readings[:, last_sample, 0])}"
+        f" in {last_sample} steps."
+    )
+    mean_power_sensor_readings[:, :, 0] = mean_power_sensor_readings[:, :, 0] - mean_power_sensor_readings[:, :1, 0]
+
+    fig = Figure(tight_layout=True)
+    ax = fig.add_subplot(1, 1, 1)
+    ax.set_xlabel("Timestamp")
+    ax.set_ylabel("Mean Power")
```
amishatishpatel: I think it would be useful to add units of measurement here.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,174 @@
...
+    pdf_report.step("Compare mean-power sensors against TACV power.")
+    test_passed, last_sample = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
+    with check:
+        assert test_passed, f"Power does not agree to within 0.5% after {samples} retries."
+        assert tacv_power > 0.0
+
+    pdf_report.detail(
+        f"Mean power sensor readings from {np.min(mean_power_sensor_readings[0, :, 0])}"
+        f" to {np.max(mean_power_sensor_readings[:, last_sample, 0])}"
+        f" in {last_sample} steps."
```
amishatishpatel: Similar to the `ax.set_{x, y}label` comment, the `from` and `to` values don't have much meaning without some kind of unit of measurement.

Great idea, just needs something to tell the reader what we're counting.

amishatishpatel (changes requested):
Apologies once more for the delay in getting this back to you. I see you made a commit since we spoke earlier. I trust you will ignore/comment if something I've flagged is out of date.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: 4. I don't agree with `retry_attempt`, I see what you are trying to do with the prefix similarities, but the 'retry' in  'retry attempt' doesn't mean anything to me (do we retry the attempt? is the retry attempted? etc) where a retry interval is a very specific thing:  interval (implied as time) at which we retry.

5. To clarify, are you suggesting that the updated `test_procedure` is still insufficient?

EatonEmmerich:
added new report

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,191 @@
...
+    tacv_data = tacv_data[0][pass_channels]
+    tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
+    pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
+
+    sample_rate = 5  # TODO: NGC-2099 These are arbitrary values to get data for plotting mean power values.
+    samples = int(1e3 * sample_rate)
+    mean_power_sensor_readings = np.zeros(shape=(len(sensor_names), samples, 2), dtype=np.float64)
+
+    async def wait_mean_power_steady_state(j: int) -> bool:
+        sample_index = j - 1
```
amishatishpatel: Ooff, you're going to be so mad at me, but I'll buy you a chocolate next time we're in the office.

TL;DR Please roll this and `max_retry_test` back to indexing from 0, and just do `total_retries + 1` on Line 173.

- Completely acknowledge I was the one that suggested the whole "please index from 1" story.
- Seeing this, and the note in the `max_retry_test` docstring, that just doesn't seem right.
  - Specifically, warning the developer of `the attempt number (starting from 1)`
  - Just to solve (my issue of) the `pdf_report.detail` on Line 173 potentially reading `0 steps.`
- I think it's just a socially-accepted thing (outside of Matlab) to index from 0, and I shouldn't impose my weirdness on that

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,191 @@
...
+    pdf_report.step("Measure power from tied-array channelised voltage.")
+    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk()
+    tacv_data = tacv_data.astype(np.float64).view(np.complex128)[..., 0]  # Convert to complex128
+    # Only use the pass channels for beam zero for the power calculation.
+    tacv_data = tacv_data[0][pass_channels]
+    tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
+    pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
+
+    sample_rate = 5  # TODO: NGC-2099 These are arbitrary values to get data for plotting mean power values.
+    samples = int(1e3 * sample_rate)
```
amishatishpatel: I somehow missed this tweak during my previous review. Any particular reason you upped this to O(1e3) samples?
- I have some memory of logging the shape and it was something definitely less than 100 samples (?)

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: > 4. I don't agree with retry_attempt, I see what you are trying to do with the prefix similarities, but the 'retry' in 'retry attempt' doesn't mean anything to me (do we retry the attempt? is the retry attempted? etc) where a retry interval is a very specific thing: interval (implied as time) at which we retry.

I mean, fair enough. Can I then at least ask for `attempt_nr --> attempt_num`? I don't think we use `nr` to indicate number anywhere in the repo.

> 5. To clarify, are you suggesting that the updated test_procedure is still insufficient?

No no, I think you've more than covered it with the updated docstring. Just modulo my comment elsewhere Re: Updating the indexing (please).

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -134,40 +150,41 @@ async def test_mean_power(
...
         assert tacv_power > 0.0
 
     pdf_report.detail(
-        f"Mean power sensor readings from {np.min(mean_power_sensor_readings[0, :, 0])}"
-        f" to {np.max(mean_power_sensor_readings[:, last_sample, 0])}"
-        f" in {last_sample} steps."
+        f"Mean power sensor readings from {datetime.fromtimestamp(np.min(mean_power_sensor_readings[:, 0, 0]), UTC)}"
+        f" to {datetime.fromtimestamp(np.max(mean_power_sensor_readings[:, total_retries - 1, 0]), UTC)}"
+        f" in {total_retries} steps."
     )
```
amishatishpatel, line 157: Nice one, much appreciated.

amishatishpatel (changes requested):
Sorry I know I'm verging on being unreasonable now. Just updating my thoughts/opinions with the new information/updates you're providing!

Can I also ask you to give it one full qualification run with all config combinations? Just this test, but with a range of options.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: It used to be a value determined by the steady state time. which at the higher level doesn't really track, I made it 100 (seconds not samples) because the highest I've seen it is 70 seconds.

EatonEmmerich:
uploaded new report

amishatishpatel (approved):
Much appreciated. I used to get annoyed when reviewers would pick up something they missed in a previous round of reviewing, but I see how that's possible now that I'm on the other side!

qualification/report/generate_pdf.py
```diff
@@ -68,6 +68,7 @@
 GROUP_NAMES = {
     "general": "General tests",
     "antenna_channelised_voltage": "Antenna channelised voltage tests",
+    "tied_array_resampled_voltage": "Tied-array resampled voltage tests",
```
bmerry: Move to below `tied_array_channelised_voltage` (I think I had the same comment on NGC-1923, so hopefully this will synchronise the branches).

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+
+from katgpucbf.pytest_plugins.reporter import Reporter, plot_focus
+from katgpucbf.utils import TimeConverter
+from qualification.cbf import CBFRemoteControl
+
+from ..recv import TiedArrayChannelisedVoltageReceiver, TiedArrayResampledVoltageReceiver
+
+
+@pytest.fixture
+async def sensor_watcher(cbf: CBFRemoteControl) -> AsyncGenerator[aiokatcp.SensorWatcher, None]:
```
bmerry: This looks like copy-paste from test_control.py. I don't think you need this fixture (see my comments below where it is used), but if you still do, rather move it to the top-level conftest.py so that it doesn't need to be repeated.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    """Establish a secondary connection to the product controller with a sensor watcher.
+
+    The yielded sensor watcher is not yet synchronised.
+    """
+    # aiokatcp doesn't currently handle adding watchers after the connection
+    # is already established; SensorWatcher is also somewhat expensive. So
```
bmerry: I think the first half of that sentence might not be true as of aiokatcp 2.2.0.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    Parameters
+    ----------
+        test_procedure
+            A subroutine that takes an integer indicating the attempt number (starting from 1)
+            and returns a boolean `True` if the test passed.
+        max_retries
+            The maximum number of retries.
+        retry_interval
+           The retry interval in seconds.
+
```
bmerry: Remove one level of indentation.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    Returns
+    -------
+         A tuple containing a boolean indicating whether the test passed and the number of attempts.
```
bmerry: See https://numpydoc.readthedocs.io/en/latest/format.html#returns. There should be a type and description for each tuple element. So something like

```suggestion
    bool
        Whether the test passed
    int
        The number of attempts
```

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    for attempt_num in range(max_retries):
+        start_time = time.time()
+        if await test_procedure(attempt_num):
+            return True, attempt_num
```
bmerry, line 62: In this case it's returning the number of *unsuccessful* attempts. If that's intentional, the docstring should make that clear.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Test a subroutine with a maximum number of retries and a retry interval.
+
+    Parameters
+    ----------
+        test_procedure
+            A subroutine that takes an integer indicating the attempt number (starting from 1)
+            and returns a boolean `True` if the test passed.
+        max_retries
+            The maximum number of retries.
```
bmerry: Looking at the code, it's the maximum number of tries, rather than retries (i.e. if max_retries=1, it will just make one attempt, rather than a first attempt and then retry once).

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+async def max_retry_test(
+    test_procedure: Callable[[int], Awaitable[bool]], max_retries: int, retry_interval: float
+) -> tuple[bool, int]:
+    """
+    Test a subroutine with a maximum number of retries and a retry interval.
+
+    Parameters
+    ----------
+        test_procedure
+            A subroutine that takes an integer indicating the attempt number (starting from 1)
```
bmerry: In the code the attempt number seems to start from 0.

Also make it clear that the subroutine is asynchronous.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+            and returns a boolean `True` if the test passed.
```
bmerry:
```suggestion
            and returns a boolean `True` if the test passed or `False` if the test failed.
```
(To make it clear that even on failure, it shouldn't raise an exception)

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    Returns
+    -------
+         A tuple containing a boolean indicating whether the test passed and the number of attempts.
+    """
+    for attempt_num in range(max_retries):
+        start_time = time.time()
+        if await test_procedure(attempt_num):
+            return True, attempt_num
+        sleep_period = retry_interval - (time.time() - start_time)
+        await asyncio.sleep(sleep_period)
```
bmerry, line 64: asyncio.sleep runs on the event loop clock, which is monotonic. So rather than using `time.time()` to get time, use `asyncio.get_running_loop().time()` (of course, one can cache `asyncio.get_running_loop()` once in a local variable).

Apart from making the code robust to system clock changes (the point of a monotonic clock), it will make it easier to test this function using async_solipsism in future. I'm undecided about whether that's necessary at this point: if we expected this function to persistent in the long term I would 100% say it needs testing given that I've found a bug even after it's been reviewed by Amish. But I'm hoping this code will get deleted in the near future.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    return False, attempt_num
```
bmerry, line 65: This looks like an off-by-one bug: at this point `attempt_num` will be `max_retries - 1`, but there will have been `max_retries` attempts.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    Verification method
+    -------------------
+    Verified by means of test. Inject a white noise signal and set beam weights
+    so that a single antenna contributes. Wait until each ``mean-power`` sensor
+    timestamp is after the system steady-state timestamp (plus one
+    ``power-int-time`` so the averaging window is entirely post-steady-state).
+    Measure mean power from the tied-array channelised voltage stream over the
+    passband channels, and compare against each ``mean-power`` sensor.
+    """
+    assert receive_tied_array_resampled_voltage is not None
```
bmerry: I'd just remove the `| None` from the parameter declaration; then you don't need to do this.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    cbf: CBFRemoteControl,
+    sensor_watcher: aiokatcp.SensorWatcher,
+    pass_channels: slice,
+) -> None:
+    """Test mean-power sensor values against tied-array channelised voltage.
+
+    Verification method
+    -------------------
+    Verified by means of test. Inject a white noise signal and set beam weights
+    so that a single antenna contributes. Wait until each ``mean-power`` sensor
```
bmerry: "single antenna contributes"... what?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    async with asyncio.TaskGroup() as tg:
+        for i, name in enumerate(receive_tied_array_channelised_voltage.stream_names):
+            weights = [0.0] * len(receive_tied_array_channelised_voltage.source_indices[i])
+            weights[0] = 1.0
+            tg.create_task(pcc.request("beam-weights", name, *weights))
+
+    pdf_report.detail("Beam weights set to use antenna 0.")
+
+    pdf_report.step("Inject white noise signal.")
+    dsim_signals = "common=wgn(0.02);common;common;"
```
bmerry, line 101: I would suggest using different power levels in each polarisation, to ensure that we don't get the polarisations mixed up (which is a concern I'd like to test).

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    async with asyncio.TaskGroup() as tg:
+        for dsim_name in cbf.dsim_names:
+            tg.create_task(pcc.request("dsim-signals", dsim_name, dsim_signals))
+    pdf_report.detail(f"Set dsim signals to {dsim_signals}.")
+
+    pdf_report.step("Wait for mean-power sensors to reach steady state.")
+    await sensor_watcher.synced.wait()  # Implicitly waits for connection too
+    time_converter = TimeConverter(receiver.sync_time, receiver.scale_factor_timestamp)
+    steady_state_unix = time_converter.adc_to_unix(await cbf.steady_state_timestamp())
+    # TODO: See NGC-2099: Because the data is zero for the first several seconds,
```
bmerry: Also reference NGC-1689.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    # for now just retry with .2 second intervals (sample rate of 5) since steady state is
+    # not known beforehand.
```
bmerry: This comment seems to have drifted away from the code it applies to.

qualification/tied_array_resampled_voltage/test_vlbi.py
bmerry: 1e3 is 1000 seconds.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+
+    sample_rate = 5  # TODO: NGC-2099 These are arbitrary values to get data for plotting mean power values.
+    samples = int(1e3 * sample_rate)
+    mean_power_sensor_readings = np.zeros(shape=(len(sensor_names), samples, 2), dtype=np.float64)
+
+    async def wait_mean_power_steady_state(j: int) -> bool:
+        for i, name in enumerate(sensor_names):
+            reading = sensor_watcher.sensors[name].reading
+            mean_power_sensor_readings[i, j, 0] = reading.timestamp
+            mean_power_sensor_readings[i, j, 1] = reading.value
```
bmerry: First check `reading.status.valid_value()`. If it's false, then the reading.value is invalid (you could possibly substitute `np.nan` in that case). That could happen if the connection to the product controller drops, for example.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    return False, attempt_num
+
+
+@pytest.mark.name("VLBI mean power")
+async def test_mean_power(
+    pdf_report: Reporter,
+    receive_tied_array_resampled_voltage: TiedArrayResampledVoltageReceiver | None,
+    receive_tied_array_channelised_voltage: TiedArrayChannelisedVoltageReceiver,
+    cbf: CBFRemoteControl,
+    sensor_watcher: aiokatcp.SensorWatcher,
```
bmerry: Using a SensorWatcher is rather heavy-weight, since it gets immediate updates of all sensor values. It's even causing me some trouble in NGC-1813 (with lost packets). Since you're just pulling specific sensors when you want them, you can just use `await pcc.sensor_reading(...)` to get the readings on demand.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+
+        return bool(
+            np.all(mean_power_sensor_readings[:, j, 0] >= min_sensor_time)
+            and np.all(mean_power_sensor_readings[:, j, 1] == pytest.approx(tacv_power, rel=5e-3))
+        )
+
+    pdf_report.step("Compare mean-power sensors against TACV power.")
+    test_passed, total_retries = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
+    with check:
+        assert test_passed, f"Power does not agree to within 0.5% after {samples} retries."
```
bmerry, line 150: That tolerance should be stated in the test procedure (the docstring). Is it backed by any statistical analysis or just a thumbsuck?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+        assert tacv_power > 0.0
+
+    pdf_report.detail(
+        f"Mean power sensor readings from {datetime.fromtimestamp(np.min(mean_power_sensor_readings[:, 0, 0]), UTC)}"
+        f" to {datetime.fromtimestamp(np.max(mean_power_sensor_readings[:, total_retries, 0]), UTC)}"
+        f" in {total_retries + 1} steps."
+    )
+    mean_power_sensor_readings[:, :, 0] = mean_power_sensor_readings[:, :, 0] - mean_power_sensor_readings[:, :1, 0]
```
bmerry: Any particular reason not to shorten this to

```suggestion
    mean_power_sensor_readings[:, :, 0] -= mean_power_sensor_readings[:, :1, 0]
```

?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk()
+    tacv_data = tacv_data.astype(np.float64).view(np.complex128)[..., 0]  # Convert to complex128
+    # Only use the pass channels for beam zero for the power calculation.
+    tacv_data = tacv_data[0][pass_channels]
+    tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
+    pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
+
+    sample_rate = 5  # TODO: NGC-2099 These are arbitrary values to get data for plotting mean power values.
+    samples = int(1e3 * sample_rate)
+    mean_power_sensor_readings = np.zeros(shape=(len(sensor_names), samples, 2), dtype=np.float64)
```
bmerry, line 131: I'm finding the use of a length-2 axis to mean "timestamp" and "value" confusing and hard to follow. Can I suggest using two separate variables instead? Or if you want them in the same array, using a numpy record type so that they can still be referenced by name.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,190 @@
...
+        f"Mean power sensor readings from {datetime.fromtimestamp(np.min(mean_power_sensor_readings[:, 0, 0]), UTC)}"
+        f" to {datetime.fromtimestamp(np.max(mean_power_sensor_readings[:, total_retries, 0]), UTC)}"
+        f" in {total_retries + 1} steps."
+    )
+    mean_power_sensor_readings[:, :, 0] = mean_power_sensor_readings[:, :, 0] - mean_power_sensor_readings[:, :1, 0]
+
+    fig = Figure(tight_layout=True)
+    ax = fig.add_subplot(1, 1, 1)
+    ax.set_xlabel("Timestamp (s)")
+    ax.set_ylabel("Mean Power (dBmV)")
```
bmerry: While the telescope does measure electrical fields, the raw numbers involved are in unknown units. It's only after everything has been calibrated that one can assign absolute power to pixels in an image, and even then I don't think the voltages are known.

```suggestion
    ax.set_ylabel("Mean Power (dB)")
```

qualification/conftest.py
```diff
@@ -669,3 +688,12 @@ async def receive_tied_array_channelised_voltage(
...
     pdf_report.spead2_statistics("tied_array_channelised_voltage", delta_stats)
+
+
+@pytest.fixture
+async def receive_tied_array_resampled_voltage(
+    cbf: CBFRemoteControl,
+) -> TiedArrayResampledVoltageReceiver | None:
+    """Get the receiver for ingesting the tied-array-resampled-voltage streams."""
+    receiver = cbf.tied_array_resampled_voltage_receiver
+    return receiver
```
bmerry, line 699: I assume this will get updated from the code in NGC-1923.

qualification/requirements.txt
bmerry, line 1: Are these changes coming from merges?

bmerry (changes requested):

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: Ah right, I didn't clock the `| None` in the parameter declaration...

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: Just thumbsuck at this point.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: It was like that initially and then suggested by amish to put into a single array.
I don't think there are any good reasons for either other than code readability which is subjective.

Still trying to get into the coding style of the repo!

bmerry: @amishatishpatel what advantage do you see from bundling two heterogeneous fields into one array?

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: I have updated the documentation to reflect this, I'm not convinced this is a bug.

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: Just the old go mantra "trade cleverness for clarity" :stuck_out_tongue: 
I can never remember correctly whether it assigns first or subtracts first.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -0,0 +1,175 @@
...
+            The maximum number of attempts.
+        retry_interval
+           The retry interval in seconds.
+
+    Returns
+    -------
+        bool
+            Whether the test passed.
+        int
+            Zero-indexed attempt index upon completion.
```
bmerry: Remove a level of indentation (the type should be at the same level as "Returns"

qualification/tied_array_resampled_voltage/test_vlbi.py
amishatishpatel: @bmerry Sorry I missed this tag. 
> what advantage do you see from bundling two heterogeneous fields into one array?

I think it stemmed from a couple things
1. I didn't know the update rate of the sensor
  a. So when I saw two separate lines grabbing `sensor.timestamp` and `sensor.value`, it did raise a question of "Can these values not correspond to each other/drift at any point"
  b. That being said, I see it's still done that way now (Lines 155 and 156), so perhaps my concern is more misunderstanding than anything else
2. I figured having the sensor's `timestamp` and `value` explicitly stored alongside each other ensured there was no mistake in indexing for timestamp-value pairs e.g. when plotting
  a. I know the counter-response might be around "we're in control, we're writing the test, we're reviewing it", but maybe storing alongside each other avoids that? I thought so, anyway
 3. Re: heterogeneous fields
   a. I take it I'm being a bit naive in thinking "they're both floats, and usually yield more information when viewed _together_" ?
  b. Do you say heterogeneous because they each hold meaning in their own right? And don't necessarily need to be viewed as... symbiotic (or even codependent)

amishatishpatel: TL;DR I don't see clear and obvious advantages, more thinking it reduces the probability of error/confusion

But, as you say
> I'm finding the use of a length-2 axis to mean "timestamp" and "value" confusing and hard to follow.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -41,27 +40,27 @@ async def max_retry_test(
...
     Returns
     -------
-        bool
-            Whether the test passed.
-        int
-            Zero-indexed attempt index upon completion.
+    bool
+        Whether the test passed.
+    int
+        Zero-indexed attempt index upon completion.
```
bmerry, line 56: It now describes the actual behaviour, but I don't find this completely clear. I would say something like "Zero-indexed number of the last attempt (whether successful or not)."

qualification/tied_array_resampled_voltage/test_vlbi.py
bmerry: Well if the docs and code don't agree, there is a bug, but it might be a bug in the docs :-) Looks fixed now.

qualification/tied_array_resampled_voltage/test_vlbi.py
bmerry: Hasn't been addressed yet.

qualification/tied_array_resampled_voltage/test_vlbi.py
bmerry: Yeah I can see the arguments for keeping them together to ensure e.g. that you don't misassociate things when you slice them together. But I think having to remember what "0" and "1" mean is more harmful than what you're trying to avoid. When I say "heterogeneous" I mean that for example it doesn't make any sense to take a sum along this axis (value + timestamp is nonsensical).

A good compromise is probably a numpy record type, which keeps things together while also giving things names instead of indices.

qualification/tied_array_resampled_voltage/test_vlbi.py
bmerry: Fair enough, as long as the actual variation you observe is well below that (e.g. 0.1% or less), so that we don't fail the test when we're unlucky enough to get an outlier. In theory it may be possible to apply some statistics, but after going through Fourier transforms and resampling filters I suspect everything will be correlated with everything else in hard-to-untangle ways and so it's not worth trying.

qualification/requirements.txt
bmerry: Hasn't been answered.

bmerry (changes requested):

qualification/requirements.txt
EatonEmmerich: sorry I was meaning to revert this, as I don't remember at this point.

EatonEmmerich: Actually I already did. I see this is on the old commit

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -98,7 +98,7 @@ async def test_mean_power(
     pdf_report.detail("Beam weights set to use antenna 0.")
 
     pdf_report.step("Inject white noise signal.")
-    dsim_signals = "common=wgn(0.02);common;common;"
+    dsim_signals = "0;wgn(0.02);"  # [h, v] pol ordering
```
bmerry: Rather than testing only v, can we test both, just using different power values e.g. `wgn(0.01);wgn(0.02);` ?

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -109,53 +109,65 @@ async def test_mean_power(
...
-    ]
+    sensor_names = [f"{receiver.stream_name}.x{chan}.mean-power" for chan in range(receiver.n_chans)]
 
     pdf_report.step("Measure power from tied-array channelised voltage.")
-    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk()
+    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk_data()
     tacv_data = tacv_data.astype(np.float64).view(np.complex128)[..., 0]  # Convert to complex128
     # Only use the pass channels for beam zero for the power calculation.
-    tacv_data = tacv_data[0][pass_channels]
+    tacv_data = tacv_data[1][pass_channels]  # the first baseline's `v polarity` data.
```
bmerry: "baselines" are not applicable to beams.

It's also a bit confusing that this line has two comments: one above it and one to the right of it. Combine them into a single comment.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -109,53 +109,65 @@ async def test_mean_power(
...
     tacv_power = (np.square(tacv_data.real) + np.square(tacv_data.imag)).mean()
     pdf_report.detail(f"Mean TACV power over passband channels: {tacv_power}.")
 
     # TODO: See NGC-2099 & 1689: Because the data is zero for the first several seconds,
     # for now just retry with .2 second intervals (sample rate of 5) since steady state is
     # not known beforehand.
-    sample_rate = 5  # TODO: NGC-2099 These are arbitrary values to get data for plotting mean power values.
+    sample_rate = 5  # TODO: NGC-2099 & 1689: These are arbitrary values to get data for plotting mean power values.
```
bmerry: Again, I find it confusing to have one line commented both above and on the same line. The inline comment can just be deleted since the top-comment says everything it needs to.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -109,53 +109,65 @@ async def test_mean_power(
...
     samples = int(1e2 * sample_rate)
-    mean_power_sensor_readings = np.zeros(shape=(len(sensor_names), samples, 2), dtype=np.float64)
+    mean_power_sensor_readings = np.zeros(
```
bmerry: If you use `np.recarray` instead of `np.zeros`, then you'll be able to access timestamp and value as attributes rather than with dictionary-style access.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -109,53 +109,65 @@ async def test_mean_power(
...
-                mean_power_sensor_readings[i, j, 1] = reading.value
+                mean_power_sensor_readings[i, j]["value"] = reading.value
             else:
-                mean_power_sensor_readings[i, j, 1] = np.nan
+                mean_power_sensor_readings[i, j]["value"] = np.nan
 
         return bool(
-            np.all(mean_power_sensor_readings[:, j, 0] >= min_sensor_time)
-            and np.all(mean_power_sensor_readings[:, j, 1] == pytest.approx(tacv_power, rel=5e-3))
+            np.all(mean_power_sensor_readings["timestamp"][:, j] >= min_sensor_time)
```
bmerry: Be consistent about whether you're indexing by attribute ("timestamp") or coordinates first. When filling in the array you've done coordinates then attribute, while here you're doing attribute then coordinate.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -109,53 +109,65 @@ async def test_mean_power(
...
-        f"Mean power sensor readings from {datetime.fromtimestamp(np.min(mean_power_sensor_readings[:, 0, 0]), UTC)}"
-        f" to {datetime.fromtimestamp(np.max(mean_power_sensor_readings[:, total_retries, 0]), UTC)}"
+        f"Mean power sensor readings from {
+            datetime.fromtimestamp(np.min(mean_power_sensor_readings['timestamp'][:, 0]), UTC)
+        }"
+        f" to {datetime.fromtimestamp(np.max(mean_power_sensor_readings['timestamp'][:, total_retries]), UTC)}"
         f" in {total_retries + 1} steps."
     )
-    mean_power_sensor_readings[:, :, 0] -= mean_power_sensor_readings[:, :1, 0]
+    mean_power_sensor_readings["timestamp"] -= mean_power_sensor_readings["timestamp"][:, :1]
```
bmerry: Maybe add a comment e.g. "Shift timestamps to be relative to the start."

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,176 @@
...
+        The maximum number of attempts.
+    retry_interval
+        The retry interval in seconds.
+
+    Returns
+    -------
+    bool
+        Whether the test passed.
+    int
+        Zero-indexed attempt of the last attempt (whether successful or not).
```
bmerry: "attempt of the last attempt" doesn't make sense to me. Maybe you meant something like

```suggestion
        Zero-based index of the last attempt (whether successful or not).
```

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,176 @@
...
+
+    Verification method
+    -------------------
+    Verified by means of test. Inject a white noise signal on the v polarity dsim stream
+    and set beam weights so that a single antenna contributes to the resampled voltages.
+    Wait until each ``mean-power`` sensor timestamp is after the system steady-state timestamp
+    (plus one ``power-int-time`` so the averaging window is entirely post-steady-state).
+    Measure mean power from the tied-array channelised voltage stream v polarity over the
+    passband channels, and compare against x polarity ``mean-power`` sensor.
+    The values must agree to within 0.5%.
```
bmerry: Also add that you check that the y polarity mean-power sensors are zero.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,176 @@
...
+    time_converter = TimeConverter(receiver.sync_time, receiver.scale_factor_timestamp)
+    steady_state_unix = time_converter.adc_to_unix(await cbf.steady_state_timestamp())
+    min_sensor_time = steady_state_unix + receiver.power_int_time
+
+    sensor_names = [f"{receiver.stream_name}.x{chan}.mean-power" for chan in range(receiver.n_chans)]
+
+    pdf_report.step("Measure power from tied-array channelised voltage.")
+    _, tacv_data = await receive_tied_array_channelised_voltage.next_complete_chunk_data()
+    tacv_data = tacv_data.astype(np.float64).view(np.complex128)[..., 0]  # Convert to complex128
+    # Only use the pass channels for beam zero for the power calculation.
```
bmerry:
```suggestion
    # Only use the passband channels for beam zero for the power calculation.
```

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,176 @@
...
+            else:
+                mean_power_sensor_readings[i, j]["value"] = np.nan
+
+        return bool(
+            np.all(mean_power_sensor_readings["timestamp"][:, j] >= min_sensor_time)
+            and np.all(mean_power_sensor_readings["value"][:, j] == pytest.approx(tacv_power, rel=5e-3))
+        )
+
+    pdf_report.step("Compare mean-power sensors against TACV power.")
+    test_passed, total_retries = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
```
bmerry: Based on the updated docstring for `max_retry_test`, I would rename `total_retries` to something like `last_attempt`.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,176 @@
...
+    for i, name in enumerate(sensor_names):
+        plot_focus(
+            ax,
+            slice(0, total_retries + 1),
+            mean_power_sensor_readings["timestamp"][i, : total_retries + 1],
+            mean_power_sensor_readings["value"][i, : total_retries + 1],
+            label=name,
+        )
+    ax.legend()
+    pdf_report.figure(fig)
```
bmerry: https://github.com/ska-sa/katgpucbf/pull/1094 changed a lot of tests to run `pdf_report.figure(fig)` asynchronously (using the new `run_async` fixture) to avoid blocking the event loop. Can you do the same here?

bmerry (changes requested):

qualification/tied_array_resampled_voltage/test_vlbi.py
EatonEmmerich: Seems like mypy doesn't like that, I have suppressed the errors for now but thinking it might not be worth it

bmerry: I agree, the point of my suggestion was to make the code prettier but now it's uglier. Let's revert.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,180 @@
...
+        The maximum number of attempts.
+    retry_interval
+        The retry interval in seconds.
+
+    Returns
+    -------
+    bool
+        Whether the test passed.
+    int
+        Zero-index attempt (whether successful or not).
```
bmerry:
```suggestion
        Zero-indexed attempt (whether successful or not).
```

bmerry (changes requested):

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,180 @@
...
+
+    # TODO: See NGC-2099 & 1689: Because the data is zero for the first several seconds,
+    # for now just retry with .2 second intervals (sample rate of 5) since steady state is
+    # not known beforehand.
+    sample_rate = 5
+    samples = int(1e2 * sample_rate)
+    mean_power_sensor_readings = np.recarray(
+        (len(sensor_names), samples),
+        dtype=np.dtype([("timestamp", np.float64), ("value", np.float64)]),
+    )
```
bmerry: While @amishatishpatel initially convinced me of the merit of packaging the timestamps and values together (so that readings are objects that can be sliced and diced as a unit), looking through the code I see we don't actually take advantage of that anywhere. Every single use is `mean_power_sensor_readings[<index>].timestamp` or `mean_power_sensor_readings[<index>].value`. So I think we might as well replace those with `mean_power_sensor_timestamps` and `mean_power_sensor_values`.

qualification/tied_array_resampled_voltage/test_vlbi.py
```diff
@@ -14,14 +14,180 @@
...
+            else:
+                mean_power_sensor_readings[i, j].value = np.nan  # type: ignore[attr-defined]
+
+        return bool(
+            np.all(mean_power_sensor_readings[:, j].timestamp >= min_sensor_time)  # type: ignore[attr-defined]
+            and np.all(mean_power_sensor_readings[:, j].value == pytest.approx(tacv_power, rel=5e-3))  # type: ignore[attr-defined]
+        )
+
+    pdf_report.step("Compare mean-power sensors against TACV power.")
+    test_passed, last_attempts = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
```
bmerry:
```suggestion
    test_passed, last_attempt = await max_retry_test(wait_mean_power_steady_state, samples, 1 / sample_rate)
```

bmerry:
@EatonEmmerich I made some changes myself and I'm merging, so that we can unblock other work (e.g. NGC-2129). If you dislike any of my changes I'm happy to revisit them when you're back from leave.

bmerry (approved):

PR #1085, opened 2026-08-25
Improve debugging of qualification test packet drops

bmerry:
There are two changes here:
1. Use DiscardingChunkIterator (originally developed for vgpu) in the qualification test, so that when we're e.g. waiting for dsim signals to be set, we're discarding chunks instead of them backing up in the ringbuffer and leading to dropped packets. This significantly reduces the amount of time in which the qualification test machine shows rdma out-of-buffer errors. It doesn't completely eliminate it, which I think is because the qualification test has some bits of code that hog the CPU for significant time and prevent the discarding iterator from doing its discarding. That can be improved later by identifying such code and either farming it out to another thread or introducing yields to the event loop.
2. Capture the stream statistics before and after each test, and report the number of heaps that were incomplete or too old, and the number of times the worker thread was blocked.

Between the two of these, it should become easier to diagnose why tests sometimes lose heaps: there should be fewer cases of "qualification test host has RDMA out-of-buffer errors, but we don't know if it was while trying to capture contiguous data for the test". If there were no out-of-buffer errors, we can also detect heaps that arrived too late (in which case we need to increase the size of the reorder buffer (there have been failures in the lab where I've hypothesized that this was the case, but I had no data).

This doesn't cover reporting of incomplete chunks, which might be a useful further addition.

- [x] If qualification tests are changed: attach or link to a sample qualification report: [report.pdf](https://github.com/user-attachments/files/31425127/report.pdf) (some figures might be wonky because I didn't have fonts installed when I ran this manually); full nightly: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_karoo/705/

See NGC-1813.

### DIFF
--- a/qualification/antenna_channelised_voltage/test_delay.py
+++ b/qualification/antenna_channelised_voltage/test_delay.py
@@ -706,6 +706,39 @@ async def test_group_delay(
     await pcc.request("delays", "antenna-channelised-voltage", receiver.sync_time, *delay_models)
     pdf_report.detail(f"Set delays to {delay_model} for all inputs.")
 
+    def compute_delay(
+        rel_freqs: tuple[float, float], first_timestamp: int, raw_data: np.ndarray
+    ) -> tuple[float, float, float]:
+        """Process the data collected by :func:`measure_once` to get the result.
+
+        This is split into a helper function so that it can be dispatched to
+        a worker thread to avoid blocking the event loop.
+        """
+        # Convert Gaussian integers to complex128
+        data = raw_data.astype(np.float64).view(np.complex128)[..., 0]
+        phase = np.angle(data[1] * data[0].conj())  # Takes the difference between phases
+        del data  # Free up some memory
+        # Pick a reference timestamp at which we know the dsim will be outputting
+        # both signals with phase 0.
+        ref_timestamp = first_timestamp // dsim_period * dsim_period
+        # Phase-rotate everything to be referenced to ref_timestamp
+        rel_timestamps = np.arange(n_spectra) * receiver.n_samples_between_spectra + (first_timestamp - ref_timestamp)
+        rel_times = rel_timestamps / receiver.scale_factor_timestamp
+        delta = rel_freqs[1] - rel_freqs[0]
+        phase -= 2 * np.pi * delta * rel_times[np.newaxis, :]
+        # The phases should all be similar, but we need to avoid steps of 2pi.
+        # np.unwrap is problematic since it accumulates rounding errors. The
+        # following is good enough as long as all phases are within pi of each
+        # other.
+        phase = wrap_angle(phase - phase[0, 0]) + phase[0, 0]
+        mean_phase = wrap_angle(np.mean(phase))
+        std_phase = np.std(phase) / np.sqrt(phase.size - 1)
+        scale = receiver.scale_factor_timestamp / (2 * np.pi * delta)
+        delay = -mean_phase * scale
+        period = 2 * np.pi * scale
+        std = std_phase * scale
+        return delay, period, std
+
     async def measure_once(rel_freqs: tuple[float, float]) -> tuple[float, float, float]:
         """Estimate group delay for a single pair of frequency offsets.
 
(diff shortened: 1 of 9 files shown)


bmerry:
I've pushed some changes since the attached qualification test. I'll need to do another run to ensure that I haven't broken anything.

bmerry:
RDMA out-of-packet errors:

main:
main

This branch:
branch

qualification/antenna_channelised_voltage/test_delay.py
```diff
@@ -775,8 +776,8 @@ async def measure_once(rel_freqs: tuple[float, float]) -> tuple[float, float, fl
         # First axis corresponds to the 2 signals we're comparing.
         raw_data = np.ones((2, n_channels, n_spectra, COMPLEX), np.int8)
         try:
-            async with asyncio.timeout(acc_time * 3 + 10.0):
-                async for timestamp, chunk in receiver.complete_chunks():
+            async with asyncio.timeout(acc_time * 3 + 10.0), aclosing(receiver.complete_chunks()) as it:
```
amishatishpatel, line 779: I feel like I just read a note recommending using this, but I can't remember where.

That being said, why was the old Line 779 insufficient? Re: Ensuring
> the DiscardingIterator context manager is exited when we stop iterating heaps.

qualification/recv.py
```diff
@@ -177,6 +178,10 @@ async def complete_chunks(
 
         Each yielded value is a ``(timestamp, chunk)`` pair.
 
+        It is important that this asynchronous generator is closed as soon as
+        it is no longer needed, for example, by using
+        :func:`contextlib.aclosing`.
```
amishatishpatel, line 183: Ah, I read this in my editor before coming back to review this commit.

qualification/recv.py
```diff
@@ -642,3 +696,21 @@ def chunk_place(data_ptr, data_size, user_data_ptr):
...
+    meaningfully differenced. Note that the dictionary is only populated on
+    exit from the context manager.
+    """
+    delta_stats: dict[str, int] = {}
+    init_stats = functools.reduce(operator.add, [stream.stats for stream in stream_group])
+    yield delta_stats
+    final_stats = functools.reduce(operator.add, [stream.stats for stream in stream_group])
+    for stat in final_stats.config:
+        if stat.mode == spead2.recv.StreamStatConfig.Mode.COUNTER:
+            delta_stats[stat.name] = final_stats[stat.name] - init_stats[stat.name]
```
amishatishpatel, line 716: Sorry I know I'm being a bit lazy, just want to make this comment before I lose my train of thought. TL;DR Why/How does the report "only" show four stats?
- I'm trying to trace where the `stream.stats` gets initialised/told to only track those in the screenshot below
- `grep`'ing about `spead2` I found the [Core Statistics](https://spead2.readthedocs.io/en/latest/recv-stats.html#core-statistics) as well as [stream stat indices.pyi](https://github.com/ska-sa/spead2/blob/master/src/spead2/recv/stream_stat_indices.pyi) which is a larger list/collection

Image

amishatishpatel (commented):
Minor comments for clarification. I think I lost my train of thought a couple of times while I started delving deeper into some of the changes/updates.

On the whole, very smart addition and no real complaints. Perhaps one slightly-related question:
- Did you think about moving `utils.py::DiscardingIterator` into `recv.py` ?

qualification/antenna_channelised_voltage/test_delay.py
bmerry: If the timeout fires, it will throw a TimeoutError exception, which will break out of the `async for` loop without closing the iterator.

qualification/recv.py
bmerry: I'm collecting all the counter-mode stats in the JSON file, but only putting the ones I think are important in the report. That way if we later need more stats while debugging we can still get them retroactively.

bmerry:
> Did you think about moving utils.py::DiscardingIterator into recv.py ?

I didn't think of it. I'm not against it if you think it's a good idea, but also DiscardingIterator is a general-purpose class even though we only currently use it for recv.

qualification/recv.py
amishatishpatel: Oh gosh, it's right there (but also, good to know not all stats are counter-mode).

amishatishpatel:
> > Did you think about moving utils.py::DiscardingIterator into recv.py ?
> 
> I didn't think of it. I'm not against it if you think it's a good idea, but also DiscardingIterator is a general-purpose class even though we only currently use it for recv.

Honestly I'm 50/50 at the moment. Maybe it's because I can't envision any other usage for it other than in a recv-type scenario. I'll create a ticket in the event it becomes more obvious it's solely a recv-specific thing.

amishatishpatel (approved):
One thing I forgot to respond to from your PR blurb
> This doesn't cover reporting of incomplete chunks, which might be a useful further addition.

Would you like to track this under the same ticket?

I've separately created NGC-2108 in the event DiscardingIterator shows more interest in moving in to recv.py.

bmerry:
> One thing I forgot to respond to from your PR blurb
> 
> > This doesn't cover reporting of incomplete chunks, which might be a useful further addition.
> 
> Would you like to track this under the same ticket?

Yes, but I'll make a new PR for it.

PR #1089, opened 2026-09-08
vgpu: increase reorder tolerance

bmerry:
It was previously hardcoded to 2 chunks, with no consideration for how
big a chunk is, and in practice this seems to be too small in some cases
(NGC-2099).

Replace it by having the user specify the tolerance in bytes (which
makes the memory usage predictable), and compute the number of chunks
from that. It's also made a command-line parameter so that it'll be
easier to experiment with in future.

The default value is 128 MiB, which is 4x the previous effective
value, while still being reasonably modest.

Additionally, the calculation of the total number of chunks to allocate had to be updated. It was previously hard-coded to 4, allowing for 2 chunks in the reorder buffer plus 2 being worked on or queued. I've changed that to 2 plus the reorder buffer size.

- [x] If qualification tests are changed: attach or link to a sample qualification report: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_karoo/740/ (it includes some extra commits, but it gives me confidence that this is working)
- [x] Re-test latest version on site

Closes NGC-2099.

### DIFF
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -159,6 +159,7 @@ class RecvConfig:
     comp_vector: int
     buffer_size: int
     pols: tuple[str, str]
+    reorder_tol_bytes: int
 
     @property
     def pol_labels(self) -> list[str]:
@@ -362,7 +363,12 @@ class VEngine(Engine):
         layout = config.recv_config.layout
         dtype = np.dtype(f"int{layout.sample_bits}")
         recv_group = recv.make_stream_group(
-            layout, data_ringbuffer, free_ringbuffer, config.recv_config.affinity, config.recv_config.pol_labels
+            layout,
+            data_ringbuffer,
+            free_ringbuffer,
+            config.recv_config.affinity,
+            config.recv_config.pol_labels,
+            config.recv_config.reorder_tol_bytes,
         )
         for _ in range(recv_chunks):
             chunk = recv.Chunk(
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -18,6 +18,7 @@
 
 import asyncio
 import logging
+import math
 from collections.abc import AsyncIterator, Sequence
 from dataclasses import dataclass
 from fractions import Fraction
@@ -354,10 +355,13 @@ class VEngine(Engine):
 
     def _init_recv(self) -> None:
         """Initialise the receiver state."""
-        config = self.config
-        layout = config.recv_config.layout
+        recv_config = self.config.recv_config
+        layout = recv_config.layout
         data_ringbuffer_chunks = 2  # TODO: may need tuning
-        max_active_chunks = recv.max_active_chunks(layout, config.recv_config.reorder_tol_bytes)
+        # The + 1 is because we want the distance from the end of the oldest chunk
+        # to the start of the newest chunk (a distance of n-1 chunks) to be at
+        # least REORDER_BYTES.
+        max_active_chunks = math.ceil(recv_config.reorder_tol_bytes / layout.chunk_bytes) + 1
         total_chunks = data_ringbuffer_chunks + max_active_chunks
         data_ringbuffer = ChunkRingbuffer(
             data_ringbuffer_chunks, name="recv_data_ringbuffer", task_name="run", monitor=self.monitor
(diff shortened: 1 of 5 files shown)


bmerry:
This should probably go through a full qualification test (at least of test_control.py) before being merged, but it's successfully fixed the configuration identified in NGC-2099.

test/vgpu/test_recv.py
```diff
@@ -98,7 +99,11 @@ def stream_group(
     It is connected to the :func:`queues` fixture for input and
     :func:`data_ringbuffer` for output.
     """
-    stream_group = recv.make_stream_group(layout, data_ringbuffer, free_ringbuffer, -1, POL_LABELS)
+    # We want this to be larger than the entire input in test_recv, so that
+    # the test doesn't depend on the two receiving threads making progress at
+    # the same rate.
+    reorder_tol_bytes = math.ceil(30 / layout.chunk_batches) * layout.chunk_bytes
+    stream_group = recv.make_stream_group(layout, data_ringbuffer, free_ringbuffer, -1, POL_LABELS, reorder_tol_bytes)
```
EatonEmmerich: Question: Would it be possible to simply have a change the size of the layout to be large enough so the reorder_tol_bytes here could be zero?

src/katgpucbf/vgpu/recv.py
```diff
@@ -202,12 +203,21 @@ def chunk_place_impl(data_ptr, data_size, user_data_ptr):
         return chunk_place_impl
 
 
+def max_active_chunks(layout: Layout, reorder_tol_bytes: int) -> int:
```
EatonEmmerich: nit: I'm not convinced this is the correct module for this function.

src/katgpucbf/vgpu/recv.py
```diff
@@ -37,8 +38,8 @@
 from ..utils import TimeConverter
 from . import METRIC_NAMESPACE
 
-#: Number of chunks to allow to be under construction
-MAX_CHUNKS = 2  # TODO: may need to increase to tolerate reordering
+#: Number of bytes to allow for chunks under construction
+REORDER_BYTES = 64 * 1024 * 1024
```
EatonEmmerich: Unused const?

EatonEmmerich (changes requested):

src/katgpucbf/vgpu/engine.py
```diff
@@ -354,17 +355,24 @@ def __init__(
     def _init_recv(self) -> None:
         """Initialise the receiver state."""
         config = self.config
-        recv_chunks = 4  # TODO: may need tuning?
+        layout = config.recv_config.layout
+        data_ringbuffer_chunks = 2  # TODO: may need tuning
+        max_active_chunks = recv.max_active_chunks(layout, config.recv_config.reorder_tol_bytes)
+        total_chunks = data_ringbuffer_chunks + max_active_chunks
```
EatonEmmerich: seems a bit strange to assign to a var for a single use

test/vgpu/test_recv.py
bmerry: You mean, increasing the chunk size enough that the original `MAX_CHUNKS = 2` would have been sufficient? That might have worked, but it would have meant that the chunks flowing down the GPU processing pipeline would have gotten bigger too, and I didn't want to mess with that when its working (e.g. it might have bloated GPU memory usage too much). I felt it was better to decouple the chunk sizing for GPU work from the reordering tolerance for networking so that we can tune either without significantly interfering with the other (other than that the reorder buffer will always be an integral number of chunks).

src/katgpucbf/vgpu/recv.py
bmerry: I see what I did in xbgpu is compute max_active_chunks in engine and pass it to `make_stream`. I'll update vgpu to follow the same pattern, which will eliminate the need for a separate function for this.

src/katgpucbf/vgpu/recv.py
bmerry: Yes - not sure why mypy didn't flag that (maybe because it's public). I was originally going to just hard-code but then went with a cmdline arg.

src/katgpucbf/vgpu/engine.py
bmerry: It's used in two places.

EatonEmmerich: My bad missed one I guess :bow:

EatonEmmerich (approved):

PR #1091, opened 2026-09-10
Prime the pipeline on startup

bmerry:
cupy uses memory pooling to avoid having to reallocate GPU memory when arrays of the same size are allocated over and over. That means that the first capture-start has to do a lot of extra work to do the initial allocations, which stalls the pipeline on the first capture-start and tends to lead to a few seconds of data loss.

To reduce this (it doesn't seem to eliminate it completely), create a dummy pipeline on startup and push some dummy data through it, to get some suitably-sized chunks into the pool.

NGC-2126 will take this a bit further, but I'm leaving that for now to avoid conflicts with NGC-1689.

- [ ] If design has changed: ensure documentation is up to date (**still TODO**)

Helps with NGC-2099.

### DIFF
--- a/src/katgpucbf/vgpu/engine.py
+++ b/src/katgpucbf/vgpu/engine.py
@@ -22,6 +22,7 @@ import math
 from collections.abc import AsyncIterator, Sequence
 from dataclasses import dataclass
 from fractions import Fraction
+from typing import override
 
 import aiokatcp
 import cupy as cp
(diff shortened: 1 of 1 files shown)


src/katgpucbf/vgpu/engine.py
```diff
@@ -353,6 +379,31 @@ def __init__(
...
+        )
+        # It's arbitrary, but needs to be in the past so that it's in the
+        # scope of leap second tables.
+        time_base = Time("2020-01-01T00:00:00", scale="utc")
+        time_scale = Fraction(samples_between_spectra) / Fraction(recv_config.time_converter.adc_sample_rate)
+        it = DummyRecvStream(arr, time_base, time_scale)
+        dummy_sensors = aiokatcp.SensorSet()
+        # Use a large value for sensor timeout, since we're not actually interested
+        self._populate_sensors(dummy_sensors, recv_config.pol_labels, self.config.send_config.pols, 1e6)
+        frameset_it = _CaptureSession.frameset_iterator(self.config, dummy_sensors, it)
```
EatonEmmerich, line 403: Question: would these dummy sensors be available via katcp after this point?

src/katgpucbf/vgpu/engine.py
```diff
@@ -410,9 +461,8 @@ def _init_recv(self) -> None:
                 recv_config.pol_labels,
             )
         )
-
-        for stream in recv_group:
-            stream.start()
+        # We do not start the streams yet; that's done in start() so that
+        # it's done only once we're 100% ready to start receiving data.
```
EatonEmmerich, line 465: I don't think the comments here are neccesary as this is what's also done in xbgpu (unless I'm mistaken?)

EatonEmmerich (approved):
Looks good, some questions & nits though

src/katgpucbf/vgpu/engine.py
bmerry: No, because `dummy_sensors` is independent of the server's sensor set.

src/katgpucbf/vgpu/engine.py
bmerry: xbgpu is not quite the same because it doesn't seem to use spead's explicit_start feature (I think because it only uses one stream instead of a stream group); but it's similar in that the readers are only added during DeviceServer.start. I don't think one should need to read xbgpu's code to understand what's happening in vgpu though.

PR #1094, opened 2026-09-16
Drastically reduce RDMA out-of-buffer packet drops in qualification

bmerry:
I don't seem to have eliminated them all yet, but it's a 100x reduction. Also report the number of RDMA out-of-buffer drops in the report, and log a warning if it's non-zero, to help isolate remaining cases. The remaining drops seem to be concentrated on tied_array_channelised_voltage/test_delay.py::test_delay and general/test_control.py::test_control.

Before:
image

After:
image

I edited the commit history to make a series of small self-contained commits, and it might be easier to review that way.

- [x] If qualification tests are changed: attach or link to a sample qualification report: https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_karoo/749/

Partly addresses NGC-1813.


bmerry:
Incidentally, the qualification test was 8 minutes slower than main (6h12m instead of 6h4m). I'm guessing that's because we're no longer doing parallel gain setting. I don't think it's a major issue.

vincentseotlo (approved):
Happy

PR #1099, opened 2026-09-21
A few small changes to NGC-2023 branch

bmerry:
@amishatishpatel would you mind looking at these? They're changes I want to make to Eaton's NGC-2023, and I'd like to just get it down rather than wait until he's back from leave.

I've made a qualification test run at https://jenkins.cbf.kat.ac.za/job/qualification_katgpucbf_lab/719/


amishatishpatel (approved):
Sorry I completely forgot I'd responded to you over on the original NGC-2023 PR around splitting that ndarray (and didn't go back to see any further discussion).

PR #1101, opened 2026-09-21
More precisely document how delay is computed for an output spectrum

bmerry:
Closes NGC-1814.


bmerry:
@mgouws FYI - hopefully this shouldn't be a surprise but I don't think it was previously written down precisely in MK+ CBF documentation. Potentially it may be surprising that the delay for a spectrum is based on the timestamp of the first sample, rather than the middle sample.

doc/signal_path.rst
```diff
@@ -94,6 +94,31 @@ double precision then reduced to single precision for application. Conversion
...
+
+- :math:`t_0` be the load time discussed above (after rounding to the nearest
+  ADC sample);
+- :math:`d_0` be the fixed delay from the delay polynomial;
+- :math:`d'` be the rate of change of delay; and
+- :math:`t` be the timestamp associated with an output spectrum, which
+  in turn is the (post-delay) timestamp of the *first* input sample that is used in
+  the polyphase filter-bank (see below).
+
+Then the computed delay is :math:`d_0 + (t-t_0)d'`, and similarly for phase.
```
amishatishpatel, line 117: I'll be honest, I don't think there is a line explicitly/exactly like this in e.g. [fgpu/delay.py](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/delay.py) so I'm ~trusting you on this.
- I found something close to it in LinearDelayModel::range, but I wasn't ready to dive into the other DelayModels' `range` methods
  - i.e. I can't make sense of it right now, and that shouldn't hold this up

HTH

amishatishpatel (approved):
Just FYI, in the current build of docs, this addition is ~split in two by the "Signal path for wide-band channelisation" tikz plot - see pages 5, 6 and 7.

[katgpucbf_NGC-1814.pdf](https://github.com/user-attachments/files/32556706/katgpucbf_NGC-1814.pdf)

doc/signal_path.rst
bmerry: It's at https://github.com/ska-sa/katgpucbf/blob/47e8c100f32982d64567f15614c4ee093c6e3f91/src/katgpucbf/fgpu/delay.py#L150-L151. If you consider just the first element of the range, then
- `self.start` corresponds to t_0
- `start` corresponds to t (and more generally, `rel_time` corresponds to t-t_0 as t ranges from `start` to `stop`
- `self.delay` corresponds to d_0
- `self.delay_rate` corresponds to d'

Does that help?

MultiDelayModel just accumulates other delay models over time, and AlignedDelayModel just tweaks the split into coarse and fine delays, but neither of them changes the fundamentals.

amishatishpatel: Absolutely helps because that's how I was reading it! What a relief. Appreciate you pointing it out nonetheless.

amishatishpatel (approved):

